PR #8, opened 2016-08-09
provide a preview of each field when displaying layers

yeesian:


yeesian:
failing only on nightlies, so I'm merging

PR #12, opened 2016-09-19
julia v0.5 compatibility

visr:
Note that this also adds `Factcheck.exitstatus()` and changes the results of two tests in order to make them pass. Let's see what Travis says.

Sorry about all the whitespace changes. You can add `?w=1` to the URL to ignore it in the diff.


yeesian:
Thanks!

PR #14, opened 2016-09-29
don't require or import Dates

visr:
it is in Base since julia v0.4, and putting it in the test/REQUIRE will cause the Dates.jl packages that is there for v0.3 compat to be downloaded

`import Dates` does nothing since Dates as a module is exported into Base

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,2 +1,1 @@
 FactCheck
-Dates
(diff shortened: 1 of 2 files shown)


yeesian:
Thanks!

yeesian:
Ohshucks, I did that in a hurry, and it seems to have borked Travis

visr:
Ha that was very fast indeed. But CI is now running on the new master, so if that's green then all is ok.

PR #16, opened 2017-10-01
Update to Julia 0.6, drop 0.5

evetion:
Upgrade to 0.6
- Dropped 0.5 support
- Fixed all 0.6 deprecation warnings
- Readme now includes import Base.read
- Travis builds fixed, now with colors
- Included projection tests (was commented out)

This introduced some serious errors:
- Unescaped arguments in macros in 0.6, see JuliaLang/julia/#19587 and PR #15850
- Pointer_to_array was removed in 0.5

These are fixed now, no thanks to FactCheck (FC)
- FC complained about 8 failings tests, but not where. Compact output showed roughly where.
- The errors were actually the non existent pointer_to_array functions, which somehow didn't stop FC
- FC needs to be updated to 0.6, all remaining deprecation warnings are from FC.

Would advise to move away from FC.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - release
+  - 0.6
   - nightly
 matrix:
   allow_failures:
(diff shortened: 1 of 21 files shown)


yeesian:
Thanks for this! I really appreciate it.

And yeah, I was on FC for a while when people were still complaining that `Base.Test` was slow, but that's no longer the case, and I'll be happy to make the switch.

PR #20, opened 2017-11-10
Update for latest Julia/GDAL/Clang.jl

yeesian:
Ref https://github.com/visr/GDAL.jl/pull/41

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ julia:
   - nightly
 matrix:
   allow_failures:
+  - os: linux
   - julia: nightly
   fast_finish: true
 notifications:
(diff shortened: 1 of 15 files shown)


richardreeve:
@yeesian This update is working for me, and obviously master doesn't. It would be great if you could merge this.

yeesian:
Thanks for reminding me! I paused because of the TravisCI builds; do you know what's the cause of the build failures?

richardreeve:
The error message suggests that GDAL needs to be built - using `Pkg.build("GDAL")` - and it doesn't happen, even though you seem to be doing it in `.travis.yml`.

PR #26, opened 2018-03-27
Fix travis build by circumventing GDAL problem

evetion:
Circumvents visr/GDAL.jl#42
See also visr/GDAL.jl#45


evetion:
Travis seems happy about this, so you could enable it again. I'll look into the Windows failing builds over at visr/GDAL.jl first, if I find a solution, I'll make another PR.

yeesian:
Sure, thanks for overseeing everything!

PR #29, opened 2018-03-28
Introduce ISpatialRef and enable more tests

### LINKED ISSUE #19: Enable more tests (finish unfinished tests)
In several test files, before and after statements are printed. These files often contain no functional tests and are thus only partly finished:

```julia
AG.importEPSG(26912) do spatialref
        println("before (Proj4): $(AG.toPROJ4(spatialref))")
        println("before (WKT): $(AG.toWKT(spatialref))")
        AG.morphtoESRI!(spatialref)
        println("after (Proj4): $(AG.toPROJ4(spatialref))")
        println("after: $(AG.toWKT(spatialref))")
end
```

We should finish these tests, they can be used at the least for regression testing.

evetion:
At places where things are only printed right now. Fixes #19.


test/test_cookbook_geometry.jl
```diff
@@ -502,124 +543,139 @@ end
...
     AG.fromWKT(wkt) do geom
-        println("Geometry has $(AG.npoint(geom)) points")
+        # TODO This should be 4!
+        @test AG.npoint(geom) == 0
     end
 
     # Method 2
-    println("Geometry has $(AG.npoint(AG.fromWKT(wkt))) points")
+    # TODO This should be 4!
+    @test AG.npoint(AG.fromWKT(wkt)) == 0
```
evetion: Should npoint return 4 here?

test/test_cookbook_geometry.jl
```diff
@@ -502,124 +543,139 @@ end
...
 end
 
 @testset "Iterate over Geometries in a Geometry" begin
     wkt = "MULTIPOINT (1181866.263593049 615654.4222507705, 1205917.1207499576 623979.7189589312, 1227192.8790041457 643405.4112779726, 1224880.2965852122 665143.6860159477)"
     AG.fromWKT(wkt) do geom
-        for i in 0:(AG.ngeom(geom)-1)
-            println("$i). $(AG.toWKT(AG.getgeom(geom,i)))")
-        end
+        # TODO Should getgeom use Julian counting from 1?
+        @test AG.toWKT(AG.getgeom(geom, 3)) == "POINT (1224880.29658521 665143.686015948)"
```
evetion: Should getgeom use Julian indexing?

yeesian: yeah I was conflicted on it for a while, but we should

test/test_cookbook_geometry.jl
yeesian: yeah it should! what's the reason why it's returning 0 here?

evetion: I don't know, seems a bug to me.

evetion: Created #30 for this, as tests are now failing on this.

yeesian: My recommendation is to keep it at 0 for now, and introducing a new test for `AG.ngeom` here to demonstrate the contrast in behavior, while we debate about what's the right way to go at #30

evetion: Agreed.

yeesian:
This is a good effort, thanks! I'll be happy to take over from here.

evetion:
Thanks! Haven't found the time to finish it.

PR #32, opened 2018-03-28
First Round of Documentation

yeesian:


docs/src/memory.md
```diff
@@ -0,0 +1,99 @@
...
+mutable struct StyleTable;                    ptr::GDALStyleTable         end
+mutable struct StyleTool;                     ptr::GDALStyleTool          end
+```
+and makes it the responsibility of the user to free the allocation of memory from GDAL, by calling `ArchGDAL.destroy(obj)` (which sets `obj.ptr` to `C_NULL` after destroying the GDAL object corresponding to `obj`).
+
+## Manual versus Context Management
+
+There are two approaches for doing so.
+
+1. The first uses the [`unsafe_` prefix](http://docs.julialang.org/en/release-0.4/manual/style-guide/#don-t-expose-unsafe-operations-at-the-interface-level) to indicate methods that returns objects that needs to be manually destroyed.
```
evetion: https://docs.julialang.org/en/v0.6.2/manual/style-guide/#Don't-expose-unsafe-operations-at-the-interface-level-1 is a more up to date version.

yeesian:
Any idea why the gh-page branch was not created? Travis log here: https://travis-ci.org/yeesian/ArchGDAL.jl/jobs/360027094#L1697-L2170

evetion:
Did you follow all instructions [here](https://juliadocs.github.io/Documenter.jl/v0.7.2/man/hosting.html#SSH-Deploy-Keys-1)?

I've created the gh-pages branch, as it has to be created manually according to above documentation, but that still doesn't work. You could try do the following according to Documenter:

` You can set DOCUMENTER_DEBUG to "true" in Travis to see more information.`

yeesian:
I was going by [this link](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#gh-pages-Branch-1), which says that

> By default, Documenter pushes documentation to the `gh-pages` branch. If the branch does not exist it will be created automatically by `deploydocs`. If does exist then Documenter simply adds an additional commit with the built documentation. You should be aware that Documenter may overwrite existing content without warning.
> 
> If you wish to create the `gh-pages` branch manually the that can be done following [these instructions](https://coderwall.com/p/0n3soa/create-a-disconnected-git-branch).

PR #49, opened 2018-08-01
enable more tests after fixes in GDAL.jl

visr:


yeesian:
This is great (tracing all the way to the fixes in GDAL.jl), thanks!

PR #53, opened 2018-08-31
support julia 1.0, drop 0.6

### LINKED ISSUE #51: Doesn't work with Julia 1.0
The module fails to load with Julia 1.0, some of the reasons are that `Base.Dates` was replaced with just `Dates` or that `Void` was replaced with `missing`.

visr:
Didn't have time yet to fix iterators.jl, but it thought let's just put this up.
Fixes #51

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,18 +4,19 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
+  - 0.7
+  - 1.0
   - nightly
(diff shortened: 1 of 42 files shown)


yeesian:
LGTM. I've pushed some commits to update the iterators code, feel free to merge if everything's green on your end.

yeesian:
Thanks!

PR #57, opened 2018-09-26
Added permuteddims view

### LINKED ISSUE #37: transposed rasters
If a complete band is read into an array, the x and y dimensions are flipped:
```julia
using ArchGDAL

filepath = "dem.tif"

arr = ArchGDAL.registerdrivers() do
    ArchGDAL.read(filepath) do dataset
        ArchGDAL.read(dataset, 1)
    end
end
summary(arr) # => 97×33 Array{UInt8,2}

run(`gdalinfo $filepath`)
⋮
Size is 97, 33 # this is ncol, nrow
Band 1 Block=97x1 Type=Byte, ColorInterp=Undefined # blocks are single rows
⋮
```

This is because GDAL is row-based and Julia column-based.
Some options regarding this:
- Keep as is, warn in the docs that users must take this into account
- Call `permutedims` ([docs](https://docs.julialang.org/en/latest/base/arrays/#Base.permutedims)) before returning the array. Makes a copy of the whole dataset.
- Return a `PermutedDimsArray` ([docs](https://docs.julialang.org/en/latest/base/arrays/#Base.PermutedDimsArrays.PermutedDimsArray)). This does not create a copy, but a permuted view.

I haven't really worked with `PermutedDimsArray` yet, but to me it seems like the best choice right now. Thoughts?

evetion:
Fixes #37

### DIFF
--- a/test/test_ospy_examples.jl
+++ b/test/test_ospy_examples.jl
@@ -108,7 +108,7 @@ AG.read("ospy/data1/sites.shp") do input
         end end end
         @test sprint(print, output) == """
         GDAL Dataset (Driver: Memory/Memory)
-        File(s):
+        File(s): 
 
         Number of feature layers: 1
           Layer 0: hw1b (wkbPoint)
@@ -129,7 +129,7 @@ AG.read("ospy/data1/sites.shp") do input
         end
         @test sprint(print, output) == """
         GDAL Dataset (Driver: Memory/Memory)
-        File(s):
+        File(s): 
 
         Number of feature layers: 1
           Layer 0: hw1b (wkbPoint)
@@ -335,7 +335,7 @@ AG.read("ospy/data4/aster.img") do ds
                 end
                 @test sprint(print, outDS) == """
                 GDAL Dataset (Driver: MEM/In Memory Raster)
-                File(s):
+                File(s): 
 
                 Dataset (width x height): 5665 x 5033 (pixels)
                 Number of raster bands: 1
@@ -419,7 +419,7 @@ end end end end
 
                 @test sprint(print, dsout) == """
                 GDAL Dataset (Driver: MEM/In Memory Raster)
-                File(s):
+                File(s): 
 
                 Dataset (width x height): 4500 x 3000 (pixels)
                 Number of raster bands: 1
--- a/test/test_ospy_examples.jl
+++ b/test/test_ospy_examples.jl
@@ -108,7 +108,7 @@ AG.read("ospy/data1/sites.shp") do input
         end end end
         @test sprint(print, output) == """
         GDAL Dataset (Driver: Memory/Memory)
-        File(s): 
+        File(s):
 
         Number of feature layers: 1
           Layer 0: hw1b (wkbPoint)
@@ -129,7 +129,7 @@ AG.read("ospy/data1/sites.shp") do input
         end
         @test sprint(print, output) == """
         GDAL Dataset (Driver: Memory/Memory)
-        File(s): 
+        File(s):
 
         Number of feature layers: 1
           Layer 0: hw1b (wkbPoint)
@@ -137,7 +137,7 @@ AG.read("ospy/data1/sites.shp") do input
     end
 end
 end
-    
+
 @testset "Homework 2" begin
     # http://www.gis.usu.edu/~chrisg/python/2009/lectures/ospy_hw2a.py
     open("ospy/data2/ut_counties.txt", "r") do file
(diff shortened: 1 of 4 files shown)


test/test_ospy_examples.jl
```diff
@@ -129,15 +129,15 @@ AG.read("ospy/data1/sites.shp") do input
         end
         @test sprint(print, output) == """
         GDAL Dataset (Driver: Memory/Memory)
-        File(s): 
+        File(s):
```
yeesian: the display tests are white-space sensitive.

test/test_ospy_examples.jl
```diff
@@ -108,7 +108,7 @@ AG.read("ospy/data1/sites.shp") do input
         end end end
         @test sprint(print, output) == """
         GDAL Dataset (Driver: Memory/Memory)
-        File(s): 
+        File(s):
```
yeesian: the display tests are white-space sensitive.

test/test_ospy_examples.jl
```diff
@@ -419,7 +419,7 @@ end end end end
 
                 @test sprint(print, dsout) == """
                 GDAL Dataset (Driver: MEM/In Memory Raster)
-                File(s): 
+                File(s):
```
yeesian: the display tests are white-space sensitive.

test/test_ospy_examples.jl
```diff
@@ -335,7 +335,7 @@ AG.read("ospy/data4/aster.img") do ds
                 end
                 @test sprint(print, outDS) == """
                 GDAL Dataset (Driver: MEM/In Memory Raster)
-                File(s): 
+                File(s):
```
yeesian: the display tests are white-space sensitive.

evetion:
Sorry, my editor went out of its way to delete all trailing whitespaces. Do you want me to revert all the whitespace changes?

yeesian:
Just the tests needed to be fixed -- the rest of the whitespace changes are fine.

evetion:
- Documentation needs to be updated
- @visr Keyword usage. Rasterio is the best place to change it, but the keyword change is a few levels higher up. Still have to define which functions, all `read()`?

visr:
Just discussed about the keyword usage with @evetion. Since the construction of the `PermutedDimsArray` view is quite fast, it is probably sufficient to just construct them by default, and if you don't want it permuted, you can call `parent(A)` afterwards to opt out. And if you don't want the view, you can `collect(A)` to force it to do the permutation and get a plain `Array`. If we document these two options it may be sufficient.

```julia
julia> @btime PermutedDimsArray($A2D, (2,1));
  380.641 ns (4 allocations: 176 bytes)
julia> @btime PermutedDimsArray($A3D, (3,2,1));
  1.190 μs (9 allocations: 560 bytes)
```

evetion:
Documentation is done, if you agree, we can merge this and go ahead with registering. 🎉

yeesian:
LGTM. Feel free to merge if @visr approves :)

visr:
Trying to see if this is indeed this easy haha.

The `rasterio!` functions are also used for writing (called from `write!` with `GDAL.GF_WRITE`). Should we not permute before handing it off to GDAL in this case?

And if people want to use `read!`, `readblock!`, `writeblock!` directly they should prepare their buffer in `(ncol, nrow, [nband])` as opposed to `([nbband], nrow, ncol)` right? That may be fine, we probably have to draw a line somewhere, but would be good to document that as well.

yeesian:
Oh yes, indeed you're right on all counts. Thanks for being thorough about this: I don't want to be a blocker, but I really appreciate that you're not rushing through with this PR.

evetion:
Well, at the moment the write functionality is unchanged, as is the read!.
If we do permute there, we need to change the documentation, although it's already quite vague what the dimensions of the buffer should be. The write methods specify, `write(buffer, indices, rows, cols)`, while the buffer should be reversed from that.

We could document these better, but I'm starting to doubt this PR.

- We can't detect incoming dimensions for writing, only document them
- Permuting (copying) for writing is probably unwanted
- So current writing seems best, but reading gives a view with different dimensions.

Maybe it's better to 
- Leave ArchGDAL as GDAL is, per its principles(!)
- Document that GDAL in Julia gives permuted dimensions and it's up to the user to permute
- Only have specific `imread` and `imwrite` functions that do **both** permute. (image branch)
- Or leave permuting to other packages

yeesian:
That sounds reasonable to me. To state what I understand of your comment: either the user

- (a) deals with the "raw data" (through the current `read()`/`write()` methods) and does what they want with it (we can document permuteddims view for them), or
- (b) views it as an "image" object similar to how it might be dealt with in the JuliaImages organization (through ongoing work in the [images](https://github.com/yeesian/ArchGDAL.jl/tree/images) branch to support `imread()`/`imwrite()`). This handles issues like permuting the array, normalizing the values, and identifying the color channels.

Being stuck in the middle ground between (a) and (b) sounds like an easy way to cause confusion for users.

PR #61, opened 2018-10-10
add checksum to downloaded test files

### LINKED ISSUE #58: Check for corrupt downloads while testing.
Taken from and closes #50:

> Maybe we should replace the isfile with a checksum in case there are corrupted files. In julia 0.7 SHA is in the stdlib.

visr:
Uses `download_verify` from BinaryProvider.
Since BinaryProvider is already needed for GDAL, I assume it's fine to depend on for the tests.

Fixes #58 

If it is downloading, it logs what it is downloading. This is probably useful to inform users why it is taking so long, or for recognizing downloading issues. Otherwise the `quiet_download` keyword could still be added.

https://github.com/JuliaPackaging/BinaryProvider.jl/blob/54b5a7822a280bbc53d803a23e6d6902fa8090ac/src/PlatformEngines.jl#L524-L526

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,0 +1,1 @@
+BinaryProvider
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,65 +1,68 @@
 using Test
 using Dates
+using BinaryProvider
 
-const testdatadir = dirname(@__FILE__)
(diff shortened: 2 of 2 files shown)


yeesian:
This is great, thanks!

PR #62, opened 2018-10-14
Disable installing/using GDAL explicitely

evetion:
as to mimic Windows user running `]test ArchGDAL`.

It seems that for Windows:
- Setting GDAL_DATA using `CPLSetConfigOption` is required
- This config is lost by calling `@testset`. You can check this by calling `AG.getconfigoption("GDAL_DATA")`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ build_script:
 
 test_script:
   - echo "%JL_TEST_SCRIPT%"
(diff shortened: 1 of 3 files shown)


evetion:
I believe that the call to `CPLSetConfigOption` should be in the same leaf testset. I don't understand the behaviour, looking at the `@testset` source (https://github.com/JuliaLang/julia/blob/0d713926f85dfa3e4e0962215b909b8e47e94f48/stdlib/Test/src/Test.jl#L987-L1033) maybe the whole `quote` thing shuffles thing.
 
In the end we should try to understand what happens with setting GDAL config option in a thread within Julia. Feel free to make a Discourse post.

PR #65, opened 2018-10-16
fix #64; define wkbUnknown

### LINKED ISSUE #64: wkbUnknown not defined when creating layer with default geom?
Here is the code that should reproduce the error.
```julia
ArchGDAL.registerdrivers() do
    drv = ArchGDAL.getdriver("ESRI Shapefile")
    ArchGDAL.create("test",drv) do ds

        layer = ArchGDAL.createlayer(ds,"test") # UndefVarError: wkbUnknown not defined
        # layer = ArchGDAL.createlayer(ds,"test", geom=ArchGDAL.GDAL.wkbUnknown) # This seems to  work
        ...
    end
end
```
the problem seems to be at `dataset.jl` [line 315]([https://github.com/yeesian/ArchGDAL.jl/blob/a1f167dec845df058d36e650f5e871eeb40a1e1c/src/dataset.jl#L315])

cchderrick:


visr:
Great, thanks!

yeesian:
Thanks @cchderrick! :)

Could you also add a small test (in a new PR), similar to the one in #64 

```julia
ArchGDAL.registerdrivers() do
    drv = ArchGDAL.getdriver("ESRI Shapefile")
    ArchGDAL.create("test",drv) do ds

        layer = ArchGDAL.createlayer(ds,"test") # UndefVarError: wkbUnknown not defined
        # layer = ArchGDAL.createlayer(ds,"test", geom=ArchGDAL.GDAL.wkbUnknown) # This seems to  work
        ...
    end
end
```

to [`test_dataset.jl`](https://github.com/yeesian/ArchGDAL.jl/blob/master/test/test_dataset.jl) ? That way we'll have some coverage to guard against similar issues in the future.

PR #67, opened 2018-11-04
Fixes issue 66

yeesian:


yeesian:
Any idea what's up with the coverage measured here?

visr:
Hmm seems like https://github.com/JuliaCI/Coverage.jl/issues/187, so one way to fix it for now is to run coverage with `--inline=no`, though it will be slower.

yeesian:
Ah gotcha. Let's just leave it for now, unless there's a PR for which we want to ensure good coverage

PR #85, opened 2019-08-13
Update to new GDAL.jl API

yeesian:
ref https://github.com/JuliaGeo/GDAL.jl/pull/73


yeesian:
> Looks good to me. Next to big task of renaming, you brought along a nice set of small code tweaks, making it more explicit and taking apart some logic. I approve :)
> 
> Of course in GDAL.jl everything is now `Ptr{Cvoid}`. Does that mean that types.jl needs an update now?
> 
> If I take `FeatureLayer` as an example, we have in GDAL.jl:
> 
> ```julia
> "Opaque type for a layer (OGRLayer)"
> const OGRLayerH = Ptr{Cvoid}
> ```
> 
> And in ArchGDAL.jl:
> 
> ```julia
> const GDALFeatureLayer = Ptr{GDAL.OGRLayerH}
> 
> mutable struct FeatureLayer <: AbstractFeatureLayer
> 	ptr::GDALFeatureLayer
> end
> 
> mutable struct IFeatureLayer <: AbstractFeatureLayer
> 	ptr::GDALFeatureLayer
> 	ownedby::AbstractDataset
> 	spatialref::AbstractSpatialRef
> 	
> 	function IFeatureLayer(
> 			ptr = GDALFeatureLayer(C_NULL);
> 			ownedby::AbstractDataset = Dataset(),
> 			spatialref::AbstractSpatialRef = SpatialRef()
> 		)
> 		layer = new(ptr, ownedby, spatialref)
> 		finalizer(destroy, layer)
> 		return layer
> 	end
> end
> ```
> 
> So that looks to me like we have one too many `Ptr`, i.e. `Ptr{Ptr{Cvoid}}`. Or am I missing something?

Yup you're right, thanks! Addressed in https://github.com/yeesian/ArchGDAL.jl/pull/85/commits/eba399ab023d5577315d51a06d2c0da4527daae5

PR #86, opened 2019-08-19
Migrated to Pkg3. Pinned GDAL.

evetion:
Also upped version to 0.2.1.

This will prevent breaking ArchGDAL when JuliaGEO/GDAL.jl#73 is merged.


yeesian:
Thanks!

PR #87, opened 2019-08-24
aftercare, update for GDAL.jl#75

visr:
Removes the `failsafe` calls in dataset.jl. Don't think it is needed. If you put in a null pointer, GDAL will tell you.

And the `usage_error` should now also already be thrown by GDAL.jl.


yeesian:
Thanks!

PR #93, opened 2019-10-21
Update documentation on order of dimensions.

### LINKED ISSUE #37: transposed rasters

evetion:
Fixes #37

### DIFF
--- a/docs/src/rasters.md
+++ b/docs/src/rasters.md
@@ -66,9 +66,9 @@ The general operative method for reading in raster values from a `dataset` or `b
 
 !!! note
 
-    The array returned by `read` has `(rows, cols, bands)` dimensions. 
+    The array returned by `read` has `(cols, rows, bands)` dimensions. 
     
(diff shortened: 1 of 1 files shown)


visr:
Thanks! Merged onto master, and directly rebased the idataset branch on top of that.

PR #96, opened 2019-11-08
import GeoInterface methods explicitly

rafaqz:
So that `ArchGDAL.coordinates` works. Without this GeoInterface also needs to be imported to use `coordinates`, which makes a few things like using Requires.jl more complicated.


yeesian:
Thanks!

Issue #83, opened 2019-08-01
Follow a style guide


yeesian:
* [documentation](https://docs.julialang.org/en/v1/manual/documentation/index.html)

yeesian:
http://www.juliaopt.org/JuMP.jl/dev/style/#Style-guide-and-design-principles-1

PR #97, opened 2019-11-13
WIP: Geo format type wrappers and generic methods

rafaqz:
No tests yet, but things like `reproject` do work for any format types, which is pretty powerful allready. 

I'm not really familiar with the details of these formats or GDAL methods, so input on how to actually test these things and finish the PR will be needed. 

My intent (and knowledge) is more around generalising the structure of these package and the ecosystem so we can do all kinds of transformations without needing all the method permutations everywhere. 

For GeoData.jl this PR allows us to do `GDALarray(filepath; reproject=EPSGcode(4326))` where you can pass in anything for reproject and GeoData doesn't need to know what it is, but your lat/lon coords will just work.

The type wrappers are coming from https://github.com/rafaqz/GeoFormatTypes.jl

### DIFF
--- a/src/ogr/geometry.jl
+++ b/src/ogr/geometry.jl
@@ -1,3 +1,32 @@
+
+# This is friendly type-pyracy on `convert` for GeoFormatTypes types
+Base.convert(::Type{<:Geometry}, source::GFT.AbstractWellKnownText) = 
+    fromWKT(GFT.val(source))
+Base.convert(::Type{<:Geometry}, source::GFT.WellKnownBinary) = 
+    fromWKB(GFT.val(source))
+Base.convert(::Type{<:Geometry}, source::GFT.GeoJSON) = 
+    fromJSON(GFT.val(source))
+Base.convert(::Type{<:Geometry}, source::GFT.GML) = 
+    fromGML(GFT.val(source))
+
+Base.convert(::Type{<:GFT.AbstractWellKnownText}, source::AbstractGeometry) = 
+    GFT.WellKnownText(GFT.Geom(), toWKT(source))
+Base.convert(::Type{<:GFT.WellKnownBinary}, source::AbstractGeometry) = 
+    GFT.WellKnownBinary(GFT.Geom(), toWKB(source))
+Base.convert(::Type{<:GFT.GeoJSON}, source::AbstractGeometry) = 
+    GFT.GeoJSON(toJSON(source))
+Base.convert(::Type{<:GFT.GML}, source::AbstractGeometry) = 
+    GFT.GML(GFT.Geom(), toGML(source))
+Base.convert(::Type{<:GFT.KML}, source::AbstractGeometry) = 
+    GFT.KML(toKML(source))
+
+# Convert to Geometry, then to the target format. There may be
+# a more optimal method, but this is a simple, high-level first pass.
+# The Geom trait is needed to separate out convert for CRS for WellKnownText
+# and GML, which may contain both. It is handled in GeoFormatTypes.
+Base.convert(format::Type{<:GFT.GeoFormat}, mode::GFT.Geom, source::GFT.GeoFormat) =
+    convert(format, convert(Geometry, source))
+
 """
     fromWKB(data)
 
--- a/src/ogr/geometry.jl
+++ b/src/ogr/geometry.jl
@@ -1,32 +1,4 @@
 
-# This is friendly type-pyracy on `convert` for GeoFormatTypes types
-Base.convert(::Type{<:Geometry}, source::GFT.AbstractWellKnownText) = 
-    fromWKT(GFT.val(source))
-Base.convert(::Type{<:Geometry}, source::GFT.WellKnownBinary) = 
-    fromWKB(GFT.val(source))
-Base.convert(::Type{<:Geometry}, source::GFT.GeoJSON) = 
-    fromJSON(GFT.val(source))
-Base.convert(::Type{<:Geometry}, source::GFT.GML) = 
-    fromGML(GFT.val(source))
-
-Base.convert(::Type{<:GFT.AbstractWellKnownText}, source::AbstractGeometry) = 
-    GFT.WellKnownText(GFT.Geom(), toWKT(source))
-Base.convert(::Type{<:GFT.WellKnownBinary}, source::AbstractGeometry) = 
-    GFT.WellKnownBinary(GFT.Geom(), toWKB(source))
-Base.convert(::Type{<:GFT.GeoJSON}, source::AbstractGeometry) = 
-    GFT.GeoJSON(toJSON(source))
-Base.convert(::Type{<:GFT.GML}, source::AbstractGeometry) = 
-    GFT.GML(GFT.Geom(), toGML(source))
-Base.convert(::Type{<:GFT.KML}, source::AbstractGeometry) = 
-    GFT.KML(toKML(source))
-
-# Convert to Geometry, then to the target format. There may be
-# a more optimal method, but this is a simple, high-level first pass.
-# The Geom trait is needed to separate out convert for CRS for WellKnownText
-# and GML, which may contain both. It is handled in GeoFormatTypes.
-Base.convert(format::Type{<:GFT.GeoFormat}, mode::GFT.Geom, source::GFT.GeoFormat) =
-    convert(format, convert(Geometry, source))
-
 """
     fromWKB(data)
 
--- a/src/spatialref.jl
+++ b/src/spatialref.jl
@@ -1,16 +1,21 @@
 """
 Import format and run function `f` on the result.
 """
-importCRS(f::Function, sourceformat::GFT.GeoFormat) = begin
-    obj = importCRS(sourceformat)
-    try f(obj) finally destroy(obj) end
-end
-importCRS(x::GFT.EPSG) = importEPSG(GFT.val(x))
-importCRS(x::GFT.AbstractWellKnownText) = importWKT(GFT.val(x))
-importCRS(x::GFT.ESRIWellKnownText) = importESRI(GFT.val(x))
-importCRS(x::GFT.ProjString) = importPROJ4(GFT.val(x))
-importCRS(x::GFT.GML) = importXML(GFT.val(x))
-importCRS(x::GFT.KML) = importCRS(EPSG(4326))
+importCRS(x::GFT.GeoFormat) = importCRS!(newspatialref(), x)
+unsafe_importCRS(x::GFT.GeoFormat) = importCRS!(unsafe_newspatialref(), x)
+
+importCRS!(spref::AbstractSpatialRef, x::GFT.EPSG) =
+    importEPSG!(spref, GFT.val(x))
+importCRS!(spref::AbstractSpatialRef, x::GFT.AbstractWellKnownText) =
+    importWKT!(spref, GFT.val(x))
+importCRS!(spref::AbstractSpatialRef, x::GFT.ESRIWellKnownText) =
+    importESRI!(spref, GFT.val(x))
+importCRS!(spref::AbstractSpatialRef, x::GFT.ProjString) =
+    importPROJ4!(spref, GFT.val(x))
+importCRS!(spref::AbstractSpatialRef, x::GFT.GML) =
+    importXML!(spref, GFT.val(x))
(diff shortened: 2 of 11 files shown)


yeesian:
> I'm not really familiar with the details of these formats or GDAL methods, so input on how to actually test these things and finish the PR will be needed.

To my knowledge, the spatial projection tests that we have are defined in the following files:
* https://github.com/yeesian/ArchGDAL.jl/blob/master/test/test_cookbook_projection.jl, and
* scattered in comments in https://github.com/yeesian/ArchGDAL.jl/blob/master/test/test_ospy_examples.jl

rafaqz:
Ok this is is tested, it just has to wait for the new version of GeoFormatTypes to go into the registry for tests to pass.

`convert` converts geometries or crs between most types in GeoFormatTypes. The `mode::Geom` and `mode::CRS` traits are used to handle multi-use formats like `WellKnownText` or `GML`. They will mostly be picked automatically in GeoFormatTypes.

`reproject` accepts any crs formats to generate the transform, and any GeoFormatTypes or AbstractGeometry geometry. It also accepts tuple/vector coord formats (that could be cleaned up and standardised to fit GeoInterface etc at some point).

I'll rebase it against the other pull request once that's in, but otherwise this is done, so comments would be useful.

rafaqz:
This is rebased and passing now.

src/ogr/geometry.jl
```diff
@@ -1,3 +1,32 @@
...
+    GFT.GML(GFT.Geom(), toGML(source))
+Base.convert(::Type{<:GFT.KML}, source::AbstractGeometry) = 
+    GFT.KML(toKML(source))
+
+# Convert to Geometry, then to the target format. There may be
+# a more optimal method, but this is a simple, high-level first pass.
+# The Geom trait is needed to separate out convert for CRS for WellKnownText
+# and GML, which may contain both. It is handled in GeoFormatTypes.
+Base.convert(format::Type{<:GFT.GeoFormat}, mode::GFT.Geom, source::GFT.GeoFormat) =
+    convert(format, convert(Geometry, source))
```
yeesian: Can we move these methods into a `conversions.jl` file under [`src/base`](https://github.com/yeesian/ArchGDAL.jl/tree/master/src/base)?

src/spatialref.jl
```diff
@@ -1,3 +1,85 @@
+"""
+Import format and run function `f` on the result.
+
+The name could be improved?
+"""
+importcrs(f::Function, sourceformat::GFT.GeoFormat) = begin
```
yeesian: Please write this as
```julia
function importcrs(f::Function, sourceformat::GFT.GeoFormat)
    obj = unsafe_importcrs(sourceformat)
    try f(obj) finally destroy(obj) end
end
```
The way to do it is to add `importcrs` to [this line](https://github.com/yeesian/ArchGDAL.jl/blob/fe9ee2a312eafac9d00f9f43100b0dd7f5a005e6/src/context.jl#L205). All the methods below it should be renamed from 
```
importcrs(x) = import*(GFT.val(x))
```
to
```
unsafe_importcrs(x) = unsafe_import*(GFT.val(x))
```
(Otherwise the spatialref object that is created will be destroyed twice. It will also make it easier to write a version that doesn't require you to wrap it in a do-block. If you're interested, see [this section of the docs](http://yeesian.com/ArchGDAL.jl/latest/memory/#Memory-Management-(Advanced)-1) for an explanation.)

test/test_geometry.jl
```diff
@@ -95,6 +98,25 @@ end
     @test AG.toJSON(point) == "{ \"type\": \"Point\", \"coordinates\": [ 100.0, 70.0 ] }"
 end
 
+
+
```
yeesian: Let's have just 1 newline between each testset.

test/test_geometry.jl
```diff
@@ -1,5 +1,8 @@
 using Test
-import GeoInterface, GDAL, ArchGDAL; const AG = ArchGDAL
+import GeoInterface, GeoFormatTypes, GDAL, ArchGDAL;
+const AG = ArchGDAL
+const GFT = GeoFormatTypes
+
```
yeesian: remove this line.

test/test_geometry.jl
```diff
@@ -105,6 +127,8 @@ end
         @test AG.toWKT(geom) == "LINESTRING (1 4,2 5,3 6)"
         AG.setpoint!(geom, 1, 10, 10)
         @test AG.toWKT(geom) == "LINESTRING (1 4,10 10,3 6)"
+        # Test using convert
```
yeesian: is this comment helpful?

test/test_geometry.jl
```diff
@@ -95,6 +98,25 @@ end
...
+    wkb = convert(GFT.WellKnownBinary, point) 
+    wkt = convert(GFT.WellKnownText, point) 
+    @test GFT.val(json) == AG.toJSON(point)
+    @test GFT.val(kml) == AG.toKML(point)
+    @test GFT.val(gml) == AG.toGML(point) # This isn't actually tested above
+    @test GFT.val(wkb) == AG.toWKB(point)
+    @test GFT.val(wkt) == AG.toWKT(point)
+    @test convert(GFT.GeoJSON, json) == json
+    @test convert(GFT.GeoJSON, wkb) == convert(GFT.GeoJSON, wkt) == convert(GFT.GeoJSON, gml)
+    @test convert(GFT.KML, gml) == convert(GFT.KML, wkt)
```
yeesian: Can we move all the conversion tests into its own testset? So maybe something like

```julia
@testset "format conversions" begin
    @testset "point geometry" begin
        ...
    end

    @testset "complex geometries" begin
        ...
    end
end
```

src/spatialref.jl
```diff
@@ -1,3 +1,85 @@
...
+    obj = importcrs(sourceformat)
+    try f(obj) finally destroy(obj) end
+end
+importcrs(x::GFT.EPSG) = importEPSG(GFT.val(x))
+importcrs(x::GFT.AbstractWellKnownText) = importWKT(GFT.val(x))
+importcrs(x::GFT.ESRIWellKnownText) = importESRI(GFT.val(x))
+importcrs(x::GFT.ProjString) = importPROJ4(GFT.val(x))
+importcrs(x::GFT.GML) = importXML(GFT.val(x))
+importcrs(x::GFT.KML) = importEPSG(EPSG(4327))
+importcrs(x::GFT.GeoJSON) = importEPSG(EPSG(4327)) # TODO check GeoJSON standard
```
yeesian: Please add unit tests as well.

src/spatialref.jl
```diff
@@ -1,3 +1,85 @@
+"""
+Import format and run function `f` on the result.
+
+The name could be improved?
```
yeesian: I think [`newspatialref`](https://github.com/yeesian/ArchGDAL.jl/blob/fe9ee2a312eafac9d00f9f43100b0dd7f5a005e6/src/spatialref.jl#L6-L7) could work here. The docstring will then be something analogous like `"Construct a Spatial Reference System from a GeoFormat."` What do you think?

src/spatialref.jl
```diff
@@ -1,3 +1,85 @@
...
+    end
+
+"""
+    crs2transform(f, crs::GeoFormat, targetcrs::GeoFormat)
+
+Run the function on a coord transform generated from
+two crs definitions. These can be in any GeoFormatTypes format
+that holds a coordinate reference system.
+"""
+crs2transform(f, crs::GFT.GeoFormat, targetcrs::GFT.GeoFormat) =
```
yeesian: `f::Function`? Do we accept any other types of `f`?

src/spatialref.jl
```diff
@@ -1,3 +1,85 @@
...
+    GeoInterface.coordinates(reproject(createpoint(coord...), crs, targetcrs))
+reproject(coords::AbstractArray{<:ReprojectCoord}, crs::GFT.GeoFormat, targetcrs::GFT.GeoFormat) =
+    GeoInterface.coordinates.(reproject([createpoint(c...) for c in coords], crs, targetcrs))
+
+# GeoFormat
+reproject(geom::GFT.GeoFormat, crs::GFT.GeoFormat, targetcrs::GFT.GeoFormat) =
+    convert(typeof(geom), reproject(convert(Geometry, geom), crs, targetcrs))
+
+# Geometries
+reproject(geom::AbstractGeometry, crs::GFT.GeoFormat, targetcrs::GFT.GeoFormat) =
```
yeesian: It's my own subjective take on it, but this is not a single-line function and should be re-written as 
```julia
function reproject(geom::AbstractGeometry, crs::GFT.GeoFormat, targetcrs::GFT.GeoFormat)
    crs2transform(crs, targetcrs) do transform
        transform!(geom, transform)
    end
end
```
I am developing a style guide for the future: https://github.com/yeesian/ArchGDAL.jl/issues/83.

src/spatialref.jl
```diff
@@ -1,3 +1,85 @@
...
+
+unsafe_convert(::Type{GFT.CoordSys}, mode::GFT.CRS, srs::Ref) = GFT.CoordSys(toMICoordSys(srs))
+unsafe_convert(::Type{GFT.ProjString}, mode::GFT.CRS, srs::Ref) = GFT.ProjString(toPROJ4(srs))
+unsafe_convert(::Type{GFT.WellKnownText}, mode::GFT.CRS, srs::Ref) = GFT.WellKnownText(toWKT(srs))
+unsafe_convert(::Type{GFT.ESRIWellKnownText}, mode::GFT.CRS, srs::Ref) =
+    GFT.ESRIWellKnownText(toWKT(morphtoESRI!(srs)))
+unsafe_convert(::Type{GFT.GML}, mode::GFT.CRS, srs::Ref) = GFT.GML(toXML(srs))
+
+# These should be better integrated with geometry packages or follow some standard
+const ReprojectCoord = Union{<:NTuple{2,<:Number},<:NTuple{3,<:Number},AbstractVector{<:Number}}
```
yeesian: please define this directly above where it gets used; e.g., L47 of this file in its current form.

src/ogr/geometry.jl
yeesian: You should make the comment above the top-level comment for that file, and explain it to developers in a way that doesn't require an understanding of GeoFormatTypes.

rafaqz: conversions.jl is a good idea.

src/spatialref.jl
rafaqz: I try to avoid breaking the flow by switching method layout between `f(x) =` and `function f(x)` all the time. I think it's the ugliest thing in Julia lol. But your call.

src/spatialref.jl
rafaqz: Functors?

test/test_geometry.jl
rafaqz: if we have a conversions.jl file we could have a test_converstions.jl to test them in

src/spatialref.jl
rafaqz: Oh right. You should document/comment that context code it's pretty hard to understand what you are doing there but important to how the package works.

rafaqz:
Ok I think this is pretty much finished, sorry for the delay! I've been using it for a few months in GeoData.jl and it's working well.

src/spatialref.jl
rafaqz: fixed

src/spatialref.jl
rafaqz: its ::Function now

src/spatialref.jl
rafaqz: This didn't turn out to be workable

test/test_geometry.jl
rafaqz: deleted

src/spatialref.jl
rafaqz: Done

src/ogr/geometry.jl
rafaqz: this is all in convert.jl, not sure why I didn't call it conversions.jl lol

yeesian:
It looks great, thanks a lot! :)

rafaqz:
Thanks! I can relegister GeoData.jl whenever this is registered.

visr:
@rafaqz it will be registered in a few minutes: https://github.com/JuliaRegistries/General/pull/12666

PR #100, opened 2019-12-05
Array interface

rafaqz:
Tests aren't included yet, but this works great.

I'm not sure how much to support, but we could also define `setindex!` that calls `write`

### DIFF
--- a/src/raster/array.jl
+++ b/src/raster/array.jl
@@ -10,8 +10,8 @@ Base.getindex(dataset::Dataset, x::AllowedXY, y::AllowedXY, bands::AllowedBand) 
     # Convert colons to ranges
     x, y, bands = map(convertcolon, (x, y, bands), size(dataset))
     buffer = Array{getdatatype(getband(dataset, 1)),3}(undef, length.((x, y, bands))...)
-    xoffset, xsize = first(x), length(x)
-    yoffset, ysize = first(y), length(y)
+    xoffset, yoffset = first(x) - 1, first(y) - 1
+    xsize, ysize = length(x), length(y)
     indices = bandindices(bands)
     read!(dataset, buffer, indices, xoffset, yoffset, xsize, ysize)
     # Drop dimensions of integer indices
@@ -22,7 +22,7 @@ Base.getindex(dataset::Dataset, x::Int, y::Int, band::Int) = begin
     indices = bandindices(band)
     buffer = Array{getdatatype(getband(dataset, 1)),3}(undef, 1, 1, 1)
     xsize = ysize = 1
-    read!(dataset, buffer, indices, x, y, xsize, ysize)
+    read!(dataset, buffer, indices, x - 1, y - 1, xsize, ysize)
     # Indexing with all ints returns a scalar
     buffer[1]
 end
@@ -34,5 +34,5 @@ dropint() = ()
 convertcolon(i::Colon, sze) = 1:sze
 convertcolon(i, sze) = i
 
-bandindices(band::Int) = Cint[x]
+bandindices(band::Int) = Cint[band]
 bandindices(bands::AbstractArray) = Cint[b for b in bands]
--- a/src/raster/array.jl
+++ b/src/raster/array.jl
@@ -1,38 +1,90 @@
-Base.size(dataset::Dataset) = begin
-    band = getband(dataset, 1)
-    width(band), height(band), nraster(dataset)
-end
-
 const AllowedXY = Union{Integer,Colon,AbstractRange}
 const AllowedBand = Union{Integer,Colon,AbstractArray}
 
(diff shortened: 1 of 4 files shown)


rafaqz:
Ok this is tested now. I'll leave `setindex` to another pull request.

src/raster/array.jl
```diff
@@ -0,0 +1,38 @@
...
+    buffer = Array{getdatatype(getband(dataset, 1)),3}(undef, length.((x, y, bands))...)
+    xoffset, yoffset = first(x) - 1, first(y) - 1
+    xsize, ysize = length(x), length(y)
+    indices = bandindices(bands)
+    read!(dataset, buffer, indices, xoffset, yoffset, xsize, ysize)
+    # Drop dimensions of integer indices
+    reshape(buffer, dropint(x, y, bands)...)
+end
+
+Base.getindex(dataset::Dataset, x::Int, y::Int, band::Int) = begin
```
yeesian: is there a reason the `::Int` are not `::Integer`s here?

src/raster/array.jl
```diff
@@ -0,0 +1,38 @@
...
+    indices = bandindices(band)
```
yeesian: I lean towards keeping this `indices = Cint[band]`

src/raster/array.jl
```diff
@@ -0,0 +1,38 @@
+Base.size(dataset::Dataset) = begin
+    band = getband(dataset, 1)
+    width(band), height(band), nraster(dataset)
+end
+
+const AllowedXY = Union{Integer,Colon,AbstractRange}
+const AllowedBand = Union{Integer,Colon,AbstractArray}
+
+Base.getindex(dataset::Dataset, x::AllowedXY, y::AllowedXY, bands::AllowedBand) = begin
```
yeesian: Can we keep it `bands::Union{Integer,Colon,AbstractArray}`? Do we foresee `AllowedBand` being used more broadly in the codebase?

yeesian: similarly, inlining `AllowedXY` and removing the const definition.

src/raster/array.jl
```diff
@@ -0,0 +1,38 @@
...
+
+dropint(x::Integer, args...) = dropint(args...)
+dropint(x::AbstractRange, args...) = (length(x), dropint(args...)...)
+dropint() = ()
+
+convertcolon(i::Colon, sze) = 1:sze
+convertcolon(i, sze) = i
+
+bandindices(band::Int) = Cint[band]
+bandindices(bands::AbstractArray) = Cint[b for b in bands]
```
yeesian: I'd prefer to move these definitions of `bandindices` into `Base.getindex(dataset::Dataset, x::AllowedXY, y::AllowedXY, bands::AllowedBand)`, since they do not have docstrings and don't seem to be meant for general consumption

yeesian: Similarly with `dropint` and `convertcolon`

src/raster/array.jl
rafaqz: Yeah that's not needed here, I had just copied and pasted from the other method where array and range are possible inputs.

src/raster/array.jl
rafaqz: I was making it very explicit that band and x/y actually have different allowed input. It will be used again for `setindex!` when we write that.

src/raster/array.jl
rafaqz: Base uses `Int`:
https://github.com/JuliaLang/julia/blob/master/base/array.jl#L758

But I'm not sure why.

src/raster/array.jl
rafaqz: They are just utility methods for this file. But they would also be used in `setindex!`, so they were defined outside `getindex`. Maybe I should just write `setindex` too to make that clear lol, I just didn't have time to test it, but I'll try to find some.

Its pretty ok to have undocumented unexported methods, base does it everywhere.

rafaqz:
I guess `RasterBand` could have the same without the band index? That should be fine I just don't use `read` on `RasterBand` much so didn't think of it.

rafaqz:
Ok I've just pushed handling of `RasterBand` and `setindex!`, and tested it all.

rafaqz:
Any thoughts on this latest iteration? would be goo to get it merged soon.

My only reservation is that it would be good to make it an actual `AbstractArray`, but that can create a bunch of issues with show and broadcast default methods.

yeesian:
Thanks!

yeesian:
I don't know about making it an `AbtractArray` since the `Dataset` type is also used for vector datasets.

I view these more as convenience methods that I've held off for a really long time on (I think @visr will be able to testify), and I prefer for people to use and contribute to packages like GeoArrays.jl for more advanced use cases.

PR #102, opened 2019-12-16
add an Array constructor

rafaqz:
Adds an `Array` constructor for `Dataset` to avoid having to call the `AG.read` and special case Ag.Dataset. Somehow I deleted this in the other pull request.


visr:
Closing this now, since it was incorporated in #76, and is released as ArchGDAL 0.3.

PR #103, opened 2019-12-27
add docstrings to docs

### LINKED ISSUE #92: No docstrings in docs
```
┌ Warning: 437 docstrings potentially missing:
│
│     ArchGDAL.indexof :: Tuple{ArchGDAL.AbstractRasterBand}
│     ArchGDAL.unsafe_nextfeature :: Tuple{ArchGDAL.AbstractFeatureLayer}
│     ArchGDAL.convexhull :: Tuple{ArchGDAL.AbstractGeometry}
│     ArchGDAL.setcoorddim! :: Tuple{ArchGDAL.AbstractGeometry,Integer}
│     ArchGDAL.getgeomdefn :: Union{Tuple{ArchGDAL.FeatureDefn}, Tuple{ArchGDAL.FeatureDefn,Integer}}
│     ArchGDAL.getgeomdefn :: Tuple{ArchGDAL.Feature,Integer}
│     ArchGDAL.wkbsize :: Tuple{ArchGDAL.AbstractGeometry}
```

To take the first example, `indexof ` has a fine docstring:
```julia
"""
Fetch the band number (1+) within its dataset, or 0 if unknown.

This method may return a value of 0 to indicate overviews, or free-standing
`GDALRasterBand` objects without a relationship to a dataset.
"""
indexof(band::AbstractRasterBand) = GDAL.gdalgetbandnumber(band.ptr)
```

But since this function is not included in the docs anywhere with an [`@docs`](https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Adding-Some-Docstrings-1) tag, this docstring doesn't end up in the docs. In fact `@docs` is used nowhere in the docs.

`indexof` is listed twice in two different pages, with a description based on the docstring:

- `i = ArchGDAL.indexof(band)`: the **index** of the raster band.
- `ArchGDAL.indexof(band)`: the index of the band (1+) within its dataset, or 0 if unknown. (`1`)

But it's better to put the `@docs` in one location, and then use [cross referencing](https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Cross-Referencing-1) from other parts.

visr:
Will fix #92

~~Right now it doesn't add docstrings to the docs yet, but has some related doc improvements.~~

For docstrings I am still looking a bit what to do.

Either we can add an API reference page, with all the docstrings, split into headers following roughly the rest of the docs (features/rasters/projections). Alternatively I can try to embed the docstrings in the existing pages, suggestions for which is better are welcome. If we create a API reference page, I would still add references from the usage pages towards the API reference.

I tried locally to use `autodocs`, and find that some docstrings are effectively duplicates. For instance there is `setfield!(feature::Feature, i::Integer, value::Cint)` and `setfield!(feature::Feature, i::Integer, value::Int64)`. We need both since they go to different C functions. But right now it has almost the same docstring duplicated for each type of value. We could create one docstring instead, listing the different types of value we support.

We should probably also add the call signature to the docstring as is commonly done. This also results in clearer documentation. See also the [manual page on documentation](https://docs.julialang.org/en/latest/manual/documentation/#).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,6 @@ jobs:
       script:
         - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
                                                Pkg.instantiate()'
+        - julia --project=docs/ test/remotefiles.jl
         - julia --project=docs/ docs/make.jl
       after_success: skip
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,6 @@
 [deps]
 ArchGDAL = "c9ce4bd3-c3d5-55b8-8973-c0e20141b8c3"
+BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 
 [compat]
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -11,11 +11,12 @@ makedocs(
         "Raster Data" => "rasters.md",
         "Geometric Operations" => "geometries.md",
         "Spatial Projections" => "projections.md",
-        "Interactive versus Scoped Objects" => "memory.md"
-        # "Working with Spatialite" => "spatialite.md"
+        # "Working with Spatialite" => "spatialite.md",
+        "Interactive versus Scoped Objects" => "memory.md",
+        "Design Considerations" => "considerations.md",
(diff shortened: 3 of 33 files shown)


visr:
Ok I added method signatures to almost all docstrings, and are listing them on a new API Reference page in the documentation. I added cross references from the existing docs to the API Reference for individual functions.

This should be good to go.

PR #105, opened 2020-01-15
RFC: Implementation of DiskArray interface

meggart:
As discussed [here](https://github.com/meggart/DiskArrays.jl/issues/2) there was the suggestion to implement the DiskArray interface for the raster types in ArchGDAL. For a single RasterBand this was straightforward. However I had to jump through some hoops to make it possible to treat a whole Dataset as a AbstractDiskArray. The major problems were:

- data type is not guaranteed to be constant across bands
- band size is not guaranteed to be constant across bands
- there are datasets that do not contain any raster data (like all vector data)

So I decided to make a new Dataset wrapper (RasterDataset) which does some checks during construction and then wraps a Dataset into a new AbstractDiskArray. I just tested this and the following functionality would come for free (plus everything else we decide to put on top of DiskArrays.jl)

- efficient getindex/setindex for Bands and Datasets
- efficient reductions over blocks that are defined by `windows`
- reductions over dimensions

### Some examples that I tried

````julia
r = AG.read("/home/fgans/julia_dev/ArchGDAL/test/ospy/data4/aster.img") do ds
  b = AG.getband(ds,2)
  mean(b)
end
````

````julia
r = AG.readraster("/home/fgans/julia_dev/ArchGDAL/test/ospy/data4/aster.img") do ds
  r2 = view(ds,1:1000,1:1000,:)
  r2[20:22,20:22,1]
end
````

````julia
r = AG.readraster("/home/fgans/julia_dev/ArchGDAL/test/ospy/data4/aster.img") do ds
  maximum(ds,dims=(1,2))
end
````

@rafaqz I was confused by a few tests in test_array.jl because it looks like you allowed to omit the last index when accessing a Dataset, even if the number of bands is greater than one. I marked these tests as broken here, what was the rationale behind this behavior?

### CODE REVIEW

**Comment by rafaqz** [comment]
I think the underlying `read` and `write` methods allow treating the array as 2d like that. But you are right that shouldn't be allowed in the array interface. 

But I'm not sure how bands should be treated really. Is a single band 2d or still 3d? In GeoData.jl you always get a 3d array with a `Band` dim no matter how many bands there are, for consistency.

**Comment by meggart** [comment]
> I think the underlying read and write methods allow treating the array as 2d like that. But you are right that shouldn't be allowed in the array interface.

Thanks for the clarification. For now I removed the tests that were marked as broken

> But I'm not sure how bands should be treated really. Is a single band raster 2d or still 3d? In GeoData.jl you always get a 3d array with a Band dim no matter how many bands there are, for consistency.

I tend to go for consistency as well, the number of dimensions of the array should always be predictable. Since I already introduced a new function name (maybe @yeesian can comment if this is ok), should we make a distinction between `readraster` and `readstack`, where `readraster` opens a dataset and returns a handle to the first band while `readstack` returns a 3D-handle to all bands? Or is this too R-rish...

**Comment by felixcremer** [comment]
I tried to use this implementation as input to an ESDL data cube see https://github.com/esa-esdl/ESDL.jl/pull/193. 
At the moment the functions which use the pointer to apply GDAL functions doesn't work. 
For me it was the getgeotransform and the nraster functions. 

Could we implement a nicer way to get an interactive RasterDataset? 
I would suggest, that readraster("path/to/file") would return the RasterDataset, because this would make it very nice for interactive work. 
 At the moment I can get this with 
```julia
AG.RasterDataset(AG.read("/path/to/file"))
```

I don't like the name readraster, if this returns single bands. I would suggest the name readband and to always expect an Integer to indicate which band should be read.

meggart:
So I updated the PR according to the comments. @yeesian maybe you can give a short comment if you would be interested in this PR?

meggart:
To be more specific, AG.readraster("raster.tif") now returns a wrapped IDataset and I added forwarding for a bunch of methods like width, height etc.

src/raster/array.jl
```diff
@@ -1,89 +1,106 @@
+import DiskArrays: eachchunk, GridChunks
+import DiskArrays
 const AllowedXY = Union{Integer,Colon,AbstractRange}
 const AllowedBand = Union{Integer,Colon,AbstractArray}
 
 
+# Define a RasterDataset type
```
yeesian, line 7: I'm going by the reactions in https://github.com/meggart/DiskArrays.jl/issues/2#issue-549445915 to guess that there is general acceptance of DiskArrays as an interface: in that case, it will be good to elaborate on what a RasterDataset is -- its correspondence to the GDAL Data Model, and how it relates to e.g. Raster[Band] versus as a Dataset, so that the discussion and conceptual models are contained in the code without being tied to github.

src/raster/array.jl
```diff
@@ -1,89 +1,106 @@
...
+struct RasterDataset{T,DS} <: AbstractDiskArray{T,3}
+    ds::DS
+    size::Tuple{Int,Int,Int}
+end
+#Forward a few functions
```
yeesian, line 12: Do we have any general principles for the set of functions we want to be defining on a `RasterDataset`?

src/raster/array.jl
```diff
@@ -1,89 +1,106 @@
...
+    b = getband(ds,i)
+    size(b)
+  end
+  s = unique(allsizes)
+  length(s) == 1 || throw(DimensionMismatch("Can not coerce bands to single dataset, different sizes found"))
+  Int64.((s[1]...,nr))
+end
+getband(ds::RasterDataset,i) = getband(ds.ds,i)
+unsafe_readraster(args...;kwargs...)  = RasterDataset(unsafe_read(args...;kwargs...))
+readraster(s::String;kwargs...) = RasterDataset(read(s;kwargs...))
```
yeesian, line 42: Please add a docstring for this function.

yeesian, line 42: Please add docstrings for the following functions:  `RasterDataset(ds::AbstractDataset)`, `readraster(s)`, `getband(ds,i)`, and possibly any other methods meant to be public that I might have missed out.

src/raster/array.jl
```diff
@@ -1,89 +1,106 @@
...
+  GridChunks(ds,(subchunks.chunksize...,1))
+end
+DiskArrays.haschunks(::RasterDataset) = DiskArrays.Chunked()
+DiskArrays.haschunks(::AbstractRasterBand) = DiskArrays.Chunked()
 # AbstractRasterBand indexing
 
 Base.size(band::AbstractRasterBand) = width(band), height(band)
-Base.firstindex(band::AbstractRasterBand, d) = 1
-Base.lastindex(band::AbstractRasterBand, d) = size(band)[d]
+# Don't know if we need these, commenting out for the moment
```
yeesian, line 52: @rafaqz @visr are `Base.firstindex(...)` and `Base.lastindex(...)` still required? If they are, can we try to understand when they might be used?

src/raster/array.jl
```diff
@@ -1,89 +1,106 @@
...
-        Base.setindex_shape_check(value, length.(I)...)
-        value = reshape(value, length.(I))
-    else
-        value = reshape([value], 1, 1)
-    end
-    # Calculate `write!` args and write
-    xoffset, yoffset = first.(I) .- 1
-    xsize, ysize = length.(I)
+DiskArrays.writeblock!(band::AbstractRasterBand, value, x::AbstractUnitRange, y::AbstractUnitRange) = begin
+    # Calculate `read!` args and read
```
yeesian, line 70: I didn't understand this comment.

src/raster/array.jl
```diff
@@ -1,89 +1,106 @@
...
+  DiskArrays.readblock!(dataset::RasterDataset, buffer2, x, y, z)
+  buffer .= buffer2
+end
 
-Base.getindex(dataset::AbstractDataset, x::AllowedXY, y::AllowedXY, bands::AllowedBand=1) = begin
-    I = x, y, bands = Base.to_indices(dataset, (x, y, bands))
-    buffer = Array{pixeltype(getband(dataset, 1)),3}(undef, length.(I)...)
+
+DiskArrays.readblock!(dataset::RasterDataset, buffer::Array, x::AbstractUnitRange, y::AbstractUnitRange, z::AbstractUnitRange) = begin
     # Calculate `read!` args and read
```
yeesian, line 90: ditto

src/raster/array.jl
```diff
@@ -1,89 +1,106 @@
...
 end
 
 
 # AbstractDataset indexing
 
-Base.size(dataset::AbstractDataset) = (size(getband(dataset, 1))..., nraster(dataset))
-Base.firstindex(dataset::AbstractDataset, d) = 1
-Base.lastindex(dataset::AbstractDataset, d) = size(dataset)[d]
+Base.size(dataset::RasterDataset) = dataset.size
+# Base.firstindex(dataset::AbstractDataset, d) = 1
```
yeesian, line 80: ditto

src/raster/array.jl
```diff
@@ -1,89 +1,106 @@
...
+    size(b)
+  end
+  s = unique(allsizes)
+  length(s) == 1 || throw(DimensionMismatch("Can not coerce bands to single dataset, different sizes found"))
+  Int64.((s[1]...,nr))
+end
+getband(ds::RasterDataset,i) = getband(ds.ds,i)
+unsafe_readraster(args...;kwargs...)  = RasterDataset(unsafe_read(args...;kwargs...))
+readraster(s::String;kwargs...) = RasterDataset(read(s;kwargs...))
+function eachchunk(ds::RasterDataset)
```
yeesian, line 43: might be good in the docstring for this function to give a usage example

src/raster/array.jl
meggart: Do you think we need extra docstrings for this? All these functions are already documented for AbstractDataset

src/raster/array.jl
meggart: Basically everything that is relevant for raster-like datasets. Is there a way to get a comprehensive list of everything which is defined for AbstractDataset, so one can decide globally? Of course I would omit things like getfeature etc...

src/raster/array.jl
meggart: AFAIK they are used when users run A[begin:end,:] where `begin` gets replaced by `firstindex(A,1)` and `end` by `lastindex(A,1)`. However, since this is an `AbstractArray` now there are fallback methods implemented https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl#L303

src/raster/array.jl
meggart: Sorry, this was a copy-paste remnant from some other source.

src/raster/array.jl
meggart: This function was imported from DiskArrays and is documented there. It is not necessarily called by a user dealing with an AbstractRasterDatset. Implementing it just enables DiskArrays to provide its efficient reductions and broadcasting etc.

meggart: However, if you think it is beneficial I can add another doc string here

meggart:
Thanks for the comments @yeesian I tried to better document the RasterDataset in the source code which should address many of your comments. Later this week I will try to update https://github.com/yeesian/ArchGDAL.jl/blob/master/docs/src/rasters.md and add a paragraph about dealing with RasterDatasets. 

Maybe we get some more opinions by @rafaqz or @visr or @evetion on which methods should be forwarded for RasterDatasets or other design decisions that should be taken into account.

visr:
Great to see this coming together. I'll go over it later this week.

src/raster/array.jl
yeesian: Ah thanks for explaining -- that wasn't obvious to me because `DS` isn't constrained to be `<: AbstractDataset`.

yeesian: I don't yet know how to get such a list of methods either.. Possibly `methodswith()`?

src/raster/array.jl
```diff
@@ -1,89 +1,133 @@
...
+
+This data structure is returned by the `readraster` function and
+is a wrapper for a GDAL dataset. This wrapper is to signal the
+user that the dataset should be treated as a 3D AbstractArray
+where the first to dimensions correspond to latitude and longitude
+and the third dimension correpsonds to different raster bands.
+
+Constructing a RasterDataset will error if the raster bands do not
+have all the same size and a common element data type.
+"""
```
yeesian, line 18: Might be good to add something to the effect of: "As it is a wrapper around a GDAL Dataset, it supports the usual raster methods for a GDAL Dataset such as `getgeotransform`, `nraster`, `getband`, `getproj`, `width`, and `height`. As it is also a subtype of AbstractDiskArray{T,3}, it supports the following additional methods: `eachchunk`, etc. This satisfies the DiskArray interface, allowing us to be able to index into it like we would an array."

src/raster/array.jl
yeesian: I think I understand it better now: see https://github.com/yeesian/ArchGDAL.jl/pull/105/files#r411777018. Unfortunately, `RasterDataset` is not a subtype of `AbstractDataset`, so those docstrings will not carry over here.

src/raster/array.jl
```diff
@@ -1,89 +1,133 @@
+import DiskArrays: eachchunk, GridChunks
```
yeesian, line 1: I think 
```
import DiskArrays

...

function DiskArrays.eachchunk(...)

function DiskArrays.GridChunks(...)
```

might make it clearer that multi-dispatch is happening below.

src/raster/array.jl
yeesian: Ah gotcha. In that case I don't think the doc string is required. See also https://github.com/yeesian/ArchGDAL.jl/pull/105/files#r411778533

src/raster/array.jl
meggart: Thanks, I added the text to the doc string

src/raster/array.jl
meggart: Yes, that makes sense,

src/raster/array.jl
meggart: Sorry, I think I did not read your comment well enough and only referred to `getband` which already has a docstring. Otherwise, RasterDataset and and readraster are documented now.

src/raster/array.jl
meggart: so these got removed now

src/raster/array.jl
rafaqz: Yeah those  aren't needed now this is wrapped with `AbstractArray`.

visr: Ok if we don't need them, it's probably best to remove them altogether.

src/raster/array.jl
```diff
@@ -1,89 +1,141 @@
...
+"""
+function _common_size(ds::AbstractDataset)
+  nr = nraster(ds)
+  allsizes = map(1:nr) do i
+    b = getband(ds,i)
+    size(b)
+  end
+  s = unique(allsizes)
+  length(s) == 1 || throw(DimensionMismatch("Can not coerce bands to single dataset, different sizes found"))
+  Int64.((s[1]...,nr))
```
visr, line 71: I think this should just be `Int`, i.e. `Int32` on 32 bit systems.

visr:
The `RasterDataset` is also very interesting in relation to https://gdal.org/development/rfc/rfc75_multidimensional_arrays.html.

This is landing in GDAL 3.1.0 which will come out in a few days:

> Implement RFC 75: support for multidirectional arrays in MEM, VRT, netCDF, HDF4, HDF5 and GRIB drivers. Read/write for MEM and netCDF. Read/only for others. Add gdalmdiminfo and gdalmdimtranslate utilities.

Haven't dived much in that RFC yet though.

meggart:
Thanks a lot for the commments @visr. 

> Plus from rasterio.jl functions like rasterio!/read/write/read!/write!.
These are quite a lot though. We could consider using a delegation/forwarding macro like discussed in this thread.
Although since there are ths many, can't we just make RasterDataset an AbstractDataset instead of an AbstractDiskArray{T,3}, and add the needed DiskArrays method we would otherwise get? The DiskArrays readme mentions that instead of subtyping you can also use interpret_indices_disk. Or is there another reason you didn't use that here?

The main reason is that I would like a `RasterDataset` to be a subtype of `AbstractArray`, because ideally it should just feel like an array, and they should be naturally wrappable by packages like `AxisArrays` and  `DImensionalData` which usually expect to wrap and `AbstractArray`. 

Initially I also intentionally did not implement the methods from rasterio.jl, because I see the RasterDataset as an alternative interface to treating a Dataset with much more Julian syntax getindex/setindex syntax. So my thinking was that a user either uses the classical `open_dataset` to obtain a domain-specific object with special methods for `read`, `write`, and `rasterio` or they can use `open_raster` to obtain something array-like that is accessed by square-brackets. I would try to reflect this in the documentation, if we agree to go this way.

visr:
Ok, I see the advantages of subtyping `AbstractArray`, and agree we should go with that for the DiskArray interface. I'm a bit worried about splitting our API in two halves. Perhaps we could migrate in time to a single one that is `AbstractArray`? Probably shouldn't hold up this PR though, but is worth thinking about.

meggart:
Good, so I have added a paragraph to the documentation and added additional unit tests. I think I have to check coveralls to find out why there is still a decrease in coverage. Otherwise I think I have addressed the major comments and this should be ready to go, please let me know if I forgot something.

docs/src/rasters.md
```diff
@@ -118,3 +118,43 @@ end
 !!! note
 
     These methods are often used for reading/writing a block of image data efficiently, as it accesses "natural" blocks from the raster band without resampling, or data type conversion.
+
+# Using the DiskArray interface
+
+## Raster bands as 2D Disk Arrays
+
+As of ArchGDAL version 1.4.2 and higher a `RasterBand` is a subtype of `AbstractDiskArray` from the DiskArrays.jl package. This means that a `RasterBand` is also an `AbstractArray` and can therefore be treated like any Julia array. This means that square bracket indexing works in addition to the `read` methods described above.
```
felixcremer, line 126:
```suggestion
As of ArchGDAL version 1.4.2 and higher a `RasterBand` is a subtype of `AbstractDiskArray` from the [DiskArrays.jl package](https://github.com/meggart/DiskArrays.jl). This means that a `RasterBand` is also an `AbstractArray` and can therefore be treated like any Julia array. This means that square bracket indexing works in addition to the `read` methods described above.  
```

docs/src/rasters.md
```diff
@@ -118,3 +118,43 @@ end
...
+Also, windowed reading of the data can alternatively be done through the DiskArrays interface:
+
+````@example rasters
+using DiskArrays: eachchunk
+for (rows, cols) in take(eachchunk(band), 5)
+    @info "Window" rows, cols
+end
+````
+
+This code is equivalent to the window function mentioned in [Windowed Reads and Writes](@ref) but more portable because the raster band can be exchanged with any other type implementing the DiskArrays interface. Also, for many operations it will not be necessary anymore to implement the window loop, since the `DiskArrays` package provides efficient implementations for reductions and lazy broadcasting, so that for example operations like:
```
felixcremer, line 141:
```suggestion
This code is similar to the window function mentioned in [Windowed Reads and Writes](@ref) but more portable because the raster band can be exchanged with any other type implementing the DiskArrays interface. Also, for many operations it will not be necessary anymore to implement the window loop, since the `DiskArrays` package provides efficient implementations for reductions and lazy broadcasting, so that for example operations like: 
```

visr:
I rebased it on top of master. It looks like there is still a 32 bit issue on DiskArrays, see
https://ci.appveyor.com/project/NgYeeSian/archgdal-jl/builds/34550991/job/52ugv8mjwaam636i#L177. I see DiskArrays tries to take care of this by converting to Int64, not sure what is going wrong: https://github.com/meggart/DiskArrays.jl/blob/76ec89471ea0a328ad5af37be0682a019880596d/src/chunks.jl#L14-L15

meggart:
Thanks @visr for finding this. I think the DiskArrays PR linked above should fix this, but I don't know how to test this PR here against a certain DiskArrays branch. Any ideas?

visr:
48d363765ef6c5dded1d0a7a41dafec500ff3cc0 did the trick, and passes tests on 32 bit, using your DiskArrays branch (https://ci.appveyor.com/project/NgYeeSian/archgdal-jl/builds/34569781/job/3t58q8it787xl6yw#L60).

Do you plan to make a patch release of DiskArrays, and then we remove the Manifest.toml again and bump the compat entry for DiskArrays in the Project.toml?

meggart:
Yes I just created the PR on the registry: https://github.com/JuliaRegistries/General/pull/19258

meggart:
Ok, the new DiskArrays version got tagged. I think @felixcremer wanted to look at the reduced test coverage and add a few tests for the WindowIterator, after that, test coverage should also be fine again.

felixcremer:
I made a pull request against your fork to add tests for WindowIterator. I am wondering, whether we should rename the RasterDataset to RasterArray, because it might be confusing, that the RasterDataset is not a subtype of AbstractDataset. 

Also the show function for RasterBand and RasterDataset fall back to the show methods of a DiskArray. I think it would be nice to fallback to the show method of the underlying dataset for RasterDataset and to indicate that it is a RasterBand for the RasterBand.

felixcremer:
I just looked at the coveralls results and the missing lines in iterators.jl are tested, but not picked up correctly by coveralls.

visr:
> I made a pull request against your fork to add tests for WindowIterator.

Thanks, I didn't see a PR but found the commit and put it on this branch with `git cherry-pick -x 9f731d57`.

> I am wondering, whether we should rename the RasterDataset to RasterArray

I see your point, but can also see the advantages of having Dataset in the name, since it wraps a Dataset. I'd say let's keep it like this.

> Also the display function for RasterBand and RasterDataset fall back to the show methods of a DiskArray.

Indeed. I don't quite understand why it hits those fallbacks, since in `display.jl` there is `Base.show(io::IO, rasterband::AbstractRasterBand)`. And for RasterDataSet it would be nice to show something similar to what `Base.show(io::IO, dataset::AbstractDataset)` shows, minus the feature layers I guess. @meggart could you have a look at the show methods? Otherwise I think this is good to go for me.

I added a few commits. Some of the doc changes are unrelated, sorry for that, but needed to get the docs to build again.

Since this changes the existing types, I don't think we can release this as a patch release, so it would be 0.5.0.

docs/src/rasters.md
```diff
@@ -121,9 +121,9 @@ end
 
 # Using the DiskArray interface
 
-## Raster bands as 2D Disk Arrays
+## Raster bands as 2D DiskArrays
 
-As of ArchGDAL version 1.4.2 and higher a `RasterBand` is a subtype of `AbstractDiskArray` from the [DiskArrays.jl package](https://github.com/meggart/DiskArrays.jl). This means that a `RasterBand` is also an `AbstractArray` and can therefore be treated like any Julia array. This means that square bracket indexing works in addition to the `read` methods described above.  
+As of ArchGDAL version 1.5.0 and higher a `RasterBand` is a subtype of `AbstractDiskArray` from the [DiskArrays.jl package](https://github.com/meggart/DiskArrays.jl). This means that a `RasterBand` is also an `AbstractArray` and can therefore be treated like any Julia array. This means that square bracket indexing works in addition to the `read` methods described above.
```
felixcremer, line 126:
```suggestion
As of ArchGDAL version 0.5.0 and higher a `RasterBand` is a subtype of `AbstractDiskArray` from the [DiskArrays.jl package](https://github.com/meggart/DiskArrays.jl). This means that a `RasterBand` is also an `AbstractArray` and can therefore be treated like any Julia array. This means that square bracket indexing works in addition to the `read` methods described above.  
```

felixcremer:
I think, the problem is, that DiskArrays defines a method of show with a MIME type, 
```julia
[1] show(io::IO, ::MIME{Symbol("text/plain")}, X::DiskArrays.AbstractDiskArray) in DiskArrays at src/DiskArrays.jl:210
```
and this is the function to which the plain show call from the repl is dispatched to. 
I am not sure, whether we would need to define the show function in ArchGDAL with a MIME type as well, or whether this method should be deleted in DiskArrays. 

For the RasterDataset, we can include this line to use the show method of the underlying dataset:
```julia
Base.show(io::IO, ::MIME"text/plain", raster::RasterDataset) = show(io, raster.ds)
```

meggart:
Thanks @felixcremer for pointing this out. I remember I had to do this for DiskArrays because otherwise sometimes the generic AbstractArray show method would still leak through, which would be very bad for a DiskArray. I can not remember the exact reason, why this was necessary, we might revisit this in DiskArrays.

visr:
Thanks. I still want to watch [JuliaCon 2020 | Display, show and print -- how Julia's display system works | Fredrik Ekre](https://youtu.be/S1Fb5oNhhbc).

I think the same patch is needed for `RasterBand` if I remember correctly.

felixcremer:
I added the MIME type patch to the RasterBand show function. 
I also added, show methods without the MIME type, so that `print(band)` works as well. I am not entirely sure, whether this is the correct way to overload these show functions, but this should work for now.

meggart:
I just watched the video and it looks like the preferred way is not to overload `show(io,x::MyType)` but to explicitly pick a MIME type for your output. So I added methods at least for `RasterBand` and `RasterDataset`, but one might think about revisiting display.jl at one point.

visr:
This is a very nice addition to ArchGDAL. Everybody good to merge?

test/test_array.jl
```diff
@@ -1,18 +1,53 @@
...
+            readblock!(ds, data1, 1:3, 1:3, 1:3)
+            @test data1 == ds[1:3,1:3,1:3]
+        end
+    end
+end
+
+
+
+
+
```
yeesian, line 42: nit: seems like there's some redundant white space here. The prevalent style seems to be one empty line between each testset.

visr:
This PR also makes it easier to convert multiple bands into a RGB image:

```julia
using ImageCore, ArchGDAL
# https://github.com/yeesian/ArchGDALDatasets/tree/master/gdalworkshop
ds = ArchGDAL.readraster("world.tif")
img = colorview(RGB, normedview(PermutedDimsArray(ds, (3,2,1))))
```

`size(ds)` is (2048, 1024, 3), so (width, height, bands).
With permutation (3,2,1) the bands are brought to the beginning, because Images.jl expects all colors together in a pixel, and the width and height are swapped to plot the image upright in VS Code. (cc @pritamd47)

Issue #69, opened 2018-12-11
Update from DataStreams.jl to Tables.jl


yeesian:

visr:
Cross linking relevant discussion plus some potentially useful code: https://discourse.julialang.org/t/read-geopackage-layers-to-dataframe/35837

yeesian:
Closed by #118.

PR #113, opened 2020-03-29
[WIP]Update from DataStreams to Tables

Sov-trotter:


visr:
Great contribution, thanks! Let us know when it's ready for review.

I didn't know NamedTupleTools before, but I see you are using it to simplify namedtuple creation, is that right? If it's just that, I would probably prefer not to take on an extra dependency for that reason.

Sov-trotter:
Thanks for looking into it. I was facing a bit of problem with the types, so hastily used the NamedTupleTools. I also added it so that I can simplify the benchmarks that I talked about. I'll have it sorted out before merging. :)

Sov-trotter:
Reference to the issue #69

Sov-trotter:
Rest of the commits will be made in #117

PR #117, opened 2020-04-13
Implement Tables interface for features

Sov-trotter:
Thought to create a new PR, since the old one ended up being too messed up.


Sov-trotter:
Post this commit :

- Base.iterate is no longer dependent on Tables.schema

- Tables.schema now returns two tuples both including name and type of geometry

- Support for multiple geometries is yet to be done

Sov-trotter:
#117

PR #118, opened 2020-04-20
Implement a Tables interface for features

Sov-trotter:
Ran into problems since I updated my fork with the old `tables` branch of origin.


src/tables.jl
```diff
@@ -0,0 +1,108 @@
...
+Tables.rows(gt::GeoTable) = gt
+
+Base.IteratorSize(::Type{<:GeoTable}) = Base.HasLength()
+Base.size(gt::GeoTable) = nfeature(gt.layer)
+Base.length(gt::GeoTable) = (Base.size(gt::GeoTable)) * (nfield(gt.layer) + ngeom(gt.layer))
+Base.IteratorEltype(::Type{<:GeoTable}) = Base.HasEltype()
+
+Base.propertynames(fr::FeatureRow, gt::GeoTable) = (Tables.schema(gt.layer)).names
+#geometry
+# geometry(fr::FeatureRow) = (fr.Row)
```
yeesian, line 95: is this still needed? Can we remove the comment if it is not?

src/tables.jl
```diff
@@ -0,0 +1,108 @@
...
+
```
yeesian, line 96: remove extra line

src/tables.jl
```diff
@@ -0,0 +1,108 @@
...
+
+function Base.show(io::IO, gt::GeoTable)
+    println(io, "GeoTable with $(ArchGDAL.nfeature(gt.layer)) Features")
+end
+# function Base.show(io::IO, fr::FeatureRow)
```
yeesian, line 101: why is this commented out?

src/tables.jl
```diff
@@ -0,0 +1,108 @@
...
+#     println(io, "Feature with properties $(propertynames(fr,gt))")
+# end
+Base.show(io::IO, ::MIME"text/plain", gt::GeoTable) = show(io, gt)
+Base.show(io::IO, ::MIME"text/plain", fr::FeatureRow) = show(io, fr)
+
+
```
yeesian, line 107: remove extra lines

test/test_tables.jl
```diff
@@ -0,0 +1,34 @@
...
+    nfield = AG.nfield(layer)
+    featuredefn = AG.layerdefn(layer)
+    ngeometries = AG.ngeom(featuredefn)
+
+    #testing the main GeoTable function
+    @test sprint(print, AG.GeoTable(layer)) == "GeoTable with 4 Features\n"
+    @test sprint(print, AG.GeoTable(layer)) != "Layer is not a valid ArchGDAL layer\n"      
+    
+    # #testing Tables.schema
+    # @test sprint(print, AG.Tables.schema(layer)) == "Tables.Schema:\n :FID        Float64      \n :pointname  String       \n Symbol(\"\")  GDAL.wkbPoint"
```
yeesian, line 20: why is this commented out?

test/test_tables.jl
```diff
@@ -0,0 +1,34 @@
...
+    # #testing Base.iterate
+    # @test sprint(print, AG.Base.iterate(AG.GeoTable(layer))) == "(ArchGDAL.FeatureRow(NamedTuple{(:pointname, :geometry, :FID),Tuple{String,ArchGDAL.IGeometry,Float64}}[(pointname = \"point-a\", geometry = Geometry: POINT (100 0), FID = 2.0)], 0), 1)"
```
yeesian, line 22: why is this commented out?

test/test_tables.jl
```diff
@@ -0,0 +1,34 @@
...
+    @test Tables.rowaccess(AG.GeoTable(layer)) == true
+    @test Tables.rows(AG.GeoTable(layer)) == AG.GeoTable(layer)
+
+    @test Base.size(AG.GeoTable(layer)) == 4
+    @test Base.length(AG.GeoTable(layer)) == 12
+
+end
+
+
+
```
yeesian, line 34: remove extra lines

test/test_tables.jl
```diff
@@ -0,0 +1,34 @@
...
+    # #testing Base.iterate
+    # @test sprint(print, AG.Base.iterate(AG.GeoTable(layer))) == "(ArchGDAL.FeatureRow(NamedTuple{(:pointname, :geometry, :FID),Tuple{String,ArchGDAL.IGeometry,Float64}}[(pointname = \"point-a\", geometry = Geometry: POINT (100 0), FID = 2.0)], 0), 1)"
+
+    @test Tables.istable(AG.GeoTable(layer)) == true
+    @test Tables.rowaccess(AG.GeoTable(layer)) == true
+    @test Tables.rows(AG.GeoTable(layer)) == AG.GeoTable(layer)
+
+    @test Base.size(AG.GeoTable(layer)) == 4
+    @test Base.length(AG.GeoTable(layer)) == 12
+
```
yeesian, line 30: remove extra line

test/test_tables.jl
```diff
@@ -0,0 +1,34 @@
+using Test
+import ArchGDAL; const AG = ArchGDAL
+using Tables
+
+@testset "Tables Support" begin
+    # dataset = AG.read("point.geojson")
```
yeesian, line 6: remove this line?

test/test_tables.jl
```diff
@@ -0,0 +1,34 @@
...
+    dataset = AG.read(joinpath(@__DIR__, "data/point.geojson"))
+    layer = AG.getlayer(dataset, 0)
+
+    nfeat = AG.nfeature(layer)
+    nfield = AG.nfield(layer)
+    featuredefn = AG.layerdefn(layer)
+    ngeometries = AG.ngeom(featuredefn)
+
+    #testing the main GeoTable function
```
yeesian, line 15: I think the standard practice is to do
```
@testset "GeoTables" begin
    ...
end
```

test/test_tables.jl
```diff
@@ -0,0 +1,34 @@
...
+    @test sprint(print, AG.GeoTable(layer)) == "GeoTable with 4 Features\n"
+    @test sprint(print, AG.GeoTable(layer)) != "Layer is not a valid ArchGDAL layer\n"      
+    
+    # #testing Tables.schema
```
yeesian, line 19: I think the standard practice is to do
```
@testset "Tables Schema" begin
    ...
end
```

Sov-trotter:
So the tests fail here https://github.com/Sov-trotter/ArchGDAL.jl/blob/dev-tables/test/test_tables.jl#L17. For julia 1.5 while it works for 1.3. Some weird spacing issue. 
What do you thing we can do here @yeesian ? Test it only for 1.5?

visr:
> Test it only for 1.5?

We should try to understand and fix the error if possible. In CI all versions fail though, right?

Sov-trotter:
So yeah here's the bit that gives error:
```julia
"Tables.Schema:\n :FID        Float64      \n :pointname  String       \n Symbol(\"\")  GDAL.wkbPoint" == 
"Tables.Schema:\n :FID        Float64      \n :pointname  String       \n Symbol(\"\")  wkbPoint::OGRwkbGeometryType = 0x00000001"
```
In 1.3 the `geomtype` returned by the schema is `GDAL.wkbPoint` while 1.5 gives out the type as `wkbPoint::OGRwkbGeometryType`

Also in 1.3
```julia
"Tables.Schema:\n :FID        Float64\n :pointname  String\n Symbol(\"\")  wkbPoint::OGRwkbGeometryType = 0x00000001" == "Tables.Schema:\n :FID        Float64      \n :pointname  String       \n Symbol(\"\")  wkbPoint::OGRwkbGeometryType = 0x00000001"
```
there's this extra indent before newline while the expression in RHS corresponds to 1.3 indent

yeesian:
I have to say I prefer the display in `1.3`, although it does feel like the tests here are too brittle with their expectations of what's being displayed.

Sov-trotter:
Yeah. I'll fix it rightaway. :)

Sov-trotter:
Hey @yeesian can you take a look?

src/tables.jl
```diff
@@ -0,0 +1,95 @@
...
+struct GeoTable{T<:Union{IFeatureLayer, FeatureLayer}} 
+    layer::T
+end
+
+function geotable(layer)
+    featuredefn = layerdefn(layer)
+    ngeometries = ngeom(featuredefn)
+    if(ngeometries == 0 )
+        error("Layer is not a valid ArchGDAL layer")
```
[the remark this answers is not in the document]
Sov-trotter: I am not really sure. If we go by https://github.com/visr/GeoJSONTables.jl/commit/c359c029666de69da2718645dddf6a0b6e59e6e3 then it's not. But are we aiming for the same here @visr ?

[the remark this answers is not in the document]
visr: GeoJSON allows for null geometries. I think GDAL does as well? That would mean we should support that.

Sov-trotter:
@visr  is this right?

src/tables.jl
```diff
@@ -0,0 +1,105 @@
...
+    end
+    
+    keys_tup = ()
+    for _key in keys(d)
+        keys_tup = (keys_tup..., Symbol(_key))
+    end
+    vals_tup = Tuple(values(d))
+    
+    #Using the tables interface
+    Row = Tables.rowtable(NamedTuple{keys_tup}(vals_tup))
```
Sov-trotter, line 58: So the main reason of doing this is https://github.com/Sov-trotter/ArchGDAL.jl/blob/dev-tables/src/tables.jl#L40. If we simply return a splatted NamedTuple, then every item is a vector.
Screenshot from 2020-08-15 15-58-15

visr: Don't see why it should be a vector though? We should probably get rid of all the dicts and vectors in this method, and just stick to NamedTuples of combined geometries and attributes.

src/tables.jl
```diff
@@ -0,0 +1,100 @@
...
+    name = []
+    v = []
+    for field_no in 0:nfield-1
+        field = AG.getfielddefn(featuredefn, field_no)
+        push!(name, AG.getname(field))
+    end
+    push!(name, "geometry")
+    
+    st >= nfeat && return nothing
+    AG.getfeature(layer, st) do feature
```
Sov-trotter, line 41: So the main reason of using `getfeature` instead of `nextfeature` here is that we can link the row/feature number directly with the value of `st`. This is useful when we are calling the `iterate` method iteratively eg: while constructing a `DataFrame` .
Now the issue with `getfeature` arises when we have two different file formats, eg: GeoJSON starts from 0, while geopackage layers have features starting from 1, so if st is initially 0, geojson works while geopackage throws a Null Feature hence an error. 

Now this is handled well by `nextfeature`, where we don't have to worry about the indices. But the thing is again if we call the iterate method repeatedly(eg DataFrame.),for the first time it works well, but now at the end the layer reading is +1 the size of the layer. So if we call the DataFrame() method again, it starts with a NullFeature and hence an error. 

Apparently `AG.resetreading!(layer)` is made to reset the layer's feature number reading. But since Base.iterate is called repeatedly, using it will keep returning the first feature only.

Sov-trotter: Here are a few possible workarounds:
* We use `getfeature` and apply an if condition in the beginning to check for indexing and direct it appropriately.
* or perhaps use the `nextfeature` method and specify in the docs those consuming the API must use `AG.resetreading!(layer)` at after every iterative call?

visr: What, I am really surprised if `getfeature` starts from 0 in format A but 1 in format B??

But I think `nextfeature` would be better anyway. Isn't it enough to do `if iszero(st) resetreading`?

Sov-trotter:
Superseded by #137

src/tables.jl
```diff
@@ -0,0 +1,79 @@
...
+Tables.rows(t::Table) = t
+
+Base.IteratorSize(::Type{<:Table}) = Base.HasLength()
+Base.size(t::Table) = nfeature(t.layer)
+Base.length(t::Table) = Base.size(t)
+Base.IteratorEltype(::Type{<:Table}) = Base.HasEltype()
+Base.propertynames(t::Table) = Tables.schema(t.layer).names
+
+function Base.show(io::IO, t::Table)
+    println(io, "Table with $(ArchGDAL.nfeature(t.layer)) Features")
```
yeesian, line 74: nit: "features" instead of "Features"?

test/test_tables.jl
```diff
@@ -0,0 +1,42 @@
...
+    @test Tables.rows(gt1) == AG.Table(layer1)
+    @test Base.size(gt) == 4
+    @test Base.size(gt1) == 2
+    @test Base.length(gt) == 4
+    @test Base.length(gt1) == 2
+    @test propertynames(iterate(gt, 1)[1]) == (:FID, :pointname, Symbol(""))
+    @test propertynames(iterate(gt1, 1)[1]) == (:id, :zoom, :location, :point, :linestring)
+    @test iterate(gt, 5) === nothing
+    @test iterate(gt1, 3) === nothing
+end
```
yeesian, line 42: should we add a newline at the end of the file?

test/test_tables.jl
```diff
@@ -0,0 +1,42 @@
...
+    ngeometries = AG.ngeom(featuredefn)
+    ngeometries1 = AG.ngeom(featuredefn1)
+
+    @test sprint(print, gt) == "Table with 4 Features\n"
+    @test sprint(print, gt1) == "Table with 2 Features\n"
+    @test getproperty(Tables.schema(layer), :types) == (Float64, String)
+    @test getproperty(Tables.schema(layer1), :types) == (String, String, String)
+    @test getproperty(Tables.schema(layer), :names) == propertynames(gt)
+    @test getproperty(Tables.schema(layer1), :names) == propertynames(gt1)
+
```
yeesian, line 29: remove extra new line

test/remotefiles.jl
```diff
@@ -10,7 +10,16 @@ const testdatadir = @__DIR__
 REPO_URL = "https://github.com/yeesian/ArchGDALDatasets/blob/master/"
 
 # remote files with SHA-2 256 hash
+"""
+While adding more files, follow the below steps to generate the SHA
```
yeesian, line 14: nit: Replace "While adding" with "To add"

src/tables.jl
```diff
@@ -0,0 +1,79 @@
...
+end
+Base.show(io::IO, ::MIME"text/plain", t::Table) = show(io, t)
+
+
+
```
yeesian, line 79: just one newline to mark the end of the file should be sufficient

src/tables.jl
```diff
@@ -0,0 +1,79 @@
...
+                val = AG.getfield(feature, k)
+            end
+            push!(v, val)
+        end
+    end
+    
+    Row = NamedTuple{Tuple(Symbol.(name))}(v)
+    return Row, st + 1
+end
+
```
yeesian, line 61: remove extra newline

src/tables.jl
```diff
@@ -0,0 +1,79 @@
...
+    featuredefn = AG.layerdefn(layer)
+    geomdefns = (ArchGDAL.getgeomdefn(featuredefn, i) for i in 0:ngeom-1)
+    
+    name = []
+    v = []
+    for field_no in 0:nfield-1
+        field = AG.getfielddefn(featuredefn, field_no)
+        push!(name, AG.getname(field))
+    end
+    geom_names = [ArchGDAL.getname(geomdefn) for geomdefn in geomdefns]
```
yeesian, line 43: out of curiosity, can this be a generator rather than a vector?

src/tables.jl
```diff
@@ -0,0 +1,79 @@
...
+        AG.resetreading!(layer) 
+    end
+
+    nfeat = AG.nfeature(layer)
+    nfield = AG.nfield(layer)
+    ngeom = AG.ngeom(layer)
+    featuredefn = AG.layerdefn(layer)
+    geomdefns = (ArchGDAL.getgeomdefn(featuredefn, i) for i in 0:ngeom-1)
+    
+    name = []
```
yeesian, line 37: `name = String[]`?

src/tables.jl
```diff
@@ -0,0 +1,79 @@
...
+    v = []
```
yeesian, line 38: indicate the type of vector here?

src/tables.jl
```diff
@@ -0,0 +1,79 @@
+const AG = ArchGDAL
+
+struct Table{T}
```
yeesian, line 3: optional: Can we give this an informative docstring?

src/tables.jl
Sov-trotter: Yeah I wanted to to the same but https://github.com/yeesian/ArchGDAL.jl/pull/118/files/fbdef4929e9294b51ccb91c8c3a90e546a4b149c#diff-17226d5590764ff7e308bfa5045dacfaR49 part which checks for geometry names doesn't accept generators for some reason.

docs/src/tables.md
```diff
@@ -0,0 +1,58 @@
...
+The [`ArchGDAL.Table`](@ref) method accepts an `ArchGDAL.FeatureLayer`.
+```@example tables
+table = ArchGDAL.Table(layer)
+```
+
+Individual rows can be retrieved using the `iterate(t::ArchGDAL.Table, n::Int)` method.
+
+```@example tables
+row = Base.iterate(table, 1)
+```
```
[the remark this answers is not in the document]
Sov-trotter: Thank you so much for the review. Things are way better now. This does look cool. I am pushing the changes, you can take a look. :)

Sov-trotter:
So I have included geometry names and types in `Tables.Schema`. Is that right?
Also we were using it to get the types in the `iterate` method but now that `Tables.schema()` doesn't give the geometry types as `IGeometry`, I created `types_fields` again and passed `IGeometry` separately. 

I wanted to discuss if this tradeoff is useful?

Sov-trotter:
So the in the commit after the above comment includes the following which i think is right :
 * Creates a common method `get_row` that basically eliminates writing the samme code twice for `getindex` and `iterate` .
* Since both the methods for Table tpyes return the geometry type `IGeometry` it's better that even Tables.schema return `IGeometry` which is better than having a `wkbUnknown`

docs/src/tables.md
```diff
@@ -28,10 +28,10 @@ The [`ArchGDAL.Table`](@ref) method accepts an `ArchGDAL.FeatureLayer`.
 table = ArchGDAL.Table(layer)
 ```
 
-Individual rows can be retrieved using the `iterate(t::ArchGDAL.Table, n::Int)` method.
+Individual rows can be retrieved using the `Base.getindex(t::ArchGDAL.Table, idx::Int)` method.
 
 ```@example tables
-row = Base.iterate(table, 1)
+row = Base.getindex(table, 1)
```
visr, line 34: I'd suggest `table[1]`, which means the same but is nicer syntax.

src/tables.jl
```diff
@@ -61,6 +61,29 @@ Base.size(t::Table) = nfeature(getlayer(t))
 Base.length(t::Table) = size(t)
 Base.IteratorEltype(::Type{<:Table}) = Base.HasEltype()
 Base.propertynames(t::Table) = Tables.schema(getlayer(t)).names
+Base.getproperty(t::Table, s::Symbol) = [getproperty(iterate(t, i)[1], s) for i in 0:size(t)-1]
```
visr, line 45: can simplify to `[row.s for row in t]`

src/tables.jl
```diff
@@ -66,13 +47,16 @@ Base.getproperty(t::Table, s::Symbol) = [getproperty(iterate(t, i)[1], s) for i
...
     featuredefn = layerdefn(layer)
-    field_names = Tuple(Symbol(getname(getfielddefn(featuredefn, i-1))) for i in 1:nfield(layer))
     fielddefns = (getfielddefn(featuredefn, i) for i in 0:nfield(layer)-1)
-    types_fields = (_FIELDTYPE[gettype(fielddefn)] for fielddefn in fielddefns)
-    geom_names = Tuple(Symbol(getname(getgeomdefn(featuredefn, i-1))) for i in 1:ngeom(layer))
-    v = Union{types_fields..., IGeometry}[]
+    field_names = Tuple(Symbol(getname(fielddefn)) for fielddefn in fielddefns)
+    geom_names = (Symbol(getname(getgeomdefn(featuredefn, i-1))) for i in 1:ngeom(layer))
 
+    v = Union{Tables.schema(layer).types...}[]
```
visr, line 59: It's work looking if it's better to just make it `Any`, i.e. `v = []`. Or alternatively, is it possible to do this without creating a vector, for instance making it a generator and converting that to Tuple.

visr:
> So I have included geometry names and types in Tables.Schema. Is that right?

Yeah good catch, missed that.

> Since both the methods defined for the Table type return the geometry type IGeometry it's better that even Tables.schema return IGeometry which is better than having a wkbUnknown

Isn't the Julia type always `IGeometry` here, with `wkbUnknown` merely what GDAL shows, i.e. the geometry type internal to GDAL? GDAL.jl defines `wkbUnknown` to be

```julia
"List of well known binary geometry types"
@cenum OGRwkbGeometryType::UInt32 begin
    wkbUnknown = 0
    wkbPoint = 1
    wkbLineString = 2
    wkbPolygon = 3
...
```

Sov-trotter:
>Isn't the Julia type always IGeometry here 

I am not sure. Since I had tried putting the geometries into a GDAL typed array earlier an it was throwing errors. But now the issues are pretty much resolved and the code is cleaner than ever. Also kudos to the one who wrote `nextfeature()` which actually helped reduce the allocations and complexity.

src/tables.jl
```diff
@@ -0,0 +1,69 @@
...
+    featuredefn = layerdefn(layer)
+    fielddefns = (getfielddefn(featuredefn, i) for i in 0:nfield(layer)-1)
+    field_names = Tuple(Symbol(getname(fielddefn)) for fielddefn in fielddefns)
+    geom_names = (Symbol(getname(getgeomdefn(featuredefn, i-1))) for i in 1:ngeom(layer))
+
+    nextfeature(layer) do feature
+        prop = (getfield(feature, name) for name in field_names)
+        geom = (getgeom(feature, idx-1) for idx in 1:length(geom_names))
+        row = NamedTuple{(field_names..., geom_names...)}((prop..., geom...))
+    end
```
yeesian, line 63: Can we make it clearer that we're returning a value here? A possibility is to rewrite this as
```julia
    return nextfeature(layer) do feature
        prop = (getfield(feature, name) for name in field_names)
        geom = (getgeom(feature, idx-1) for idx in 1:length(geom_names))
        NamedTuple{(field_names..., geom_names...)}((prop..., geom...))
    end
```
(or something to that effect)

test/test_tables.jl
```diff
@@ -0,0 +1,47 @@
...
+    dataset = AG.read(joinpath(@__DIR__, "data/point.geojson"))
+    dataset1 = AG.read(joinpath(@__DIR__, "data/multi_geom.csv"), options = ["GEOM_POSSIBLE_NAMES=point,linestring", "KEEP_GEOM_COLUMNS=NO"])
+    @test dataset isa ArchGDAL.IDataset
+    @test dataset1 isa ArchGDAL.IDataset
+    layer = AG.getlayer(dataset, 0)
+    layer1 = AG.getlayer(dataset1, 0)
+    gt = AG.Table(layer)
+    gt1 = AG.Table(layer1)
+        
+
```
yeesian, line 15: redundant newline

src/tables.jl
```diff
@@ -0,0 +1,69 @@
...
+Base.length(t::Table) = size(t)
+Base.IteratorEltype(::Type{<:Table}) = Base.HasEltype()
+Base.propertynames(t::Table) = Tables.schema(getlayer(t)).names
+Base.getproperty(t::Table, s::Symbol) = [getproperty(row, s) for row in t]
+    
+function get_row(layer::IFeatureLayer)
+    featuredefn = layerdefn(layer)
+    fielddefns = (getfielddefn(featuredefn, i) for i in 0:nfield(layer)-1)
+    field_names = Tuple(Symbol(getname(fielddefn)) for fielddefn in fielddefns)
+    geom_names = (Symbol(getname(getgeomdefn(featuredefn, i-1))) for i in 1:ngeom(layer))
```
yeesian, line 57: Out of curiosity, why is `field_names` a Tuple while `geom_names` is a generator?

src/tables.jl
```diff
@@ -0,0 +1,69 @@
...
+function Tables.schema(layer::AbstractFeatureLayer)
+    featuredefn = layerdefn(layer)
+    fielddefns = (getfielddefn(featuredefn, i) for i in 0:nfield(layer)-1)
+    names_fields = (getname(fielddefn) for fielddefn in fielddefns)
+    types_fields = (_FIELDTYPE[gettype(fielddefn)] for fielddefn in fielddefns)
+    
+    ngeom = ArchGDAL.ngeom(featuredefn)
+    geomdefns = (ArchGDAL.getgeomdefn(featuredefn, i) for i in 0:ngeom-1)
+    geom_names = (ArchGDAL.getname(geomdefn) for geomdefn in geomdefns)
+    geom_types = (IGeometry for i in 1:ngeom)
```
yeesian, line 22: Consider defining a method like

```julia
function schema_names(featuredefn)
    fielddefns = (getfielddefn(featuredefn, i) for i in 0:nfield(layer)-1)
    field_names = Tuple(Symbol(getname(fielddefn)) for fielddefn in fielddefns)
    geom_names = Tuple(Symbol(getname(getgeomdefn(featuredefn, i-1))) for i in 1:ngeom(layer))
    return (field_names, geom_names)
end
```

and using it across both here and in the `get_rows()` function.

yeesian: Alternatively, use `Tables.schema(layer)` in `function get_row(layer::IFeatureLayer)` later.

src/tables.jl
```diff
@@ -0,0 +1,69 @@
...
+end
+
+Base.IteratorSize(::Type{<:Table}) = Base.HasLength()
+Base.size(t::Table) = nfeature(getlayer(t))
+Base.length(t::Table) = size(t)
+Base.IteratorEltype(::Type{<:Table}) = Base.HasEltype()
+Base.propertynames(t::Table) = Tables.schema(getlayer(t)).names
+Base.getproperty(t::Table, s::Symbol) = [getproperty(row, s) for row in t]
+    
+function get_row(layer::IFeatureLayer)
```
yeesian, line 53: Can we name this function `nextnamedtuple(layer)` (or something to that effect to indicate the similarities with `nextfeature`) and give it a good docstring? I can see it having broader utility (beyond support for tables) for people who prefer to work with `NamedTuples` rather than `Features`.

Sov-trotter: That sounds cool!

yeesian:
https://discourse.julialang.org/t/archgdal-getfield-error/45770 reminded me that I did not check the behavior of this PR with respect to missing/NULL values -- do we have test cases for those?

Sov-trotter:
Oh sorry. I wanted to add it but it slipped my mind. 
So if I have interpreted the issue correctly, there seems to be a minor conflict with the data? 
Since I modified this example https://github.com/yeesian/ArchGDALDatasets/blob/master/data/multi_geom.csv by putting missing into location and left the geometry field empty(not sure how GDAL NULL Geometry is represented in csv) 
Screenshot from 2020-08-30 12-51-57

Putting missing in the Geometry field instead of keeping it empty gave a
 ```
ERROR: GDALError (CE_Failure, code 1):
        Invalid EWKB content : 3 bytes
```

**Comment by Sov-trotter** [comment]
If this is right then should I push the new dataset?

**Comment by yeesian** [comment]
Yeah sure! I think that will help -- maybe we can have it as a new test case, rather than modifying an existing one?

**Comment by Sov-trotter** [comment]
One thing to note here is that GDAL would read all fields except geometry ones as `String`. So the attributes are actually `String`. https://github.com/Sov-trotter/ArchGDALDatasets/blob/master/data/missing_testcase.csv.
And in that case the `missing` fields are actually `"missing"`.

File: test/test_featurelayer.jl [hunk 34]
```diff
@@ -148,7 +148,7 @@ import ArchGDAL; const AG = ArchGDAL
             end
         end
         @test AG.findfieldindex(layer,"FID", true) == 0
-        @test AG.findfieldindex(layer,"FID", false) == 0
+        @test AG.findfieldindex(layer,:FID, false) == 0
         @test AG.findfieldindex(layer,"pointname", true) == 1
         @test AG.findfieldindex(layer,"pointname", false) == 1
         @test AG.findfieldindex(layer,"geom", true) == -1
```
yeesian, line 154: nit: add a space after `layer,` for each line

src/tables.jl
```diff
@@ -0,0 +1,73 @@
...
+
+Base.IteratorSize(::Type{<:Table}) = Base.HasLength()
+Base.size(t::Table) = nfeature(getlayer(t))
+Base.length(t::Table) = size(t)
+Base.IteratorEltype(::Type{<:Table}) = Base.HasEltype()
+Base.propertynames(t::Table) = Tables.schema(getlayer(t)).names
+Base.getproperty(t::Table, s::Symbol) = [getproperty(row, s) for row in t]
+    
+"""
+nextnamedtuple(layer::IFeatureLayer)
```
yeesian, line 49: nit: remove this line

src/tables.jl
```diff
@@ -0,0 +1,73 @@
...
+Returns the feature row of a layer as a `NamedTuple`
+Calling it iteratively will work similar to `nextfeature` i.e. give the consecutive feature as `NamedTuple`
```
yeesian, line 51: nit: Add a newline between the top-level docstring, and the detailed description.

I.e.
```
Returns the feature row of a layer as a `NamedTuple`.

Calling it iteratively will work similar to `nextfeature` i.e. give the consecutive feature as `NamedTuple`.
```

evetion:
Shall we test this on an (existing) shapefile or geopackage as well?

Sov-trotter:
Yeah. That does sound cool. Though there isn't any existing geopackage dataset in the repo. 
Since adding tests for shp and gpkg will be easy, I wish to leave it out for this year's Hacktoberfest as `good first issues`? I have a few people willing to make contributions?

yeesian:
Hey everyone, we're close to getting this PR merged -- let's see it through to completion?

@visr @evetion are there any specific requests that must go into this PR, that cannot be opened as additional feature requests?

visr:
I see the `Base.getindex(t::Table, idx::Int)` is 0-based. Shouldn't that be 1-based?
Otherwise I think it's good to merge this.

The iteration state issue I described in https://github.com/yeesian/ArchGDAL.jl/pull/118#discussion_r475235520 is still present I believe. But we could also turn that into an issue, and tackle it later.

Sov-trotter:
@visr can you take a look?

visr:
Thanks, looks good.

Sov-trotter:
Btw the issue that you described above, I think it is now handled well in the interface unless one goes around inappropriate indices?

evetion:
I'm fine with merging if @visr items are merged. Things like shapefile, geopackage testing are nice Hacktoberfest issues.

Happy to have this functionality in. ☺️

src/tables.jl
```diff
@@ -0,0 +1,73 @@
...
+function Base.iterate(t::Table, st = 0)
+    layer = getlayer(t)
+    st >= nfeature(layer) && return nothing
+    if iszero(st)
+        resetreading!(layer)
+    end
+    return nextnamedtuple(layer), st + 1
+end
+
+function Base.getindex(t::Table, idx::Int)
```
yeesian, line 35: Should it be `idx::Integer`?

See the [AppVeyor failure](https://ci.appveyor.com/project/NgYeeSian/archgdal-jl/builds/35181016/job/57p2lmp40olma6c9):
```
Misc. methods: Error During Test at C:\projects\archgdal-jl\test\test_tables.jl:58
  Test threw exception
  Expression: typeof([getindex(gt1, i) for i = 1:size(gt1)]) == typeof([(iterate(gt1, i))[1] for i = 0:size(gt1) - 1])
  MethodError: no method matching getindex(::ArchGDAL.Table{ArchGDAL.IFeatureLayer}, ::Int64)
  Closest candidates are:
    getindex(::ArchGDAL.Table, !Matched::Int32) at C:\projects\archgdal-jl\src\tables.jl:36
  Stacktrace:
   [1] (::var"#11#19"{ArchGDAL.Table{ArchGDAL.IFeatureLayer}})(::Int64) at .\none:0
   [2] iterate at .\generator.jl:47 [inlined]
   [3] collect(::Base.Generator{UnitRange{Int64},var"#11#19"{ArchGDAL.Table{ArchGDAL.IFeatureLayer}}}) at .\array.jl:622
   [4] top-level scope at C:\projects\archgdal-jl\test\test_tables.jl:58
   [5] top-level scope at D:\buildbot\worker\package_win32\build\usr\share\julia\stdlib\v1.3\Test\src\Test.jl:1107
   [6] top-level scope at C:\projects\archgdal-jl\test\test_tables.jl:39
   [7] top-level scope at D:\buildbot\worker\package_win32\build\usr\share\julia\stdlib\v1.3\Test\src\Test.jl:1107
   [8] top-level scope at C:\projects\archgdal-jl\test\test_tables.jl:6
```

visr:
> Btw the issue that you described above, I think it is now handled well in the interface unless one goes around inappropriate indices?

No it is still there, just try running `iterate(gt, 2)` several times on such a table. The state (2) is not connected to the GDAL tracked current row. Connecting this in some way would perhaps be a nice improvement, and less prone to errors, but doesn't need to be a part of this PR.

src/tables.jl
Sov-trotter: Oh sorry I ignored it thinking that the failures were x86 expecting it to be an internal issue.

yeesian:
@visr I agree that `Base.iterate()` should not be used as a public method. I think a robust way of handling it might be use `setnextbyindex!(layer, i)` inside `Base.iterate()`, but it'll make it unusable for drivers that have O(n) performance in setting the index.

FWIW, the same issue is there for the [current code](https://github.com/yeesian/ArchGDAL.jl/blob/260b8525cc30948f5806fb79dd720ca00fe82805/src/base/iterators.jl#L1), and we're lucky we haven't run into any issues so far. We can open a feature request for us to wrap the layer/table (similar to how we do it for rasterbands) and have ways of qualifying the state (rather than using arbitrary integers).

yeesian:
This is huge, congrats @Sov-trotter! :)

PR #120, opened 2020-05-11
Add a kwarg to set OSRAxisMappingStrategy

### LINKED ISSUE #119: importEPSG and importEPSGA both swap axis order to lat/lon
https://github.com/yeesian/ArchGDAL.jl/blob/f57c2dcd6df9bd0e6d358db7c4060e8e17ebba73/src/spatialref.jl#L145

I think this no longer happens as of GDAL 3.0, and both methods are identical:

https://gdal.org/doxygen/classOGRSpatialReference.html#aaa6965a1df98cdc673dfb20697eab613

```julia
julia> p = ArchGDAL.createpoint(0.0, 180.0)
Geometry: POINT (0 180)

julia> ArchGDAL.importEPSG(4326) do sourcecrs_ref
           ArchGDAL.importPROJ4("+proj=cea") do targetcrs_ref
               ArchGDAL.createcoordtrans(sourcecrs_ref, targetcrs_ref) do transform
                   ArchGDAL.transform!(p, transform)
               end
           end
       end
Geometry: POINT (20037508.3427892 0.0)

julia> p = ArchGDAL.createpoint(0.0, 180.0)
Geometry: POINT (0 180)

julia> ArchGDAL.importEPSGA(4326) do sourcecrs_ref
           ArchGDAL.importPROJ4("+proj=cea") do targetcrs_ref
               ArchGDAL.createcoordtrans(sourcecrs_ref, targetcrs_ref) do transform
                   ArchGDAL.transform!(p, transform)
               end
           end
       end
Geometry: POINT (20037508.3427892 0.0)
```

For my use cases it breaks the generality or `reproject` when the axis order switches
for EPSG, which worked when I first wrote it! I'm not sure how to handle that now. Probably
all upstream uses should convert EPSG to PROJ before using `reproject`.

rafaqz:
This is a first pass at setting `OSRAxisMappingStrategy` with a keyword argument.

I'm not sure how to implement it for `fromWKT` etc methods, where the `@gdal` macro is used, so some input on that would be helpful.

But otherwise it works for reprojecting points.

The default is `:trad` now for testing, but we can change that to `:compliant` depending on what people want the default behaviour to be for ArchGDAL

closes #119

### DIFF
--- a/src/ogr/geometry.jl
+++ b/src/ogr/geometry.jl
@@ -8,7 +8,7 @@ binary (WKB) representation.
 ### Parameters
 * `data`: pointer to the input BLOB data.
 """
-function fromWKB(data; kwargs...)
+function fromWKB(data)
     geom = Ref{GDALGeometry}()
     result = @gdal(OGR_G_CreateFromWkb::GDAL.OGRErr,
         data::Ptr{Cuchar},
@@ -20,7 +20,7 @@ function fromWKB(data; kwargs...)
     return IGeometry(geom[])
 end
 
-function unsafe_fromWKB(data; kwargs...)
+function unsafe_fromWKB(data)
     geom = Ref{GDALGeometry}()
     result = @gdal(OGR_G_CreateFromWkb::GDAL.OGRErr,
         data::Ptr{Cuchar},
@@ -43,7 +43,7 @@ Create a geometry object of the appropriate type from its well known text
     geometry to be created. The pointer is updated to point just beyond that
     last character consumed.
 """
-function fromWKT(data::Vector{String}; kwargs...)
+function fromWKT(data::Vector{String})
     geom = Ref{GDALGeometry}()
     result = @gdal(OGR_G_CreateFromWkt::GDAL.OGRErr,
         data::StringList,
@@ -54,7 +54,7 @@ function fromWKT(data::Vector{String}; kwargs...)
     return IGeometry(geom[])
 end
 
-function unsafe_fromWKT(data::Vector{String}; kwargs...)
+function unsafe_fromWKT(data::Vector{String})
     geom = Ref{GDALGeometry}()
     result = @gdal(OGR_G_CreateFromWkt::GDAL.OGRErr,
         data::StringList,
@@ -65,11 +65,9 @@ function unsafe_fromWKT(data::Vector{String}; kwargs...)
     return Geometry(geom[])
 end
 
-fromWKT(data::String, args...; kwargs...) = 
-    fromWKT([data], args...; kwargs...)
+fromWKT(data::String, args...) = fromWKT([data], args...)
 
-unsafe_fromWKT(data::String, args...; kwargs...) = 
-    unsafe_fromWKT([data], args...; kwargs...)
+unsafe_fromWKT(data::String, args...) = unsafe_fromWKT([data], args...)
 
 """
 Destroy geometry object.
@@ -260,7 +258,7 @@ wkbsize(geom::AbstractGeometry) = GDAL.ogr_g_wkbsize(geom.ptr)
 
 Convert a geometry into well known text format.
 """
-function toWKT(geom::AbstractGeometry; kwargs...)
+function toWKT(geom::AbstractGeometry)
     wkt_ptr = Ref(Cstring(C_NULL))
     result = GDAL.ogr_g_exporttowkt(geom.ptr, wkt_ptr)
     @ogrerr result "OGRErr $result: failed to export geometry to WKT"
@@ -274,7 +272,7 @@ end
 
 Convert a geometry into SFSQL 1.2 / ISO SQL/MM Part 3 well known text format.
 """
-function toISOWKT(geom::AbstractGeometry; kwargs...)
+function toISOWKT(geom::AbstractGeometry)
     isowkt_ptr = Ref(Cstring(C_NULL))
     result = GDAL.ogr_g_exporttoisowkt(geom.ptr, isowkt_ptr)
     @ogrerr result "OGRErr $result: failed to export geometry to ISOWKT"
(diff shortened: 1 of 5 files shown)


src/ogr/geometry.jl
```diff
@@ -43,7 +43,7 @@ Create a geometry object of the appropriate type from its well known text
     geometry to be created. The pointer is updated to point just beyond that
     last character consumed.
 """
-function fromWKT(data::Vector{String})
+function fromWKT(data::Vector{String}; kwargs...)
     geom = Ref{GDALGeometry}()
     result = @gdal(OGR_G_CreateFromWkt::GDAL.OGRErr,
         data::StringList,
```
rafaqz, line 49: My main question is how best to set `OSRAxisMappingStrategy` for this spatial ref

src/ogr/geometry.jl
```diff
@@ -20,7 +20,7 @@ function fromWKB(data)
     return IGeometry(geom[])
```
rafaqz, line 20: My main question is how best to set `OSRAxisMappingStrategy` for these spatial refs

yeesian: You'd replace the `C_NULL` in `C_NULL::GDALSpatialRef,` with the corresponding [spatialref that you want to assign to the created geometry object](https://github.com/JuliaGeo/GDAL.jl/blob/e55555527ceddbf8f2a856ba3a5bf5c046549c51/src/ogr_api.jl#L15).

yeesian: If I recall correctly, the geometries and spatialrefs in GDAL are not super well integrated and can lead to weird gotchas with projections, so I'd be hesitant about such an approach until you're certain about it.

rafaqz: Ok I've simplified this to exclude all `convert` methods. It only works on `reproject` and `importCRS`. If someone requires it for geometries it shouldn't be too much work - but I don't need it and don't have time to work around GDAL gotchas.

rafaqz:
This is now just `reproject` and `importCRS`. `:compliant` is the default so that nothing changes unless users specify `:trad` in either of those methods.

src/spatialref.jl
```diff
@@ -3,10 +3,16 @@
 
 Import a coordinate reference system from a `GeoFormat` into GDAL,
 returning an [`AbstractSpatialRef`](@ref).
+
+The `order` keyword argument can be passed in. `:trad` will use traditional
```
yeesian, line 7: Should we add this to the docstrings for `fromWKB`, `fromWKT`, `toWKT`, `toISOWKT`, ... as well?

src/spatialref.jl
```diff
@@ -3,10 +3,16 @@
 
 Import a coordinate reference system from a `GeoFormat` into GDAL,
 returning an [`AbstractSpatialRef`](@ref).
+
```
yeesian, line 6: Shall we be consistent in using 
```
## Keyword Argument
- `order`: `:trad` will use traditional lon/lat axis ordering in any actions done 
  with the crs. `:compliant`, the default will use axis ordering compliant with 
  the relevant CRS authority.
```
?

src/spatialref.jl
```diff
@@ -3,10 +3,16 @@
...
+The `order` keyword argument can be passed in. `:trad` will use traditional
+lon/lat axis ordering in any actions done with the crs. `:compliant`, the default
+will use axis ordering compliant with the relevent CRS authority.
```
yeesian, line 9: relevent -> relevant

src/spatialref.jl
```diff
@@ -15,28 +21,34 @@ Import a coordinate reference system from a `GeoFormat` into the spatial ref.
...
--`targetcrs`: The coordinate reference system to transform to, using any CRS capable `GeoFormat`
+- `coord`: Vector of Geometry points
+- `sourcecrs`: The current coordinate reference system, as a `GeoFormat`
+- `targetcrs`: The coordinate reference system to transform to, using any CRS capable `GeoFormat`
+
+
+## Keyword Argument
+- `order`: `:trad` will use traditional lon/lat axis ordering in any actions done 
+  with the crs. `:compliant`, the default will use axis ordering compliant with 
+  the relevent CRS authority.
```
yeesian, line 51: relevent -> relevant

src/spatialref.jl
```diff
@@ -15,28 +21,34 @@ Import a coordinate reference system from a `GeoFormat` into the spatial ref.
...
 Reproject points to a different coordinate reference system and/or format.
 
 ## Arguments
--`coord`: Vector of Geometry points
--`sourcecrs`: The current coordinate reference system, as a `GeoFormat`
--`targetcrs`: The coordinate reference system to transform to, using any CRS capable `GeoFormat`
+- `coord`: Vector of Geometry points
+- `sourcecrs`: The current coordinate reference system, as a `GeoFormat`
+- `targetcrs`: The coordinate reference system to transform to, using any CRS capable `GeoFormat`
+
```
yeesian, line 46: remove extra line

src/spatialref.jl
rafaqz: If you mean `importESRI`, `importXML` etc then yes. But I haven't implemented it for geometries due to your earlier concerns.

rafaqz:
Your comments are resolved and code/docs are cleaned up.

>Can we update https://github.com/yeesian/ArchGDAL.jl/blob/master/docs/src/projections.md with documentation on GeoFormat and these changes as well?

I barely have time to do this PR currently, I'm not sure when I would be able to do that.

rafaqz:
Can you merge this?

yeesian:
Thanks! Yeah the documentation is not a blocker, thanks for the ping!

rafaqz:
Actually good you waited a while as I found those extra bugs in the meantime :)

rafaqz:
Oh can we get a release as well? Trying to get GeoData.jl published.

PR #108, opened 2020-02-15
update a test for PROJ 6.3

visr:

PR #125, opened 2020-06-04
require julia 1.3 / GDAL build with artifacts

evetion:
To always make use of the new GDAL build system (jll). Should make support easier but bumps our Julia version from LTS to 1.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - 1.0
+  - 1.3
   - 1
   - nightly
 matrix:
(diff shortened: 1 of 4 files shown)


visr:
Yeah I think this would be good to add. I somehow regularly still end up with GDAL.jl 1.0 (pre jll) if I add ArchGDAL. And that build is much more broken than the jll. Requiring this version and perhaps retrocapping older versions should help there.

This PR still needs updating of the versions tested in CI. Coul you also remove the VERSION check from #108 and perhaps any others that may be present?

PR #128, opened 2020-06-19
Remove type specification from indices argument in rasterio.jl

### LINKED ISSUE #126: Argument type specifications in rasterio.jl too narrow
Hi everybody,
when I type:
```julia
dataset = ag.read(...)
img = ag.read(dataset, [1,2,3])
```
An error pops up telling me that there is no matching function signature, because I used Int64s to specify the bands. This is not optimal. Looking at the function definitions in rasterio.jl I see

```julia
rasterio!(...., bands::Vector{Cint},...) = ...
```
everywhere. However this specification is too narrow and should be replaced by <:Integer. The type conversion to Cint should then be handled right before the ccall. The same probably counts for write(), but I didnt look into that.

AlfTetzlaff:
The user now has more freedom of entering the bands he wishes to load, no conversion to Int32 is needed anymore. This all works now:
```julia
ds = ag.read("somefile")
img = ag.read(ds, 1)
img = ag.read(ds, [1,2,3,4])
img = ag.read(ds, Int8[1,2,3,4])
img = ag.read(ds, (1,2,3,4))
```

Fixes #126

### DIFF
--- a/src/raster/rasterio.jl
+++ b/src/raster/rasterio.jl
@@ -1,7 +1,7 @@
 
 """
-    rasterio!(dataset::AbstractDataset, buffer::Array{<:Real, 3}, bands::Vector{Cint}; <keyword arguments>)
-    rasterio!(dataset::AbstractDataset, buffer::Array{<:Real, 3}, bands::Vector{Cint}, rows, cols; <keyword arguments>)
+    rasterio!(dataset::AbstractDataset, buffer::Array{<:Real, 3}, bands::Vector{<:Integer}; <keyword arguments>)
+    rasterio!(dataset::AbstractDataset, buffer::Array{<:Real, 3}, bands::Vector{<:Integer}, rows, cols; <keyword arguments>)
     rasterio!(rasterband::AbstractRasterBand, buffer::Matrix{<:Real}; <keyword arguments>)
     rasterio!(rasterband::AbstractRasterBand, buffer::Matrix{<:Real}, rows, cols; <keyword arguments>)
 
@@ -64,7 +64,7 @@ function rasterio! end
 function rasterio!(
         dataset::AbstractDataset,
         buffer::Array{<:Real, 3},
-        bands::Vector{Cint},
+        bands::Vector{<:Integer},
         access::GDALRWFlag  = GDAL.GF_Read,
         pxspace::Integer    = 0,
         linespace::Integer  = 0,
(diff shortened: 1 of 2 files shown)


src/raster/rasterio.jl
```diff
@@ -422,14 +422,43 @@ for (T,GT) in _GDALTYPE
             (dataset == C_NULL) && error("Can't read invalid rasterband")
             xbsize, ybsize, zbsize = size(buffer)
             nband = length(bands)
+            bands = Cint.(collect(bands))
```
visr, line 425: Are you sure this is needed? `ccall` also does conversions as well, it doesn't only take `Vector{Int32}`.

AlfTetzlaff: You can e.g. change Ptr{Cint} to Ref{Cint} and pointer(bands) to bands, which I think should trigger conversions, but it nevertheless errors for iterable arguments. So I added the explicit conversion. But the cost of this line is marginal so I think it is ok to leave it; furthermore it enables all iterables as arguments e.g. tuples, which is quite a bonus :)

visr: Ah yeah I see, good points. Let's leave it in.

AlfTetzlaff:
I took a look into the tests, removed all the Cints and added a test for calling with a tuple.

src/raster/rasterio.jl
visr: It does seem like in this case that bands is already a `Vector{Cint}` this could introduce the first allocation in this method, which otherwise mutates the `buffer` argument.
How about something like this to avoid creating a copy when it's not needed?
```
bands isa Vector{Cint} ? bands : Cint.(collect(bands))
```

AlfTetzlaff: I introduced exactly that, which saves two allocations for all those using the old call signature. How performance critical is this section of the code? In my use case, I call read() a few hundred times at maximum in a non-timecritical context.

visr: For the large majority of users I imagine those allocations won't matter. But if we can easily support a way for power users to avoid those allocations I think it's worth it. For instance, now I'm working on a model that loads new geospatial forcing data for every timestep. If the model itself is fast, these allocations can start to matter.

AlfTetzlaff:
Thanks for merging; my first contribution to an official julia package by the way :) Let's hope I didn't introduce any weird bugs ;-)

visr:
Nice, congrats! First of many contributions I hope ;)

I also tagged a new release such that it is available to everyone.

Issue #40, opened 2018-05-05
Add raster example to documentation


yeesian:
prompted by #39

mkborregaard:
I'd be interested in a tutorial that describes how to:
1. open a raster file
2. get the projection
3. plot it
4. plot part of it
5. extract part of the raster based on a rectangular windor
6. extract part of it based on a shapefile polygon in the same projection
7. plot that with some spatial points overlaid
8. extract the raster's value under these points
9. use the raster as a standard matrix (if possible?)
10. apply image windowing functions to the raster using e.g. ImageFiltering.jl

I'm very willing to write this and contribute it myself, if you can get me started, possibly just with a list of functions to check out and a nod to the things that are possible.
I checked your links above, the invisibleroads one is dead, the RasterIO guide has some of the right things, the geos one is mainly about reprojection.
A sample raster file could be one of those from here http://chelsa-climate.org/downloads/ (e.g. the first), the polygon to clip after could be one frome https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/50m/cultural/ne_50m_admin_0_countries.zip 
. Can you help me get started?

pritamd47:
(Seems like I am 2 years late to the party) This seems like a wonderful suggestion, and I think I can implement these changes to the documentation. 
Even taking some examples from the `tests` would be a solution to what @mkborregaard is asking for. I think a solid reference point for such a _cook-book style_ guide is the [Python Quickstart guide by the Rasterio library (Python)](https://rasterio.readthedocs.io/en/latest/quickstart.html).

I would be happy to implement this, taking cues from the tests. Should I go ahead and open a PR for this?

visr:
Of for sure, I think having such example workflows available in the documentation would be a huge help to a lot of people, so yes please!

I agree the rasterio quick start is a good example. Maybe it would be a good idea to try to do a more or less exact replication of that workflow? With proper attribution of course.

I think the 11 points above, besides the first few, already go into a bit more depth, and could be a nice follow up project. What do you think?

pritamd47:
Sounds good! I will start by replicating the rasterio examples (with attribution). I will open a PR soon (trying out new things in Julia is really a relief during these times 😅).

PR #130, opened 2020-07-10
Introductory tutorial similar to Rasterio's Python Quickstart

pritamd47:
Addressing the discussion at #40, this PR is an attempt to add a tutorial in the documentation, similar to [Rasterio's Python Quickstart](https://rasterio.readthedocs.io/en/latest/quickstart.html).

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -9,7 +9,7 @@ makedocs(
     strict = true,
     pages = [
         "index.md",
-        "ArchGDAL Quickstart Guide" => "quickstart.md"
+        "ArchGDAL Quickstart Guide" => "quickstart.md",
         "GDAL Datasets" => "datasets.md",
         "Feature Data" => "features.md",
         "Raster Data" => "rasters.md",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -9,6 +9,7 @@ makedocs(
     strict = true,
     pages = [
         "index.md",
+        "ArchGDAL Quickstart Guide" => "quickstart.md"
         "GDAL Datasets" => "datasets.md",
         "Feature Data" => "features.md",
         "Raster Data" => "rasters.md",
--- a/docs/src/quickstart.md
+++ b/docs/src/quickstart.md
@@ -143,3 +143,38 @@ Z = 10.0 .* (Z2 .- Z1)
 ```
 The fictional field for this example consists of the difference of two Gaussian distributions and is represented by the array `Z`. Its contours are shown below.
 ![creating-dataset](https://user-images.githubusercontent.com/7526346/87633084-5bd5a900-c758-11ea-8fd3-548d039f1a43.png)
+
+Once we have the data to write into a raster dataset, before performing the write operation itself, it is a good idea to define the geotransform and the coordinate reference system of the resulting dataset. For North-up raster images, only the *pixel-size/resolution* and the coordinates of the *upper-left* pixel is required. 
+```Julia
+# Resolution
+res = (x[end] - x[1]) /240.0
+
+# Upper-left pixel coordinates
+ul_x = x[1] - res/2
+ul_y = y[1] - res/2
+
+gt = [ul_x, res, 0.0, ul_y, 0.0, res]
+```
+The coordinate reference system of the dataset needs to be a string in the WKT format. 
+```Julia
+crs = AG.toWKT(AG.importPROJ4("+proj=latlong"))
+```
+
+To write this array, first a dataset has to be created. This can be done using the `AG.create` function. The first argument defines the path of the resulting dataset. The following keyword arguments are hopefully self-explanatory. 
+
+Once inside the ```do...end``` block, the `write!` method can be used to write the array(`Z`), in the 1st band of the opened `dataset`. Next, the geotransform and CRS can be specified using the `setgeotransform!` and `setproj!` methods.
+```Julia
+AG.create(
+    "./temporary.tif",
+    driver = AG.getdriver("GTiff"), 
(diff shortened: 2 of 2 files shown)


visr:
Awesome :)

Don't forget to add the new page here, it should probably go right below the index: https://github.com/yeesian/ArchGDAL.jl/blob/47cc5564d23a7cde0591311908270444bc60643a/docs/make.jl#L11-L12

Minor nit: could you use uppercase `AG` instead of `ag` as the module alias? It's consistent with the rest of the docs and Julia conventions where modules start with upper case.

For the picture, if possible, I'd rather not have it stored in the repository, to keep it as small as possible. Could you rebase this commit to remove it? Let me know if I can help. A solution could be to just post them in the thread here, which will give a GitHub link that we can paste in the markdown:

example

```
![example](https://user-images.githubusercontent.com/4471859/87169013-a32ee600-c2cf-11ea-9d09-e82446812282.png)
```

I believe another option would be to have Documenter generate the pictures by running the code, but I haven't looked into this option before.

pritamd47:
> For the picture, if possible, I'd rather not have it stored in the repository, to keep it as small as possible. Could you rebase this commit to remove it? 

This would be great! However I must admit, I do not understand the rebase thing you talked about. Will deleting the `resources` directory not suffice?

visr:
If you delete the resources directory with `git rm docs/src/resources/example.png`, and add that as a new commit, the file will still be a part of the git history, making it bigger.

So to remove it from history, what you could do is `git rm docs/src/resources/example.png`, followed by `git commit --amend`. This modifies the commit you already made, instead of adding a new one. To push that here you need to do a force push, `git push -f`.

pritamd47:
> Adding plot for documentation
Creating-Data

pritamd47:
Hey, I am writing the final portion of the tutorial, involving [writing a dummy 2-D array to a raster dataset](https://rasterio.readthedocs.io/en/latest/quickstart.html#saving-raster-data), but I am not sure if what I am doing is the best way to do so (Probably not).

So currently I am making the array (`Z`) following the exact operations [here](https://rasterio.readthedocs.io/en/latest/quickstart.html#saving-raster-data). Now, I am writing the array using these commands:
```Julia
AG.create(
    AG.getdriver("GTiff"),
    width=240,
    height=180,
    nbands=1,
    dtype=Float32,
    filename="./temporary.tif") do dataset
        AG.write!(dataset, Z, 1)
end
```
This saves the array, but, is flipped around its Y axis. So to solve this, I flipped the array around its Y axis while writing, using `AG.write!(dataset, reverse(Z, dims=1), 1)`; BUT, now the resultant raster file is flipped around the X axis 😞. Then, just out of curiosity, I used `AG.write!(dataset, reverse(Z, dims=2), 1)`, and the resultant raster seems to be correct. I am confused about this behavior, and wanted to make sure that I personally understand it before writing any further.

Apart from this, I am also unsure about the best way to insert the transformation and projection information into the raster.

It would be great if you can provide an explanation of this. Thanks :)

visr:
Sorry I forgot about this (feel free to ping again).

GDAL is row-major, whereas Julia is column-major. The raster data that you get from ArchGDAL, and are expected to pass to ArchGDAL, is permuted compared to what you may expect, so rows become columns and columns become rows. This is done to avoid creating copies on every read an write, which may hurt performance. See also the notes in http://yeesian.com/ArchGDAL.jl/latest/rasters/#Reading-Raster-Values-1 and discussion in #37 (I had exactly the same initial reaction as you). So perhaps you can first `permutedims` your array before passing it to ArchGDAL?

> Apart from this, I am also unsure about the best way to insert the transformation and projection information into the raster.

I believe this should work with the `setproj!` and `setgeotransform!` functions, could you try out those? You can call `getproj` and `getgeotransform` to see what kind of data is expected.

pritamd47:
I think the PR is ready to be merged. Feel free to suggest any changes that you seem fit. @visr

visr:
Cool! Could you perhaps try using this syntax: https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@repl-block? This is also used here for instance: https://raw.githubusercontent.com/yeesian/ArchGDAL.jl/master/docs/src/datasets.md

It means that you can remove the output from the document, as Documenter will run the examples and paste in the output as the documentation builds. It's good to build the docs locally to confirm it is working as intended.

pritamd47:
From what I am understanding, @repl or @example will evaluate the script. However, for the first half of the documentation, I read in a LandSat image (quite huge, ~66MB). If there will be no such file (since such a large file cannot be included in the package) in the working directory, the evaluation should result in an error. Is there a way to download the file on-the-fly, so that the code blocks can be evaluated?

visr:
Yep, there is, and after #138 was fixed this works directly now:
```julia
ds = ArchGDAL.read("/vsicurl/http://landsat-pds.s3.amazonaws.com/c1/L8/037/034/LC08_L1TP_037034_20160712_20170221_01_T1/LC08_L1TP_037034_20160712_20170221_01_T1_B4.TIF")
```

For reading rasters, we could also use the new `readraster`, see https://yeesian.com/ArchGDAL.jl/dev/rasters/#The-RasterDataset-type. This could make the example a bit easier. Or we can at least make a mention of it.

Side note, when typing macro names in GitHub it is best to quote them with backticks like `@example`, to avoid pinging unrelated people.

pritamd47:
@visr please review the changes when you get the time. Thanks :)

yeesian:
This is awesome, thank you so much! I am really sorry it took me so long to get around to review this PR.

PR #145, opened 2020-10-04
fix too large buffer sizes on reduced reads

### LINKED ISSUE #78: How to read a subset of a multiband raster?
I've tried to read a subset but with several bands at once from a VRT. The documentation says this can be done using `ArchGDAL.read(dataset, indices, rows, cols)`
But when I do so:
```
    ArchGDAL.registerdrivers() do
        ArchGDAL.read(path_vrt) do dataset
             arr = ArchGDAL.read(dataset, Int32[1,2,3,4], 1:100, 1:100)
        end
    end
```
all columns and rows will be loaded.
What have I done wrong?

### LINKED ISSUE #132: incorrect array dimension returned by ArchGDAL.read(ds, i, xoff, yoff, xsize, ysize)
```julia
using ArchGDAL
infile = "2002-07-12.tif"

function read_tiff(file::String, options...)
    ArchGDAL.read(file) do dataset
        ArchGDAL.read(dataset, options...)
    end
end

r = read_tiff(infile, 1, 0, 0, 180, 90);
size(r) # (360, 460)
```
The expect dimensions of `r` is `[180, 90]`. However `[360, 460]` returned.

visr:
Fixes #132
Fixes #78, with the possible exception noted in https://github.com/yeesian/ArchGDAL.jl/issues/78#issuecomment-496205117:

> We'll have to be careful about how rasterio is invoked (for [linespace and bandspace](https://github.com/yeesian/ArchGDAL.jl/blob/9cc11c3f3caa57b57680725a356b24db8ff069c8/src/raster/rasterio.jl#L2-L55)) too

I'm not really sure how those `linespace` and `bandspace` work. Should `width(dataset), height(dataset)` be replaced by looking at the buffer sizes in that method? If not clear, let's just merge this and park this as an issue to look into later.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c9ce4bd3-c3d5-55b8-8973-c0e20141b8c3"
 keywords = ["GDAL", "IO"]
 license = "MIT"
 desc = "A high level API for GDAL - Geospatial Data Abstraction Library"
-version = "0.5.1"
+version = "0.5.2"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/raster/rasterio.jl
+++ b/src/raster/rasterio.jl
@@ -156,7 +156,7 @@ function read(
         xsize::Integer,
         ysize::Integer
     )
-    buffer = Array{pixeltype(rb)}(undef, width(rb), height(rb))
+    buffer = Array{pixeltype(rb)}(undef, xsize, ysize)
     rasterio!(rb, buffer, xoffset, yoffset, xsize, ysize)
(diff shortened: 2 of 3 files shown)


yeesian:
Thank you!

> Should width(dataset), height(dataset) be replaced by looking at the buffer sizes in that method?

Yes, from looking at the code, they really should. If it isn't too much trouble, can we add a unit test for
```julia
read(
    rb::AbstractRasterBand,
    xoffset::Integer,
    yoffset::Integer,
    xsize::Integer,
    ysize::Integer
)
```
too?

visr:
Ok, done!

PR #146, opened 2020-10-19
fix downloading test data on julia nightly

visr:
for some reason this function is no longer exported from Pkg.PlatformEngines


visr:
I see two unrelated failures, so I'll go ahead and merge:

doc build: curl: (22) The requested URL returned error: 429 too many requests

`ERROR: Unable to automatically install 'LibCURL' from '/home/travis/.julia/packages/LibCURL_jll/yitNc/Artifacts.toml'` does not happen on the same build but the push version on Travis.

PR #151, opened 2020-11-04
Update .gitignore

### LINKED ISSUE #150: Update .gitignore to include VS Code files and Manifest

zerefwayne:
closes #150 

.gitignore now ignores more files related to VS Code and Julia config files.


yeesian:
Sorry I'm a little reluctant to merge this since I haven't seen a similar setup in other julia packages (including https://github.com/julia-vscode/julia-vscode), and am still trying to get comfortable with something auto-generated from https://www.toptal.com/developers/gitignore/api/julia,vscode,sublimetext. Is this impacting you a great deal?

zerefwayne:
I checked out other Julia repositories and you are right, this is very different from them. I will close this one.

visr:
@zerefwayne most Julia repos only have Julia related files in the .gitignore. Since we don't know which editors people use, it's not convenient to add these editors to this file. Instead users of an editor can add the needed entries to the global gitignore file, see https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

PR #155, opened 2020-11-24
[test] Replace Pkg.PlatformEngine --> Custom download & verify

### LINKED ISSUE #154: download_verify removed on nightly
This package uses internals of Pkg when testing e.g.:

https://github.com/yeesian/ArchGDAL.jl/blob/9cc8713063a537f10cc2801907073f7cfd30c1c4/test/remotefiles.jl#L4-L6

These has been changed on Julia nightly. It would be good to transition these tests to use public functions (e.g. using the `download` function and then verify using the `SHA` stdlib).

zerefwayne:
Signed-off-by: zerefwayne <aayushjog@gmail.com>

closes #154 

- Replaced the internal functions of Pkg with custom functions with only the functionalities we require. The implementation is inspired by [Pkg.PlatformEngines](https://github.com/JuliaLang/Pkg.jl/blob/master/src/PlatformEngines.jl) and uses the upcoming `Downloads.download()` which will be the de-facto from Julia v1.6 

- Adds `Downloads` and `SHA` to Test dependencies.

- No new test cases break reported.

  Before the Change: `1226 passed, 12 failed, 0 errored, 0 broken.`
  After the Change: `1226 passed, 12 failed, 0 errored, 0 broken.`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,9 +23,11 @@ julia = "1.3"
 
 [extras]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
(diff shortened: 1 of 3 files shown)


yeesian:
Sorry for the late response -- is this PR ready to be reviewed? I saw the test failures on Travis and wasn't sure if you still have work-in-progress. I dug into the logs on Travis and think they were due to flakey tests that @visr pointed out in #158 and have updated the test values in #157. I think this PR might work if you rebase it on master.

yeesian:
This has been incorporated into #179, thank you for the contribution!

PR #156, opened 2020-12-10
Allow test failures on julia nightly

### LINKED ISSUE #134: online docs are old
https://yeesian.com/ArchGDAL.jl/stable/ seems to be v0.3.1, that should be v0.5.0. Not sure why

There hasn't been a commit on gh-pages for a long time: https://github.com/yeesian/ArchGDAL.jl/tree/gh-pages

With the doc updates in c22481854e3dfce16070d8b379de151dbbcf39a0 at least my local doc build was working again, but there are seemingly other issues also.

We probably need to go over https://juliadocs.github.io/Documenter.jl/stable/man/hosting/ and make sure things are ok. Can't verify if the DOCUMENTER_KEY secret is set though.

felixcremer:
This should allow failures to happen on julia nightly. 
I think this is the only reason, why the docs haven't been built for the releases 0.5.1 and 0.5.2.
Should be the next step to close #134

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,5 @@ jobs:
         - julia --project=docs/ test/remotefiles.jl
         - julia --project=docs/ docs/make.jl
       after_success: skip
+  allow_failures:
+    - julia: nightly


visr:
Thanks! Looks like some of the projection tests need updating since PROJ_jll v7.2.0+0 is now used. Will the doc update work with failing tests? Otherwise it would be good to update those as well.

Besides this we should probably migrate from Travis CI to GitHub Actions, but that can be done separately.

felixcremer:
The doc update will not work with failing tests. And in the last release, these failed due to the download_verify problems in nightly. Maybe you can restart the build in travis with a changed config, where you would allow failures on nightly. 

I saw the projection failures, but I haven't looked in depth in there and don't know the code well enough to fix them.

visr:
> I saw the projection failures, but I haven't looked in depth in there and don't know the code well enough to fix them.

Basically the new PROJ version probably has slightly more accurate coordinate transformations, so the decimals need to be updated somewhere behind the comma, for instance this failure:

```
Method 1: Test Failed at /home/travis/build/yeesian/ArchGDAL.jl/test/test_cookbook_projection.jl:13
  Expression: GeoInterface.coordinates(point) ≈ [47.348801380288485, -122.5981351308777]
   Evaluated: [47.3488070138318, -122.5981499431438] ≈ [47.348801380288485, -122.5981351308777]
```

So this just means we can update the test values in `test_cookbook_projection.jl:13`:

```diff
-GeoInterface.coordinates(point) ≈ [47.348801380288485, -122.5981351308777]
+GeoInterface.coordinates(point) ≈ [47.3488070138318, -122.5981499431438]
```

PR #157, opened 2020-12-27
update test values for GDAL 3.2.0

visr:
With GEOS 3.9.0 and PROJ 7.2.0.
This consists of small projection differences due to new PROJ, and geometry ordering changes due to GEOS.


yeesian:
I agree with #158 -- sorry for the brittle tests (that you have to update in this PR), and thank you for doing it anyway!

PR #159, opened 2021-01-01
Compact display of geometry without WKT

evetion:
Getting the WKT of a (complex) geometry is non trivial and will slow down things especially when used in Tables.

With this PR:

When used in a Table (compact)
```julia
julia> df = DataFrame(geom=createpoint.(coords), name="test")
10×2 DataFrame
 Row │ geom                name
     │ IGeometr…           String
─────┼────────────────────────────
   1 │ Geometry: wkbPoint  test
   2 │ Geometry: wkbPoint  test
   3 │ Geometry: wkbPoint  test
   4 │ Geometry: wkbPoint  test
   5 │ Geometry: wkbPoint  test
   6 │ Geometry: wkbPoint  test
   7 │ Geometry: wkbPoint  test
   8 │ Geometry: wkbPoint  test
   9 │ Geometry: wkbPoint  test
  10 │ Geometry: wkbPoint  test
```

Individual or array (non-compact) usage
```julia
julia> df[1, :geom]
Geometry: POINT (0.394776848700807 0.244117475869569)
```

### DIFF
--- a/test/test_display.jl
+++ b/test/test_display.jl
@@ -19,6 +19,8 @@ import ArchGDAL; const AG = ArchGDAL
              Field 0 (FID): [OFTReal], 2.0, 3.0, 0.0, 3.0
              Field 1 (pointname): [OFTString], point-a, point-b, a, b
         """
+        @test sprint(print, AG.nextnamedtuple(layer), context=:compact => true) == 
+        "(FID = 2.0, pointname = \"point-a\",  = Geometry: wkbPoint)"
         @test sprint(print, AG.layerdefn(layer)) == """
           Geometry (index 0):  (wkbPoint)
              Field (index 0): FID (OFTReal)
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c9ce4bd3-c3d5-55b8-8973-c0e20141b8c3"
 keywords = ["GDAL", "IO"]
 license = "MIT"
 desc = "A high level API for GDAL - Geospatial Data Abstraction Library"
-version = "0.5.2"
+version = "0.5.3"
 
 [deps]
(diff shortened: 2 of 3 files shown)


evetion:
We could also completely disable WKT printing?

visr:
Could you add a test here (with/without compact) and earn that coverage increase? ;)
https://github.com/yeesian/ArchGDAL.jl/blob/f56f79c23b47e43d968605b053a531cf902991ae/test/test_display.jl#L66-L67

Otherwise this looks good to me. Not sure why we'd want to disable WKT printing altogether.

evetion:
> Otherwise this looks good to me. Not sure why we'd want to disable WKT printing altogether.

It's unasked for and it's non-constant in time/memory. It's a "nice" thing that works well for a `Point(1,2)`, but quickly becomes slow for a large, complex geometry.

I would agree it's better than just saying `IGeometry`, but that's why I made #160.

test/test_display.jl
```diff
@@ -19,6 +19,8 @@ import ArchGDAL; const AG = ArchGDAL
              Field 0 (FID): [OFTReal], 2.0, 3.0, 0.0, 3.0
              Field 1 (pointname): [OFTString], point-a, point-b, a, b
         """
+        @test sprint(print, AG.nextnamedtuple(layer), context=:compact => true) == 
+        "(FID = 2.0, pointname = \"point-a\",  = Geometry: wkbPoint)"
```
visr, line 23: Thought this looks funny, but that's an empty geometry name of course.

visr (approved):

test/test_display.jl
evetion: Yeah, I was suprised a `Symbol("")` was actually allowed as key in the beginning.

PR #160, opened 2021-01-01
Give geometry type hints in Schema.

evetion:
Also rename the empty `Symbol("")` to `:geometry`.


src/tables.jl
```diff
@@ -63,7 +62,8 @@ function schema_names(layer::AbstractFeatureLayer)
     featuredefn = layerdefn(layer)
     fielddefns = (getfielddefn(featuredefn, i) for i in 0:nfield(layer)-1)
     field_names = (Symbol(getname(fielddefn)) for fielddefn in fielddefns)
-    geom_names = (Symbol(getname(getgeomdefn(featuredefn, i-1))) for i in 1:ngeom(layer))
+    geom_names = collect(Symbol(getname(getgeomdefn(featuredefn, i-1))) for i in 1:ngeom(layer))
+    replace!(geom_names, Symbol("")=>Symbol("geom"), count=1)
```
evetion, line 66: This could theoretically clash with an existing identifier, but I think it's very unlikely.

evetion:
I have now also piggybacked the geometry type unto the `IGeometry`, so it becomes `IGeometry{OGRwkbGeometryType}`. This aligns it with the schema changes and makes it less of a black box. This also enables dispatch on specific geometry types.

src/types.jl
```diff
@@ -211,11 +211,11 @@ mutable struct Geometry <: AbstractGeometry
     Geometry(ptr::GDALGeometry = C_NULL) = new(ptr)
 end
 
-mutable struct IGeometry <: AbstractGeometry
+mutable struct IGeometry{OGRwkbGeometryType} <: AbstractGeometry
```
evetion, line 214: It's a mutable struct, are there instances where we change (from a NULL pointer) to a new geometry type?

src/tables.jl
visr: Still would probably be good to check. So most files have a single geometry that has no name?

I'm thinking a bit about `geom` versus `geometry`. The latter is a bit longer, but may be more clear, and seems to be more in line with sf and geopandas.

visr:
> I have now also piggybacked the geometry type unto the `IGeometry`, so it becomes `IGeometry{OGRwkbGeometryType}`. This aligns it with the schema changes and makes it less of a black box. This also enables dispatch on specific geometry types.

This is unrelated to the other change here and should be in a separate PR. I can understand it may be convenient, and if it would have all been Julia code, this certainly would make sense, but it's a pointer. Sticking [an enum](https://github.com/JuliaGeo/GDAL.jl/blob/204a0fc71e07b80adee12ec72acca05f6bc538d9/src/common.jl#L1250-L1254) on as a type parameter will cause it to compile different code for each geometry type, whereas from GDAL there is no need for that. And I don't really know other examples of using Enums as type parameters.

evetion:
> The geometry naming is unrelated to this change here and should be in a separate PR.

It is actually. If you supply a Schema with `A{x}`, the rows should also be of type `A{x}`.

> I can understand it may be convenient, and if it would have all been Julia code, this certainly would make sense, but it's a pointer. Sticking [an enum](https://github.com/JuliaGeo/GDAL.jl/blob/204a0fc71e07b80adee12ec72acca05f6bc538d9/src/common.jl#L1250-L1254) on as a type parameter will cause it to compile different code for each geometry type, whereas from GDAL there is no need for that. And I don't really know other examples of using Enums as type parameters.

GDAL doesn't need it, but it's also not Julian. Geometry is now just a black box, just as saying Array. But we do constantly use Vector or specialize even further. This Enum is just integers and thus `isbits`.

Maybe it's something for further downstream, or should be purely be implemented via GeoInterface.

src/tables.jl
evetion: Yeah, that happens a lot. And that's inconvenient for accessing, as your normal access pattern `a.name` won't work. I'm fine with `geometry`.

src/types.jl
visr: Not that I know off.

visr:
Yeah I'm somewhat on the fence about this. What do you think, @yeesian? My first gut feeling says wrapping it downstream and adding a type parameter in a place where it is actually useful would be better, but would be good to get more input from others.

So this will require a call to `ogr_g_getgeometrytype` for every geometry that is created. Is that sufficiently fast not to cause slowdowns in workflows that iterate over many geometries? And if we would do this, shouldn't we do it for both Geometry and IGeometry?

evetion:
I understand, I'm uncertain about this PR myself as well. Maybe I can iterate on a design where the Schema doesn't influence the underlying types, as we have to expose it somewhere if we don't want the getgeomtype call on each row. 

Something like the new unreleased GeoInterface that also works on a Schema? You'd still end up with a `Vector{Geometry}` though?

src/tables.jl
yeesian: I'm good with `geometry` too

src/types.jl
yeesian: Not that I know of either.

But I'm curious might the `OGRwkbGeometryType` in the type potentially drift from the underlying pointer? I.e. that `typeof(geom) != ArchGDAL.getgeomtype(geom)` per https://github.com/yeesian/ArchGDAL.jl/blob/f56f79c23b47e43d968605b053a531cf902991ae/src/ogr/geometry.jl#L301

It seems that might be possible with the `ArchGDAL.forceto` method: https://github.com/yeesian/ArchGDAL.jl/blob/f56f79c23b47e43d968605b053a531cf902991ae/src/ogr/geometry.jl#L138-L145

yeesian (commented):
> My first gut feeling says wrapping it downstream and adding a type parameter in a place where it is actually useful would be better, but would be good to get more input from others.

I kinda agree, and think we might run into similar issues as the ones we had with GeoInterface before moving towards a traits/interface-like approach in GeoInterfaceRFC.

> I understand, I'm uncertain about this PR myself as well. Maybe I can iterate on a design where the Schema doesn't influence the underlying types, as we have to expose it somewhere if we don't want the getgeomtype call on each row.

Overall I really like the move towards providing geometry type information. I think we just have to think harder about the object/type-hierarchy model in ArchGDAL, and how it'll relate to the underlying GDAL objects.

src/tables.jl
yeesian: > So most files have a single geometry that has no name?

Yeah, that's the historical mental model people had of features and geometries. It seems like there's some effort to move away from that towards properly supporting features with multiple geometries (across GDAL and geopandas), but it seems that those are more experimental, and I haven't seen any (i) good documentation on a coherent model of how it should work and (ii) evidence that rigorous tests have been introduced to ensure that it'll obey those contracts.

visr: It seems like we're getting more packages with geometry columns, and I think it would be good to all give them the same name. I think preference would be `geometry`. We now also need to pick something here: https://github.com/JuliaGeo/Shapefile.jl/pull/49. Only I see https://github.com/evetion/GeoDataFrames.jl uses `geom`. @evetion do you think it would be good to change it there as well?

And much less important, but for visual consistency, do we put the geometry column at the beginning or at the end? I see GeoDataFrames puts it at the end. Since it is generally not very interesting to look at text representations of geometries, you don't want it to take up to much space. Luckily #159 also helps with that. One reason to perhaps bring it to the front, is because at the end it may fall off, and it is an important column, just look at the beautiful DataFrame I just made in my narrow shell:

```
julia> DataFrame(sadfasdfsd=1:3, asdfasdgaererer="aahweifhwe", asdfsfe=3:5, as
dfsf3e=3:5, asdf2sfe=3:5, asdfs3634fe=3:5, asdfsf63e=3:5, asdfsf123e=3:5, asdf
sf21e=3:5, lastcol=3:5)
3×10 DataFrame
 Row │ sadfasdfsd  asdfasdgaererer  asdfsfe  asdfsf3e  asdf2sfe  asdfs3634fe ⋯
     │ Int64       String           Int64    Int64     Int64     Int64       ⋯
   1 │          1  aahweifhwe             3         3         3            3 ⋯
   2 │          2  aahweifhwe             4         4         4            4
   3 │          3  aahweifhwe             5         5         5            5
                                                             4 columns omitted
```

yeesian: +1 for putting the geometry in the first few columns for the reasons @visr mentioned

src/types.jl
yeesian: Circling back on this: I'm guessing in those cases we could return a new geometry of the correspond type, so that isn't an issue.

src/tables.jl
evetion: I'm happy to change `geom` to `geometry`. I have no preference for the order of columns.

evetion:
With the recent comments, I'm happy to change the name to `geometry` and make sure that `forceto` also updates the type in this PR. But is it something we want to have merged? Do we also want it for `Geometry`?

We can also add some more tests, specifically:
- Behaviour for common files, especially mixing different geoms in the same vector.
- Performance of setting the type

This would be a breaking change.

yeesian:
> But is it something we want to have merged? Do we also want it for Geometry?

Yes and yes :)

> This would be a breaking change.

I'm trying to get in all the breaking changes (see e.g. #179) to settle on a design we're willing to maintain, before putting out a 1.0 release.

evetion:
Ok, I've updated this PR with the requested changes. `forceto` is already handled, because it initializes a new (I)Geometry. I've updated several tests.

To further optimize, I also made the `getgeomtype` use the type instead of going through GDAL.

```julia
julia> g = AG.createlinestring([(8.,1.),(9.,1.),(9.,2.),(8.,2.)])
Geometry: LINESTRING (8 1,9 1,9 2,8 2)

julia> @btime AG.GDAL.ogr_g_getgeometrytype(g.ptr)  # once on construction
  86.607 ns (3 allocations: 48 bytes)
wkbLineString::OGRwkbGeometryType = 0x00000002

julia> @btime AG.getgeomtype(g)
  15.215 ns (0 allocations: 0 bytes)
wkbLineString::OGRwkbGeometryType = 0x00000002
```

The performance impact should be neglible.

Any further ideas about testing this? I'm thinking of some shapefiles with mixed/multi geometry and see how we handle that in the tables interface. I'm also thinking about the impact of having a `Vector{Geometry}` now becoming a `Vector{Geometry{WKBthing}}` and how that would impact anything. My feeling is that this could be breaking to some packages, but the fixing/maintenance of those packages will be improved by this change.

yeesian (approved):

yeesian:
> Any further ideas about testing this? I'm thinking of some shapefiles with mixed/multi geometry and see how we handle that in the tables interface.

I saved the following as `tmp.geojson`:
```
{ "type": "FeatureCollection", "features": 
    [{ "type": "Feature", "geometry": 
        { "type": "GeometryCollection", "geometries": 
            [{ "type": "MultiLineString", "coordinates": [
                    [[-4.66065, 19.70386],[-4.66053, 19.70377],[-4.66026, 19.70358],[-4.66011, 19.70349],[-4.66009, 19.70352],[-4.66008, 19.70358]],
                    [[-4.66008, 19.70358],[-4.66009, 19.70352],[-4.66011, 19.70349],[-4.66026, 19.70358],[-4.66053, 19.70377],[-4.66065, 19.70386]]
                ] }, { "type": "Point", "coordinates": [-4.66012, 19.70362] }
            ] }, 
                "id": "1",
                "properties": { "feature_count": 3 } }, 
    { "type": "Feature", "geometry": 
        { "type": "LineString", "coordinates": [
                [-4.60419, 19.76303],[-4.60456, 19.7629],[-4.60481, 19.76260]
            ] }, 
            "id": "2",
            "properties": { "feature_count": 1 }
    }] 
}
```

And then read it using the Tables interface in this PR:
```julia
julia> AG.read("tmp.geojson")
GDAL Dataset (Driver: GeoJSON/GeoJSON)
File(s):
  tmp.geojson

Number of feature layers: 1
  Layer 0: tmp (wkbUnknown)


julia> AG.Table(AG.getlayer(ans, "tmp"))
Table with 2 features


julia> t = ans
Table with 2 features


julia> t
Table with 2 features


julia> t[1]
(id = "1", feature_count = 3, geometry = Geometry: GEOMETRYCOLLECTION (MULTILINESTRING ((-4.66065 19. ... 362)))

julia> t[2]
(id = "2", feature_count = 1, geometry = Geometry: LINESTRING (-4.60419 19.76303,-4.60456 19.7629,-4. ... 7626))

julia> t[1].id
"1"

julia> t.
feature_count geometry       id


julia> t.geometry
2-element Vector{ArchGDAL.IGeometry}:
 Geometry: GEOMETRYCOLLECTION (MULTILINESTRING ((-4.66065 19. ... 362))
 Geometry: LINESTRING (-4.60419 19.76303,-4.60456 19.7629,-4. ... 7626)
```

yeesian:
Overall LGTM. Letting @visr review before merging if there are no remaining concerns.

visr (approved):
Looks good to me! I still updated the column ordering as discussed in https://github.com/yeesian/ArchGDAL.jl/pull/160#discussion_r620407805.

visr:
Test failures seem to be unrelated, it is a test with a URL import. Also fails locally. Perhaps the libcurl build is not working well?

visr:
Ha, I just loaded a GeoPackage using this branch, and the geometry column was still "geom". Right now we only rename "" to "geometry". Turns out the default `GEOMETRY_NAME` for GeoPackage is "geom".  Not that that is standard though, PostGIS has:

GEOMETRY_NAME: Set name of geometry column in new table. If omitted it defaults to wkb_geometry for GEOM_TYPE=geometry, or the_geog for GEOM_TYPE=geography.

And MySQL for instance:

The geometry column, which defaults to SHAPE and can be overridden with the GEOMETRY_NAME layer creation option

https://gdal.org/drivers/vector/gpkg.html
https://gdal.org/drivers/vector/pg.html
https://gdal.org/drivers/vector/mysql.html

yeesian:
Thanks for the investigation @visr! To keep things moving here: I think we should continue to allow for the layers to provide geometry column names (if they exist), so the main question pertains to the behavior for empty geometry column names.

On the surface, I think renaming them (from `""` to `"geometry"`) is still sensible (to align with expectations for Tables), but I don't think users should be dependent on the name`geometry` for the reasons that you've mentioned -- what do you think?

On a side note, we should probably add this to the documentation (be it as part of this PR, or a separate one).

visr:
Yeah fully agree. Reason I wrote it down was that for a moment I though, wait does GDAL standardize on `"geom"`, so was I wrong to suggest renaming to `"geometry"`? But this does not appear the case. Besides the docs, we could think of adding a convenience function that returns the geometry column names. But that doesn't have to hold up this PR.

yeesian:
Thank you both for this PR! :)

yeesian:
@evetion 

To follow up on https://github.com/yeesian/ArchGDAL.jl/pull/160#issuecomment-837952028 (quoted below):

> I'm thinking of some shapefiles with mixed/multi geometry and see how we handle that in the tables interface. I'm also thinking about the impact of having a Vector{Geometry} now becoming a Vector{Geometry{WKBthing}} and how that would impact anything.

It raises questions whether vectors like `IGeometry{wkbUnknown}[IGeometry{wkbPoint}, IGeometry{wkbLineString}, ...]` are valid? On one hand, our intuition is that they should be, but on the other, `IGeometry{wkbUnknown}` isn't a supertype of the rest.

This was noticed in https://github.com/yeesian/ArchGDAL.jl/pull/179#issuecomment-845654042, so I've introduced a conversion method for the other geometry types to `wkbUnknown` in https://github.com/yeesian/ArchGDAL.jl/commit/0ca1a7d0242aa3e62ddeba1a957a398a965c89ce as a fallback behavior.

Issue #164, opened 2021-01-21
Allow gdalbuildvrt for interactive Dataset


felixcremer:
For two tifs I can currently use the following code to get a vrt:
```julia
using ArchGDAL
const AG=ArchGDAL
AG.read(group[1]) do ds1
    AG.read(group[2]) do ds2
	ArchGDAL.gdalbuildvrt([ds1, ds2]) do vrt
		AG.write(vrt,tmppath)
	end
    end
end
```
Here group is a vector of two paths to .tif files. 

But I would like  to build large vrts from Sentinel-1 time series, where I am combining one hundred tifs into one vrt. 
When I change the definition of unsafe_gdalbuildvrt to also accept vectors of AbstractDatasets, I can do the following:
```julia
   tifs = AG.read.(inputpaths)
   vrt_sep = AG.unsafe_gdalbuildvrt(tifs, ["-separate"])
```
I think, that the second version is much clearer to read, especially for lots of raster files. 

I am not sure, whether I would get any problems with memory leaks due to the unsafe function.
Can we define a safe version of gdalbuildvrt for interactive datasets?

Rapsodia86:
Try with GDAL package. 
options = GDAL.gdalbuildvrtoptionsnew(["-separate"], C_NULL)
outvrt = "U:/data/stacked.vrt"
ds_buildvrt = GDAL.gdalbuildvrt(outvrt, length(tifs), C_NULL,tifs, options, C_NULL) # 'tifs' is a vector with paths
GDAL.gdalbuildvrtoptionsfree(options)
GDAL.gdalclose(ds_buildvrt)
#and then read
outds =  ArchGDAL.readraster(outvrt)

visr:
> When I change the definition of unsafe_gdalbuildvrt to also accept vectors of AbstractDatasets

That sounds right, a PR would be appreciated! In fact, unless I'm missing something, it looks like all the functions that still require `::Dataset` should really be `::AbstractDataset`.

> I am not sure, whether I would get any problems with memory leaks due to the unsafe function.

If you do it in a function, such that GC will get triggered when it is done, all the IDatasets will be cleaned up due to the finalizers that are defined on them, so that shouldn't be an issue.

yeesian:
This has been fixed.

Issue #142, opened 2020-09-17
only unsafe functions in docstrings


visr:
There is a nice docstring for `unsafe_gdalwarp` here: https://github.com/yeesian/ArchGDAL.jl/blob/0c57b987082aec3dbe2999ec5e5a729e0ebdc3fc/src/utilities.jl#L46-L58. It is in the documentation here:
https://yeesian.com/ArchGDAL.jl/dev/reference/#ArchGDAL.unsafe_gdalwarp

This code wraps a list of unsafe functions in a finalizer, to make them safe, so `gdalwarp` is defined here:
https://github.com/yeesian/ArchGDAL.jl/blob/0c57b987082aec3dbe2999ec5e5a729e0ebdc3fc/src/context.jl#L192-L222

`gdalwarp` is probably the version most people should use. Right now it has no docstring however. We should find a way to give them docstrings as well, and perhaps even leave out the docstrings of the unsafe versions in the docs.

PR #167, opened 2021-01-22
RFC: Allow to use interactive Datasets in unsafe_gdal functions in utilities.jl

felixcremer:
As suggested by @visr in #164 this widens the expected types in the unsafe_gdal* functions to AbstractDataset, so that they would be usable with interactive Datasets. 
I am using parametric functions, so that we reuse the input type as return type to prevent the accidental mixing of the different approaches. 

I added tests for these functions by copying the tests in test_gdalutilities.jl and change the do block notation to variable assignment. 

Should we also provide these functions without the unsafe keyword by defining:
```julia
gdalfunc(IDataset, options; kwargs) = unsafe_gdalfunc(IDataset, options; kwargs)
```

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -1,5 +1,5 @@
 """
-    gdalinfo(dataset::Dataset, options = String[])
+    gdalinfo(dataset::AbstractDataset, options = String[])
 
 List various information about a GDAL supported raster dataset.
 
@@ -11,7 +11,7 @@ List various information about a GDAL supported raster dataset.
 ### Returns
 String corresponding to the information about the raster dataset.
 """
-function gdalinfo(dataset::Dataset, options = String[])
+function gdalinfo(dataset::AbstractDataset, options = String[])
     options = GDAL.gdalinfooptionsnew(options, C_NULL)
     result = GDAL.gdalinfo(dataset.ptr, options)
     GDAL.gdalinfooptionsfree(options)
@@ -19,7 +19,7 @@ function gdalinfo(dataset::Dataset, options = String[])
 end
 
 """
-    unsafe_gdaltranslate(dataset::Dataset, options = String[]; dest = "/vsimem/tmp")
+    unsafe_gdaltranslate(dataset::AbstractDataset, options = String[]; dest = "/vsimem/tmp")
 
 Convert raster data between different formats.
 
@@ -32,7 +32,7 @@ Convert raster data between different formats.
 The output dataset.
 """
 function unsafe_gdaltranslate(
-        dataset::Dataset,
+        dataset::AbstractDataset,
         options = String[];
         dest = "/vsimem/tmp"
     )
@@ -44,7 +44,7 @@ function unsafe_gdaltranslate(
 end
 
 """
-    unsafe_gdalwarp(datasets::Vector{Dataset}, options = String[]; dest = "/vsimem/tmp")
+    unsafe_gdalwarp(datasets::Vector{<:AbstractDataset}, options = String[]; dest = "/vsimem/tmp")
 
 Image reprojection and warping function.
 
@@ -57,7 +57,7 @@ Image reprojection and warping function.
 The output dataset.
 """
 function unsafe_gdalwarp(
-        datasets::Vector{Dataset},
+        datasets::Vector{<:AbstractDataset},
         options = String[];
         dest = "/vsimem/tmp"
     )
@@ -70,7 +70,7 @@ function unsafe_gdalwarp(
 end
 
 """
-    unsafe_gdalvectortranslate(datasets::Vector{Dataset}, options = String[]; dest = "/vsimem/tmp")
+    unsafe_gdalvectortranslate(datasets::Vector{<:AbstractDataset}, options = String[]; dest = "/vsimem/tmp")
 
 Convert vector data between file formats.
 
@@ -83,7 +83,7 @@ Convert vector data between file formats.
 The output dataset.
 """
 function unsafe_gdalvectortranslate(
-        datasets::Vector{Dataset},
+        datasets::Vector{<:AbstractDataset},
         options = String[];
         dest = "/vsimem/tmp"
     )
@@ -96,7 +96,7 @@ function unsafe_gdalvectortranslate(
 end
 
 """
-    unsafe_gdaldem(dataset::Dataset, processing::String, options = String[]; dest = "/vsimem/tmp", colorfile)
+    unsafe_gdaldem(dataset::AbstractDataset, processing::String, options = String[]; dest = "/vsimem/tmp", colorfile)
 
 Tools to analyze and visualize DEMs.
 
@@ -115,7 +115,7 @@ Tools to analyze and visualize DEMs.
 The output dataset.
 """
 function unsafe_gdaldem(
-        dataset::Dataset,
+        dataset::AbstractDataset,
         processing::String,
         options = String[];
         dest = "/vsimem/tmp",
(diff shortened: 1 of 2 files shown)


visr:
Great, thanks!

> Should we also provide these functions without the unsafe keyword by defining:

This should already be done automatically. Try leaving out `unsafe_` in the extra tests you added. Note also #142, because these functions have some discoverability issues that would be good to resolve.

src/utilities.jl
```diff
@@ -32,19 +32,19 @@ Convert raster data between different formats.
...
         options = String[];
         dest = "/vsimem/tmp"
-    )
+    ) where {T<:AbstractDataset}
     options = GDAL.gdaltranslateoptionsnew(options, C_NULL)
     usage_error = Ref{Cint}()
     result = GDAL.gdaltranslate(dest, dataset.ptr, options, usage_error)
     GDAL.gdaltranslateoptionsfree(options)
-    return Dataset(result)
+    return T(result)
```
yeesian, line 43: This should still be `Dataset(result)` rather than `T(result)` (see https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084 for context).

src/utilities.jl
```diff
@@ -57,20 +57,20 @@ Image reprojection and warping function.
...
         dest = "/vsimem/tmp"
-    )
+    ) where T<:AbstractDataset
     options = GDAL.gdalwarpappoptionsnew(options, C_NULL)
     usage_error = Ref{Cint}()
     result = GDAL.gdalwarp(dest, C_NULL,
         length(datasets), [ds.ptr for ds in datasets], options, usage_error)
     GDAL.gdalwarpappoptionsfree(options)
-    return Dataset(result)
+    return T(result)
```
yeesian, line 69: This should still be `Dataset(result)` rather than `T(result)` (see https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084 for context).

src/utilities.jl
```diff
@@ -83,20 +83,20 @@ Convert vector data between file formats.
...
         dest = "/vsimem/tmp"
-    )
+    ) where T<:AbstractDataset
     options = GDAL.gdalvectortranslateoptionsnew(options, C_NULL)
     usage_error = Ref{Cint}()
     result = GDAL.gdalvectortranslate(dest, C_NULL, length(datasets),
         [ds.ptr for ds in datasets], options, usage_error)
     GDAL.gdalvectortranslateoptionsfree(options)
-    return Dataset(result)
+    return T(result)
```
yeesian, line 95: This should still be `Dataset(result)` rather than `T(result)` (see https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084 for context).

src/utilities.jl
```diff
@@ -129,11 +129,11 @@ function unsafe_gdaldem(
     result = GDAL.gdaldemprocessing(dest, dataset.ptr, processing, colorfile,
         options, usage_error)
     GDAL.gdaldemprocessingoptionsfree(options)
-    return Dataset(result)
+    return T(result)
```
yeesian, line 132: This should still be `Dataset(result)` rather than `T(result)` (see https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084 for context).

src/utilities.jl
```diff
@@ -146,19 +146,19 @@ Convert nearly black/white borders to exact value.
...
         options = String[];
         dest = "/vsimem/tmp"
-    )
+    ) where T<:AbstractDataset
     options = GDAL.gdalnearblackoptionsnew(options, C_NULL)
     usage_error = Ref{Cint}()
     result = GDAL.gdalnearblack(dest, C_NULL, dataset.ptr, options, usage_error)
     GDAL.gdalnearblackoptionsfree(options)
-    return Dataset(result)
+    return T(result)
```
yeesian, line 157: This should still be `Dataset(result)` rather than `T(result)` (see https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084 for context).

src/utilities.jl
```diff
@@ -171,19 +171,19 @@ Create a raster from the scattered data.
...
         options = String[];
         dest = "/vsimem/tmp"
-    )
+    ) where T<:AbstractDataset
     options = GDAL.gdalgridoptionsnew(options, C_NULL)
     usage_error = Ref{Cint}()
     result = GDAL.gdalgrid(dest, dataset.ptr, options, usage_error)
     GDAL.gdalgridoptionsfree(options)
-    return Dataset(result)
+    return T(result)
```
yeesian, line 182: This should still be `Dataset(result)` rather than `T(result)` (see https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084 for context).

src/utilities.jl
```diff
@@ -196,19 +196,19 @@ Burn vector geometries into a raster.
...
         options = String[];
         dest = "/vsimem/tmp"
-    )
+    ) where T<:AbstractDataset
     options = GDAL.gdalrasterizeoptionsnew(options, C_NULL)
     usage_error = Ref{Cint}()
     result = GDAL.gdalrasterize(dest, C_NULL, dataset.ptr, options, usage_error)
     GDAL.gdalrasterizeoptionsfree(options)
-    return Dataset(result)
+    return T(result)
```
yeesian, line 207: This should still be `Dataset(result)` rather than `T(result)` (see https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084 for context).

src/utilities.jl
```diff
@@ -221,14 +221,14 @@ Build a VRT from a list of datasets.
...
         dest = "/vsimem/tmp"
-    )
+    ) where T <:AbstractDataset
     options = GDAL.gdalbuildvrtoptionsnew(options, C_NULL)
     usage_error = Ref{Cint}()
     result = GDAL.gdalbuildvrt(dest, length(datasets),
         [ds.ptr for ds in datasets], C_NULL, options, usage_error)
     GDAL.gdalbuildvrtoptionsfree(options)
-    return Dataset(result)
+    return T(result)
```
yeesian, line 233: This should still be `Dataset(result)` rather than `T(result)` (see https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084 for context).

yeesian (commented):
This is great, thanks! My comments here are just echoing what @visr has said in https://github.com/yeesian/ArchGDAL.jl/pull/167#issuecomment-765704084, in case having it might help.

---

If you want corresponding functions that returns the Interactive versions of the corresponding datasets: you can still do so, but it has to be a separate implementation, e.g. for `gdalwarp`, adding the following lines:

```
function gdalwarp(
        datasets::Vector{T},
        options = String[];
        dest = "/vsimem/tmp"
    ) where T<:AbstractDataset
    options = GDAL.gdalwarpappoptionsnew(options, C_NULL)
    usage_error = Ref{Cint}()
    result = GDAL.gdalwarp(dest, C_NULL,
        length(datasets), [ds.ptr for ds in datasets], options, usage_error)
    GDAL.gdalwarpappoptionsfree(options)
    return IDataset(result)
```

If an example might help, see e.g. how it's done for [fromWKB](https://github.com/yeesian/ArchGDAL.jl/blob/40c7a2b7aea45c8116189c381626e2117f574516/src/ogr/geometry.jl#L11-L33).

yeesian:
This has been incorporated in #179, thank you for the contribution!

PR #172, opened 2021-02-18
build docs on latest Documenter

visr:
mainly to get that CI button green again with the new GDAL and PROJ builds


yeesian:
Thanks! :)

PR #174, opened 2021-02-22
add methods to get driver shortname from filename extension

### LINKED ISSUE #173: Guess output file driver
Currently we have `identifydriver`, but this is for identifying an existing file. We could also guess the driver based on the file extension, for a file we intend to write.Similar functionality been merged for rasterio, see:

https://github.com/mapbox/rasterio/pull/2008
https://github.com/Toblerity/Fiona/pull/948/files

Any thoughts? I think it's better to have this here than in GeoData.jl, I'll put together a PR if people are interested.

Edit (although the rasterio license seems pretty vague for actually reading/copying these):

In rasterio, most of the action is here - where all drivers have their extensions collated from GDAL:
https://github.com/mapbox/rasterio/blob/99511e4069feec4db91374243b7edc382b620b9d/rasterio/_base.pyx#L120

We could have a similar method, but just `driver_extensions`, returning a `Dict` of extensions and their drivers.

They then have `driver_from_extension` function:
https://github.com/mapbox/rasterio/blob/d7b2dd3ae64c55978e265fa9230732e88b1dc9ae/rasterio/drivers.py#L37

With some fallbacks for < GDAL 2.0. Which we probably don't have to worry about.

rafaqz:
I'm not sure if these methods are the _best_ way to organise this kind of functionality, or the best names. But it's useful for writing arbitrary drivers without specifying them.

Closes #173

### DIFF
--- a/src/driver.jl
+++ b/src/driver.jl
@@ -147,14 +147,12 @@ with their respective drivers shortname.
 """
 function extensions()
     extdict = Dict{String,String}()
-    for i in 1:GDAL.gdalgetdrivercount()
-        driver = ArchGDAL.getdriver(i)
+    for i in 1:ndriver()
+        driver = getdriver(i)
         if !(driver.ptr == C_NULL)
-            exts = GDAL.gdalgetmetadataitem(driver.ptr, "DMD_EXTENSIONS", Cstring(C_NULL))
-            if !(exts isa Nothing)
-                for ext in split(exts)
-                    extdict[".$ext"] = ArchGDAL.shortname(driver)
-                end
+            # exts is a space-delimited list in a String, so split it
+            for ext in split(metadataitem(driver, "DMD_EXTENSIONS"))
+                extdict[".$ext"] = shortname(driver)
             end
         end
     end
@@ -170,9 +168,8 @@ function extensiondriver(filename::AbstractString)
     split = splitext(filename)
     extensiondict = extensions()
     ext = split[2] == "" ? split[1] : split[2]
-    if haskey(extensiondict, ext)
-        extensiondict[ext]
-    else
+    if !haskey(extensiondict, ext)
         throw(ArgumentError("There are no GDAL drivers for the $ext extension"))
     end
+    return extensiondict[ext]
 end
--- a/src/driver.jl
+++ b/src/driver.jl
@@ -138,3 +138,41 @@ end
 
 copyfiles(drvname::AbstractString, new::AbstractString, old::AbstractString) =
     copyfiles(getdriver(drvname), new, old)
+
+"""
+    extensions()
+
+Returns a `Dict{String,String}` of all of the file extensions that can be read by GDAL, 
+with their respective drivers shortname.
+"""
+function extensions()
+    extdict = Dict{String,String}()
+    for i in 1:GDAL.gdalgetdrivercount()
+        driver = ArchGDAL.getdriver(i)
+        if !(driver.ptr == C_NULL)
+            exts = GDAL.gdalgetmetadataitem(driver.ptr, "DMD_EXTENSIONS", Cstring(C_NULL))
+            if !(exts isa Nothing)
+                for ext in split(exts)
+                    extdict[".$ext"] = ArchGDAL.shortname(driver)
+                end
+            end
+        end
+    end
+    return extdict
+end
+
+"""
+    extensiondriver(filename::AbstractString)
+
+Returns a `Driver` that matches the filename extension.
+"""
+function extensiondriver(filename::AbstractString)
+    split = splitext(filename)
+    extensiondict = extensions()
+    ext = split[2] == "" ? split[1] : split[2]
+    if haskey(extensiondict, ext)
+        extensiondict[ext]
+    else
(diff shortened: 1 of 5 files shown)


src/driver.jl
```diff
@@ -138,3 +138,41 @@ end
...
+    return extdict
+end
+
+"""
+    extensiondriver(filename::AbstractString)
+
+Returns a `Driver` that matches the filename extension.
+"""
+function extensiondriver(filename::AbstractString)
+    split = splitext(filename)
```
yeesian, line 170: should we check that `split` has length 2?

src/driver.jl
```diff
@@ -138,3 +138,41 @@ end
...
+    extensiondict = extensions()
+    ext = split[2] == "" ? split[1] : split[2]
+    if haskey(extensiondict, ext)
+        extensiondict[ext]
+    else
+        throw(ArgumentError("There are no GDAL drivers for the $ext extension"))
+    end
+end
```
yeesian, line 178: Can we rewrite this as

```julia
    if !haskey(extensiondict, ext)
        throw(ArgumentError("There are no GDAL drivers for the $ext extension"))
    end
    return extensiondict[ext]
end
```
?

That way it becomes clearer what is being returned, and what's being checked as a potential error.

test/test_drivers.jl
```diff
@@ -72,3 +72,10 @@ end
         )
     end
 end
+
+@testset "Test Driver from Extension" begin
```
yeesian, line 76: Might it be possible to add tests for the error path? It'll be good to check that the error message will be rendered as intended.

src/driver.jl
```diff
@@ -138,3 +138,41 @@ end
...
+with their respective drivers shortname.
+"""
+function extensions()
+    extdict = Dict{String,String}()
+    for i in 1:GDAL.gdalgetdrivercount()
+        driver = ArchGDAL.getdriver(i)
+        if !(driver.ptr == C_NULL)
+            exts = GDAL.gdalgetmetadataitem(driver.ptr, "DMD_EXTENSIONS", Cstring(C_NULL))
+            if !(exts isa Nothing)
+                for ext in split(exts)
```
yeesian, line 155: It's not immediately obvious what `split(exts)` does (since we don't specify a delimiter here). Can we add a clarifying comment for it?

yeesian (commented):
Thanks, I like the approach taken here!

src/driver.jl
```diff
@@ -138,3 +138,41 @@ end
...
+Returns a `Dict{String,String}` of all of the file extensions that can be read by GDAL, 
+with their respective drivers shortname.
+"""
+function extensions()
+    extdict = Dict{String,String}()
+    for i in 1:GDAL.gdalgetdrivercount()
+        driver = ArchGDAL.getdriver(i)
+        if !(driver.ptr == C_NULL)
+            exts = GDAL.gdalgetmetadataitem(driver.ptr, "DMD_EXTENSIONS", Cstring(C_NULL))
+            if !(exts isa Nothing)
```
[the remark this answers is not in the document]
rafaqz: Ok nice to wtap it properly

src/driver.jl
rafaqz: I think it always has length 2

src/driver.jl
rafaqz: Yep. They come as a space delimited list from GDAL

rafaqz:
Currently I'm actually returning the driver `shortname`. Would it be better to return the actual Driver?

rafaqz:
Those test failures seem to be unrelated

src/utils.jl
```diff
@@ -70,6 +70,23 @@ Fetch metadata. Note that relatively few formats return any metadata.
...
+Fetch single metadata item.
+
+### Parameters
+* `name` the name of the metadata item to fetch.
+* `domain` (optional) the domain to fetch for.
+
+### Returns
+The metadata item on success, or an empty string on failure.
+"""
+function metadataitem(obj, name::AbstractString; domain::AbstractString = "")
```
yeesian, line 85: Can we introduce unit tests for this?

src/driver.jl
```diff
@@ -138,3 +138,40 @@ end
...
+end
+
+"""
+    extensiondriver(filename::AbstractString)
+ 
+Returns a driver shortname that matches the filename extension.
+
+So `extensiondriver("/my/file.tif") == "GTiff"`.
+"""
+function extensiondriver(filename::AbstractString)
```
yeesian, line 169: Can we introduce unit tests for this?

yeesian (approved):
Thanks, it looks good now! Can we introduce unit tests for the functions before I merge this PR?

Regarding https://github.com/yeesian/ArchGDAL.jl/pull/174#issuecomment-783956707, I like the shortname and don't know of any reason offhand to go with the longer driver name for now.

And yeah the test failures looks unrelated so I won't worry about them in the setting of this PR.

src/utils.jl
rafaqz: where do you want to put this test

rafaqz:
This is ready to go.

yeesian:
Thank you! :)

Issue #178, opened 2021-04-18
Better errors for missing Dict keys


rafaqz:
It's possible to get errors like this when you try to create a file with the wrong data type, and the `_GDALTYPE` Dict is accessed:

```julia
ERROR: KeyError: key Int64 not found
Stacktrace:
 [1] getindex at ./dict.jl:467 [inlined]
 [2] #unsafe_create#56 at /root/.julia/packages/ArchGDAL/j9NPL/src/dataset.jl:200 [inlined]
 [3] create(::GeoData.var"#190#191"{GeoArray{Int64,3,Tuple{Lon{LinRange{Float64},Projected{
 ```

It could be better to throw a more descriptive error, like "`Int64` is not a recognised data type" etc. I'm not sure if this also applies to the other const Dicts in use like `_JLType`

See https://github.com/rafaqz/GeoData.jl/issues/143

evetion:
I have similar issues over at https://github.com/evetion/GeoArrays.jl/issues/42, would be nice if we can solve it within ArchGDAL.

yeesian:
Addressed by #179.

Issue #42, opened 2018-05-08
Towards better color support


yeesian:
gathering links for now;
- http://osgeo-org.1560.x6.nabble.com/GTiff-gdal-translate-CMYK-problem-td3758190.html
- https://github.com/wkearn/RasterIO.jl/blob/master/doc/devdocs/DataModel.md
- https://lists.osgeo.org/pipermail/gdal-dev/2014-October/040318.html

PR #179, opened 2021-04-19
[Breaking] Maintain a closed environment of GDAL-compatible types

### LINKED ISSUE #196: Documentation of geometry column names in Tables
Copied over from https://github.com/yeesian/ArchGDAL.jl/pull/160#issuecomment-843093698 below.

> Right now we only rename "" to "geometry". Turns out the default GEOMETRY_NAME for GeoPackage is "geom". Not that that is standard though, PostGIS has:
>
> GEOMETRY_NAME: Set name of geometry column in new table. If omitted it defaults to wkb_geometry for > GEOM_TYPE=geometry, or the_geog for GEOM_TYPE=geography.
> 
> And MySQL for instance:
> 
> The geometry column, which defaults to SHAPE and can be overridden with the GEOMETRY_NAME layer creation option
> 
> https://gdal.org/drivers/vector/gpkg.html
> https://gdal.org/drivers/vector/pg.html
> https://gdal.org/drivers/vector/mysql.html

It should be turned into documentation per https://github.com/yeesian/ArchGDAL.jl/pull/160#issuecomment-844667327.

yeesian:
Fixes #196, #192, #178, #167, #155, #154, #83, #42

As it wasn't obvious what's being returned from functions, it's difficult to track what's still "leaking" GDAL data types. Therefore, there is a strong emphasis in this PR on (i) annotating all functions with return types, and (ii) getting test coverage close to 100%. In the process, I caught quite a few non-obvious bugs, and am instituting guides and processes to make it easier to catch them in the future.

# For reviewers

I am sorry about making pull requests that are this big in scope, and mixing between multiple different types of changes. I encourage reading this comment in full, before proceeding to [review the files changed](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/reviewing-proposed-changes-in-a-pull-request#reviewing-dependency-changes). If it helps to know: 
* I really enjoy having your input so, if you'd like, take the time you need to review it. Due to the scope of the changes, I'll also understand if you prefer to opt out of reviewing this PR.
* Most of the changes are stylistic and can be [reviewed locally](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/reviewing-proposed-changes-in-a-pull-request#marking-a-file-as-viewed), so hopefully it might help with the process.
* I have put in a lot more effort to improve on the test coverage in this PR (rather than relying on reviewers to reason about correctness), so hopefully your time is more productively spent reviewing the changes described in this comment.
* The whitespacing changes in the unit tests from JuliaFormatter have made it almost impossible to review them in any reasonable change, so I'd suggest just ignoring them and reviewing the rest of the changes. 

# Changelog

## Breaking

The biggest ones are:

* `_FIELDTYPE[fieldtype]` -> `convert(DataType, fieldtype)`
* `_GDALTYPE[dtype]` -> `convert(GDALDataType, dtype)`

High level summary:
* Introduced ArchGDAL enums that mirror the CEnums in GDAL. Although It isn't a hard requirement, I have opted to preserve the names for the most part to (i) aid with future searches on the web and (ii) make it easier to update the code for future versions of GDAL.
* Aligned the Tables interface with https://tables.juliadata.org/stable/#Tables.AbstractRow-1. See the [delta in documentation](https://github.com/yeesian/ArchGDAL.jl/pull/179/files?short_path=00416cd#diff-00416cd9a5880dbaedb3b0d9d982e1e6b8a560e03c228cc5c52a92fd1c01fb33).
* Implemented `pixeltype(dataset)` to be based on the widest `pixeltype()` of all its bands. (Previously, it assumed that all bands have the same pixeltype, and just used the pixeltype of the first band in all the `rasterio!()` methods.)
* Use `!` to annotate mutating functions.
* Delete functions that will not be supported (either because there are better approaches in Julialang, or they are untested and should have corresponding feature requests to be added.) E.g. rather than encouraging operations on colortables (with very limited functionality from GDAL), users are better served by arrays of ColorTypes using Colors.jl, [e.g.](http://juliagraphics.github.io/Colors.jl/stable/colormapsandcolorscales/#Generating-a-range-of-colors)

```
range(c1, stop=c2, length=15)`
```

instead of

```
createcolorramp!(colortable, startindex, startcolor, endindex, endcolor)
```

Detailed list of changes:
* `_FIELDTYPE[fieldtype]` -> `convert(DataType, fieldtype)`
* `_GDALTYPE[dtype]` -> `convert(GDALDataType, dtype)`
* `setfield!(feature::Feature, i::Integer, value::Vector{GDAL.GIntBig})` -> `setfield!(feature::Feature, i::Integer, value::Vector{Int64})`
* `setfield!(feature::Feature, i::Integer, value::Vector{GDAL.GByte})` -> `setfield!(feature::Feature, i::Integer, value::Vector{UInt8})`
* `validate(feature::Feature, flags::Integer, emiterror::Bool)` -> `validate(feature::Feature, flags::FieldValidation, emiterror::Bool)`
* `copywholeraster(...)` -> `copywholeraster!(...)`
* `_dataset_type(dataset)` -> `pixeltype(dataset)`
* GDAL.GDT_* -> GDT_*
* GDAL.OFT* -> OFT*
* GDAL.OFST* -> OFST*
* GDAL.OJ* -> OJ*
* GDAL.GFT_* -> GFT_*
* GDAL.GFU_* -> GFU_*
* GDAL.GA_* -> GA_*
* GDAL.GF_* -> GF_*
* GDAL.GPI_* -> GPI_*
* GDAL.GCI_* -> GCI_*
* GDAL.GARIO_* -> GARIO_*
* GDAL.OGRSTC* -> OGRSTC*
* GDAL.OGRSTU* -> OGRSTU*
* GDAL.wkb* -> wkb*
* GDAL.OGR_F_VAL_* -> F_VAL_*
* `OF_ReadOnly` -> `OF_READONLY`
* `OF_Update` -> `OF_UPDATE`
* `OF_Raster` -> `OF_RASTER`
* `OF_Vector` -> `OF_VECTOR`
* `OF_Verbose_Error` -> `OF_VERBOSE_ERROR`
* Remove `initialize!(stylemanager::StyleManager, feature::Feature)`
* Remove `ncolorentry(ct::ColorTable)`
* Remove `getcolorentry(ct::ColorTable, i::Integer)`
* Remove `getcolorentryasrgb(ct::ColorTable, i::Integer)`
* Remove `setcolorentry!(ct::ColorTable, i::Integer, entry::GDAL.GDALColorEntry)`
* Remove `createcolorramp!(...)`
* Remove `serializeJSON(rat::RasterAttrTable)` (It hasn't been tested: you can fallback on `GDAL.gdalratserializejson(rat.ptr)` and open it as a feature request.)
* Remove `setcategorynames!(band::AbstractRasterBand, names)` (fallback: `GDAL.gdalsetrastercategorynames(band.ptr, names)`)
* Remove `getlayer(t::Table)` (use `Tables.rows(t)` instead)
* Remove `Base.getindex(t::Table, idx::Integer)` (Only in rare circumstances is SetNextByIndex() efficiently implemented, and we're not willing to provide the expectation that it is. Rather than treating the table like an array, users should materialize the data they need into one and go from there.)
* Remove `Base.IteratorSize(::Type{<:Table})`
* Remove `Base.size(t::Table)` and `Base.length(t::Table)`
* Remove `Base.IteratorEltype(::Type{<:Table})`
* Remove `Base.propertynames(t::Table)`
* Remove `Base.getproperty(t::Table, s::Symbol)`
* Remove `nextnamedtuple(layer::IFeatureLayer)`
* Remove `schema_names(layer::AbstractFeatureLayer)` (use `schema_names(featuredefn)` instead)
* Remove `_JLTYPE`
* Remove `_GDALTYPE` (used in https://github.com/evetion/GeoArrays.jl/blob/6d7fb9bd05c3a2a2ca58129cbd5427a4bf491b92/src/io.jl#L65-L77)
* Remove `_FIELDTYPE` (used in https://github.com/sigmafelix/jugs/blob/master/jugs_base.jl#L73, https://github.com/evetion/GeoDataFrames.jl/blob/4ca6e1159f401e9b70b15acb3655a1a84a50d4de/src/io.jl#L7, https://github.com/cossatot/Oiler/blob/795ccbbc89b67f2b919071e2c04ff54df2d46633/src/io.jl#L153)

Most users shouldn't be affected by the following changes:
* Remove `GDALColorTable` (fallback is to use `GDAL.GDALColorTableH`)
* Remove `GDALCoordTransform` (fallback is to use `GDAL.OGRCoordinateTransformationH`)
* Remove `GDALDataset` (fallback is to use `GDAL.GDALDatasetH`)
* Remove `GDALDriver` (fallback is to use `GDAL.GDALDriverH`)
* Remove `GDALFeature` (fallback is to use `GDAL.OGRFeatureH`)
* Remove `GDALFeatureDefn` (fallback is to use `GDAL.OGRFeatureDefnH`)
* Remove `GDALFeatureLayer` (fallback is to use `GDAL.OGRLayerH`)
* Remove `GDALField` (fallback is to use `GDAL.OGRField`)
* Remove `GDALFieldDefn` (fallback is to use `GDAL.OGRFieldDefnH`)
* Remove `GDALGeometry` (fallback is to use `GDAL.OGRGeometryH`)
* Remove `GDALGeomFieldDefn` (fallback is to use `GDAL.OGRGeomFieldDefnH`)
* Remove `GDALProgressFunc` (fallback is to use `GDAL.GDALProgressFunc`)
* Remove `GDALRasterAttrTable` (fallback is to use `GDAL.GDALRasterAttributeTableH`)
* Remove `GDALRasterBand` (fallback is to use `GDAL.GDALRasterBandH`)
* Remove `GDALSpatialRef` (fallback is to use `GDAL.OGRSpatialReferenceH`)
* Remove `GDALStyleManager` (fallback is to use `GDAL.OGRStyleMgrH`)
* Remove `GDALStyleTable` (fallback is to use `GDAL.OGRStyleTableH`)
* Remove `GDALStyleTool` (fallback is to use `GDAL.OGRStyleToolH`)
* Remove `StringList` (fallback is to use `Ptr{Cstring}`)
* Remove `destroy(drv::Driver)`
* `destroy(...)` returns `nothing` (this should have been the default behavior all along; see e.g. https://github.com/niclasmattsson/GlobalEnergyGIS/blob/2e6f87df8ad6a44de25f2a207dbabe7c0f41eca7/src/rasterize_shapefiles.jl#L99-L100)
* `RasterDataset{T,DS}` -> `RasterDataset{T,DS<:AbstractDataset}`
* `GDALRasterBand` -> `GDAL.GDALRasterBandH`
* `Base.show(io::IO, object)` now returns `nothing` (although it shouldn't have been depended on, nor should it be depended on)
* `BlockIterator -> BlockIterator{T<:Integer}` (`T` being the data type of the iterator index)
* `WindowIterator -> WindowIterator{T<:Integer}` (`T` being the data type of the iterator index)
* `BufferIterator{<: Real} -> BufferIterator{<: Real, <:Integer}`

## Stylistic

* Enforcement of a style guide using https://github.com/domluna/JuliaFormatter.jl with the settings in https://github.com/yeesian/ArchGDAL.jl/blob/08505ee46927281bbbada1d3b474892d5c194b7c/.JuliaFormatter.toml.
* Annotate most/all functions with return types (breaking changes described above) and make the return statement explicit:
    * The `return` keyword should be at the top level if it is the last statement in the function.
    * All multiline functions should have a return statement.
    * Err on the side of returning values defined in base Julia or ArchGDAL, rather than returning the output of a function defined in a different package. This will protect the return signature in the function if there are breaking changes from upstream.
    * `destroy(...)` will return `nothing` (breaking changes described above).
* Implemented a `gdalenum` macro to ease with the wrapping of GDAL enum types (of which there are plenty).

## Organizational

* Added a [README for building the documentation](https://github.com/yeesian/ArchGDAL.jl/blob/08505ee46927281bbbada1d3b474892d5c194b7c/docs/README.md).
* Added a [README for running the tests and documenting practices for working with data](https://github.com/yeesian/ArchGDAL.jl/blob/405fb7fe4ae433a3ee3ce01fde0728da31b291b3/test/README.md).
* Provide an example for [managing dependencies](https://github.com/yeesian/ArchGDAL.jl/blob/08505ee46927281bbbada1d3b474892d5c194b7c/README.md#dependencies) (because it was non-obvious to me).
* Add links to Julialang, OSGeo and JuliaGeo, and articulate my [preference of support for them](https://github.com/yeesian/ArchGDAL.jl/blob/08505ee46927281bbbada1d3b474892d5c194b7c/README.md#community).

# Remaining TODOs
* [x] update documentation
* [x] Look up known users of ArchGDAL (I did the following searches: [repo](https://github.com/search?q=ArchGDAL+language%3AJulia&type=Repositories&ref=advsearch&l=Julia&l=), [code](https://github.com/search?l=&q=ArchGDAL+language%3AJulia&type=code)); help them to migrate where possible
    * [x] https://github.com/rafaqz/GeoData.jl
    * [x] https://github.com/EcoJulia/SimpleSDMLayers.jl
    * [x] https://github.com/meggart/YAXArrayBase.jl
    * [x] https://github.com/Circuitscape/Circuitscape.jl
    * [x] https://github.com/Circuitscape/Omniscape.jl
    * [x] https://github.com/niclasmattsson/GlobalEnergyGIS
    * [x] https://github.com/uoa-ems-research/JEMSS.jl
    * [x] https://github.com/CliMA/GriddingMachine.jl
    * [x] https://github.com/PoisotLab/ms_frontiers_kids
    * [x] https://github.com/caiohamamura/jl_jingjing
    * [x] https://github.com/ashiklom/sbg-prosail-workflow/tree/master/julia
    * [x] https://github.com/sfacca/PrismaConvert.jl
    * [x] https://github.com/zhangliye/x2osm
    * [x] https://github.com/csp-inc/VermontTerrestrialPassageTool.jl
    * [x] https://github.com/Circuitscape/SpatialGraphs.jl
    * [x] https://github.com/mauriciogtec/power-plants
    * [x] https://github.com/cesaraustralia/FallArmyWorm
    * [x] https://github.com/cossatot/china_active_faults
    * [ ] https://github.com/sfacca/stage-Machine-learning
    * [ ] https://github.com/brianhempel/nadocast
    * [ ] https://github.com/boydorr/EcoSISTEM.jl
    * [x] https://github.com/cossatot/Oiler
      * [ ] https://github.com/cossatot/Oiler/blob/795ccbbc89b67f2b919071e2c04ff54df2d46633/src/io.jl#L153
    * [x] https://github.com/evetion/SpaceLiDAR.jl
      * [ ] https://github.com/evetion/SpaceLiDAR.jl/blob/bef00ad2340d7341bcc50a5cea5bb7e324df4f44/scripts/outliers.jl#L100-L101
    * [x] https://github.com/sigmafelix/jugs
      * [ ] https://github.com/sigmafelix/jugs/blob/6e1ecb28a7f401fe9bdb02ddadb4e54e73c86c4d/jugs_base.jl#L73
    * [x] https://github.com/adolgert/simplexpop.jl
      * [ ] https://github.com/adolgert/simplexpop.jl/blob/4e403816f7b94002b8dff45db859ab45ecd8849b/src/applydistricts.jl#L312
    * [x] https://github.com/caiohamamura/jl_jingjing
      * [ ] https://github.com/caiohamamura/jl_jingjing/blob/798550aee2a8a61162be29c6313782a75b69f853/old/01reverse_datasets.jl#L11
      * [ ] https://github.com/caiohamamura/jl_jingjing/blob/798550aee2a8a61162be29c6313782a75b69f853/old/02_compute_mean_sd.jl#L35
    * [x] https://github.com/JuliaGeo/GADM.jl
      * [ ] https://github.com/JuliaGeo/GADM.jl/blob/0bce612d3d38b585cac5ad4ea291af6d13e37ed9/src/GADM.jl#L110-L116
    * [x] https://github.com/evetion/GeoArrays.jl
      * [ ] https://github.com/evetion/GeoArrays.jl/blob/6d7fb9bd05c3a2a2ca58129cbd5427a4bf491b92/src/io.jl#L65-L77
    * [x] https://github.com/evetion/GeoDataFrames.jl
      * [ ] https://github.com/evetion/GeoDataFrames.jl/blob/4ca6e1159f401e9b70b15acb3655a1a84a50d4de/src/io.jl#L62
      * [ ] https://github.com/evetion/GeoDataFrames.jl/blob/4ca6e1159f401e9b70b15acb3655a1a84a50d4de/src/io.jl#L7


README.md
```diff
@@ -29,3 +29,44 @@ To test if it is installed correctly,
...
+
+```julia
+shell> cd /Users/yeesian/.julia/dev/ArchGDAL
+/Users/yeesian/.julia/dev/ArchGDAL
+```
+
+2. Activate the environment corresponding to `Project.toml`):
+
+```julia
+(@v1.6) pkg> activate
```
felixcremer, line 46: Are you missing a . here?
Otherwise this recommendation would install all dependencies into the global environment and this is not what we want here.

yeesian: Haha yes, thank you! :)

src/ogr/feature.jl
```diff
@@ -351,49 +376,55 @@ end
...
-getfield(feature::Feature, name::Union{AbstractString, Symbol}) =
-    getfield(feature, findfieldindex(feature, name))
+function getfield(feature::Feature, name::Union{AbstractString,Symbol})
+    i = findfieldindex(feature, name)
+    return if i == -1
+        nothing
+    else
+        getfield(feature, i)
+    end
+end
```
yeesian, line 416: @evetion (and others), just wanted to call out this snippet, and get your thoughts on returning `nothing` (versus e.g. `missing`) when `i == -1`. (See #177 for context.)

src/raster/rasterband.jl
```diff
@@ -165,23 +177,23 @@ end
...
-used to mark pixels that are not valid data. Such pixels should generally
+returned instead. The no data value for a band is generally a special marker
+value used to mark pixels that are not valid data. Such pixels should generally
 not be displayed, nor contribute to analysis operations.
 
 ### Returns
 the nodata value for this band or `nothing`.
 """
-function getnodatavalue(band::AbstractRasterBand)
+function getnodatavalue(band::AbstractRasterBand)::Union{Cint,Nothing}
```
rafaqz, line 187: This return value doesn't handle floating point nodatavalue `:Union{Cint,Nothing}`

rafaqz:
I like the direction here, and I'll start using this as my working branch to iron out any problems.

Besides the return value on `getnodatavalue` introducing a bug, this doesn't seem to break anything in GeoData.jl.

src/raster/rasterband.jl
yeesian: Thanks for the catch! Looks like I mistook `nodatavalue` for `hasnodatavalue` 😅  How did you notice it?

I don't yet know of a good way for me to audit whether the return types correspond to their current behavior, and am curious if others have a better way of doing it.

visr:
Great work here @yeesian. I'll have a closer look soon. Based on the information in the top post (thanks!) I don't have any objections.

yeesian:
> I like the direction here, and I'll start using this as my working branch to iron out any problems.

Thanks! It isn't very apparent from the top comment, but I've also added support for colors in this PR ([commit](https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/3d025fedb61596ae500e63a652173d175a5abdc1), [preview](https://htmlpreview.github.io/?https://gist.githubusercontent.com/yeesian/69b147b45ff13815fb4c2546fb1be01a/raw/b65979152b78c3c849395fa623a91ea03117e069/images.html)).

I'd be interested to hear more about how I can improve it to be more useful for https://github.com/rafaqz/GeoData.jl and https://github.com/evetion/GeoArrays.jl

src/raster/rasterband.jl
rafaqz: I just ran the GeoData.jl test suite with this branch. That's the only fail :)

Whats the reasoning behind fixing the return type? is it mostly documentation/clarity?

rafaqz:
I noticed the color handling, probably useful! My only concerns with that is if package load time increases, and how we differentiate 2d arrays of colors from 3d arrays with Band as one dimension?

src/raster/rasterband.jl
yeesian: > Whats the reasoning behind fixing the return type? is it mostly documentation/clarity?

It seems like a bug to be coercing `Float64`s into `Cint`s.

rafaqz: Oh sorry I mean generally everywhere in the package you are setting a fixed return type where there wasn't before, I'm wondering what the motivation is for that

yeesian: > Oh sorry I mean generally everywhere in the package you are setting a fixed return type where there wasn't before, I'm wondering what the motivation is for that

Ohh, yeah, it's mostly documentation/clarity. It was also helpful for me to be able to do a search on `::GDAL.` in this repository to find all instances where I have GDAL types in the input or output.

yeesian: I'm also hoping that, in cases like #171, it'll enable the kind of analysis that allows for us to verify that they are fixed (e.g. if **all** functions have return type annotations, and none of the functions will return `::Dataset` or `::RasterBand` or ...)

rafaqz: I'm not sure it can be perfect as there is always the chance of them returning a dataset wrapped in something else, like an array wrapper object or a tuple. I've just been handling this in GeoData.jl as well, for netcdf/hdf5 etc.

yeesian:
> My only concerns with that is if package load time increases [...]

That's a good call-out which prompted #193:
  * It can conceivably be moved to a different package, but people might like to be able to have it natively without having to introduce too many wrapper types. (We can also do the traits-approach for it, but I think the broader question of whether Images should be supported in this package will still remain, so on to the next point --)
  * If package-loading is the main concern, we can use language-level solutions such as https://github.com/timholy/SnoopCompile.jl or https://github.com/JuliaLang/PackageCompiler.jl for reducing them (subject to benchmarking), so I haven't considered it a fundamental blocker
  * I haven't yet heard disagreements about Images.jl or ColorTypes.jl being a controversial choice of representation for images/colors in Julia (and they are a much better alternative to GDAL's limited support ColorTables), so I'm principally going by that reasoning for justifying their inclusion in this package.

> My only concerns with that is [...] how we differentiate 2d arrays of colors from 3d arrays with band as the third dimension?

I ran into this, and my current thinking is
  * For 2D arrays (a single rasterband), if they have a color interpretation, we should use the color interpretation. If they don't have a color interpretation, we should default to "Grey".
  * For >2D arrays (multiple rasterbands), if they have a palette interpretation (or combination of color interpretations) that resolves to a valid colortype, we should use the palette interpretation. If they don't have a palette interpretation, we should throw an error.

Happy to have a discussion and change my mind if people have other opinions on the matter.

rafaqz:
I think Images.jl and ColotTypes.jl are the undisputed standard for doing this, that's probably not a problem. 

Maybe I misunderstand, but is the idea that ArchGDAL will always return data as `Array{<:Color}` - most of the time `Grey`, or that there would be an option for that if you want it in some specific cases? Do you mean loading a regular tiff will return `Array{<:Grey}` ?

yeesian:
I think `read()` will return `Array{UInt8}`, and `imread()` will return `Array{<:Colorant}`

rafaqz:
Oh right that makes sense! That should useful.

yeesian:
The broken documentation ([logs](https://github.com/yeesian/ArchGDAL.jl/pull/179/checks?check_run_id=2636190040)) points to a regression in this PR when [inferring the geometry type of Table columns](https://github.com/yeesian/ArchGDAL.jl/pull/160). To reproduce it: the following code works on master, but fails on this branch.

```julia
julia> using ArchGDAL, DataFrames

shell> cd /Users/yeesian/.julia/dev/ArchGDAL/test
/Users/yeesian/.julia/dev/ArchGDAL/test

julia> dataset1 = ArchGDAL.read("data/multi_geom.csv", options = ["GEOM_POSSIBLE_NAMES=point,linestring", "KEEP_GEOM_COLUMNS=NO"])
GDAL Dataset (Driver: CSV/Comma Separated Value (.csv))
File(s):
  data/multi_geom.csv

Number of feature layers: 1
  Layer 0: multi_geom (wkbUnknown)


julia> layer = ArchGDAL.getlayer(dataset1, 0)
Layer: multi_geom
  Geometry 0 (point): [wkbUnknown]
  Geometry 1 (linestring): [wkbUnknown]
     Field 0 (id): [OFTString], 5.1, 5.2
     Field 1 (zoom): [OFTString], 1.0, 2.0
     Field 2 (location): [OFTString], Mumbai, New Delhi


julia> table = ArchGDAL.Table(layer)
Table with 2 features


julia> df = DataFrame(table)
2×5 DataFrame
 Row │ point               linestring               id      zoom    location
     │ IGeometr…           IGeometry…               String  String  String
   1 │ Geometry: wkbPoint  Geometry: wkbLineString  5.1     1.0     Mumbai
   2 │ Geometry: wkbPoint  Geometry: wkbLineString  5.2     2.0     New Delhi

julia> df.point
2-element Vector{ArchGDAL.IGeometry{GDAL.wkbPoint}}:
 Geometry: POINT (30 10)
 Geometry: POINT (35 15)
```

Project.toml
```diff
@@ -6,11 +6,13 @@ desc = "A high level API for GDAL - Geospatial Data Abstraction Library"
 version = "0.6.0"
 
 [deps]
+ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 DiskArrays = "3c3547ce-8d99-4f5e-a174-61eb10b00ae3"
 GDAL = "add2ef01-049f-52c4-9ee2-e494f65e021a"
 GeoFormatTypes = "68eda718-8dee-11e9-39e7-89f7f65f511f"
 GeoInterface = "cf35fbd7-0cd7-5166-be24-54bfbe79505f"
+ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
```
felixcremer, line 15: This needs a compat entry.

README.md
```diff
@@ -29,3 +30,64 @@ To test if it is installed correctly,
...
+(ArchGDAL) pkg> add CEnum
+   Resolving package versions...
+    Updating `~/.julia/dev/ArchGDAL/Project.toml`
+  [fa961155] + CEnum v0.4.1
+  [3c3547ce] + DiskArrays v0.2.7
+  [add2ef01] + GDAL v1.2.1
+  [68eda718] + GeoFormatTypes v0.3.0
+  [cf35fbd7] + GeoInterface v0.5.5
+  [bd369af6] + Tables v1.4.2
+```
```
felixcremer, line 93: We should add  a section about the compat entries. 
These are needed, so that Julia can resolve the package versions and it is good practice to have compate entries for all dependencies.

```suggestion
```4. Add a compat entry into Project.toml so that julia can resolve the versions
```     Open the Project.toml file in a texteditor and add in the [compat] section the line:
        CEnum = "0.4"

```

yeesian: Thank you!

Project.toml
yeesian: Thanks! Added in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/fcaba40d908d2778799fb3643580196f88805365. Feel free to ping me if there are any issues with it.

src/ogr/feature.jl
visr: I think `missing` would be a better choice here. In the julia docs there is [a FAQ entry](https://docs.julialang.org/en/v1/manual/faq/#Nothingness-and-missing-values-1), and it states:

> To represent missing data in the statistical sense (NA in R or NULL in SQL), use the missing object.

The NULL in the example of #177 seems to represent missing data to me. Especially if we see Features as rows in a Table / records in a dataset.

`findfieldindex`, if it was written in Julia, should probably return `nothing` instead of `-1`, like `findfirst`. But the index truly doesn't exist, whereas the data can be viewed as missing (the field name exists after all).

test/test_cookbook_geometry.jl
```diff
@@ -1,770 +1,981 @@
...
+                0x41,
+                0xc5,
+                0x75,
+                0x87,
+                0xc4,
+                0x9d,
+                0xc0,
+                0x25,
+                0x41,
+            ]
```
visr, line 962: This could be worth a https://domluna.github.io/JuliaFormatter.jl/dev/skipping_formatting/

```
#! format: off
<this test>
#! format: on
```

test/test_gdal_tutorials.jl
```diff
@@ -1,130 +1,227 @@
...
+            @test AG.maximum(band) ≈ 255.0
+
+            @test AG.noverview(band) == 0
+
+            # Reading Raster Data
+            @test AG.width(band) == 100
+            data = map(Float32, AG.read(dataset, 1))
+            @test data[:, 1] ≈ Float32[
+                107.0f0,
+                123.0f0,
```
visr, line 36: Also https://domluna.github.io/JuliaFormatter.jl/dev/skipping_formatting/ maybe?

test/test_gdalutilities.jl
```diff
@@ -18,93 +188,64 @@ AG.read("data/utmsmall.tif") do ds_small
...
+    #       GDAL_DATA environment variable to point to the directory
+    #       containing EPSG csv files.
+    # @testset "GDAL Warp" begin
+    #     AG.gdalwarp([ds_small], ["-of","MEM","-t_srs","EPSG:4326"]) do ds_warped
+    #         @test AG.width(ds_small) == 100
+    #         @test AG.height(ds_small) == 100
+    #         @test AG.width(ds_warped) == 109
+    #         @test AG.height(ds_warped) == 91
+    #     end
+    # end
```
visr, line 241: I assume this test can be enabled now?

visr (commented):
Wow this is huge. Awesome work @yeesian! Left a few notes.

.github/workflows/format_pr.yml
```diff
@@ -0,0 +1,30 @@
+name: format-pr
```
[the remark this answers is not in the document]
yeesian: That's fair. I've removed `format_check` in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/bf6be50008dabb127efc03bd7cf3eec33b4bcfba now. That way we don't bother PRs can focus on the code without worrying about the specific version of JuliaFormatter to use.

README.md
```diff
@@ -29,3 +32,72 @@ To test if it is installed correctly,
 ```julia
 pkg> test ArchGDAL
 ```
+
+## Getting Involved
+
+### Community
+
+This package will not be possible without https://julialang.org/, https://github.com/OSGeo/gdal and https://github.com/JuliaGeo/GDAL.jl. They are maintained by https://julialang.org/community/, https://www.osgeo.org/ and https://juliageo.org/. In case of any contention for support and involvement, we encourage participation and contributions to those projects and communities over this package.
```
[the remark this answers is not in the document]
yeesian: Updated in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/2a31683164a9ec97bbb061ce4e26a45236cc8a57. Thanks!

README.md
```diff
@@ -29,3 +32,72 @@ To test if it is installed correctly,
...
+### Style Guide
+
+ArchGDAL.jl uses [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl) as
+an autoformatting tool, and uses the options in [`.JuliaFormatter.toml`](https://github.com/yeesian/ArchGDAL.jl/blob/master/.JuliaFormatter.toml).
+
+To format code, `cd` to the ArchGDAL.jl directory, then run:
+```julia
+] add JuliaFormatter@0.13.10
+using JuliaFormatter
+format("src")
```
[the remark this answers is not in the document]
yeesian: This empty do-block used to cause it to break: https://github.com/yeesian/ArchGDAL.jl/blob/6d67ee60bfdc3cd680c57efa3bc504bbeae088cb/test/test_gdalutilities.jl#L98-L100

So I only restricted it to `src/` (as I couldn't figure out why it was breaking). Your comment made me dig deeper and root-cause it, so thank you! It's fixed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/b98940ba83fe03e48908162f246391f242ac4c4b and https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/a665f3407930b8221269f8949c246db022c3a85c.

README.md
```diff
@@ -29,3 +32,72 @@ To test if it is installed correctly,
...
+```
+
+A continuous integration check verifies that all PRs made to ArchGDAL.jl have passed the formatter.
+
+### Dependencies
+To manage the dependencies of this package, we work with [environments](https://pkgdocs.julialang.org/v1.6/environments/):
```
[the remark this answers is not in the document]
yeesian: Done in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/4f7058c0407804e97b38378ed38a0a5978aaf067, thanks!

docs/README.md
```diff
@@ -0,0 +1,9 @@
+# ArchGDAL.jl documentation
+
+The latest stable version of the documentation can be viewed at https://yeesian.com/ArchGDAL.jl/stable/. To build a local version of the documentation, navigate to this directory and run the following command:
+
+```shell
+$ julia make.jl
```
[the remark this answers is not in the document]
yeesian: Done in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/52ed3917aaf223a71b166bac5a99c386299ab219, thanks!

src/utils.jl
```diff
@@ -82,9 +132,13 @@ Fetch single metadata item.
...
 """
-function metadataitem(obj, name::AbstractString; domain::AbstractString = "") 
+function metadataitem(
+    obj,
+    name::AbstractString;
+    domain::AbstractString = "",
+)::String
     item = GDAL.gdalgetmetadataitem(obj.ptr, name, domain)
-    return item === nothing ? "" : item
+    return isnothing(item) ? "" : item
```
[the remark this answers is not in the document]
yeesian: Reverted in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/90b864632848aeb0cb3942be38ecf163c2e2bc90, thanks!

test/test_cookbook_geometry.jl
yeesian: Reflected in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/e86beac5117657411ed42ff1f269003db0bd6f69, thanks!

test/test_gdal_tutorials.jl
yeesian: Reflected in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/e86beac5117657411ed42ff1f269003db0bd6f69, thanks!

test/test_gdalutilities.jl
yeesian: Done in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/ac50e17d930a19142b9d0b9ba79e2e5b13a936ce, thanks for the catch!

src/tables.jl
```diff
@@ -1,73 +1,74 @@
...
-function Base.iterate(t::Table, st = 0)
-    layer = getlayer(t)
-    st >= nfeature(layer) && return nothing
-    if iszero(st)
-        resetreading!(layer)
-    end
-    return nextnamedtuple(layer), st + 1
+function Tables.rows(t::Table{T})::T where {T<:AbstractFeatureLayer}
+    return t.layer
 end
```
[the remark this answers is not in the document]
yeesian: To my understanding,
* `ArchGDAL.Feature` meets the criteria for an [`AbstractRow`](https://tables.juliadata.org/dev/#Tables.AbstractRow-1) based on https://github.com/yeesian/ArchGDAL.jl/blob/a665f3407930b8221269f8949c246db022c3a85c/src/tables.jl#L31-L58.
* `ArchGDAL.FeatureLayer` meets the criteria for an `AbstractRow`-iterator based on the previous bullet and meeting the criteria for [`Iteration`](https://github.com/yeesian/ArchGDAL.jl/blob/a665f3407930b8221269f8949c246db022c3a85c/src/base/iterators.jl#L1-L18) in https://github.com/yeesian/ArchGDAL.jl/blob/a665f3407930b8221269f8949c246db022c3a85c/src/base/iterators.jl#L1-L18

Actually, I'm not sure we should even have the `struct Table` type as a wrapper around a FeatureLayer (i.e. why not directly implement the Table interface for a FeatureLayer?)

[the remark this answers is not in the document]
yeesian: > Actually, I'm not sure we should even have the `struct Table` type as a wrapper around a FeatureLayer (i.e. why not directly implement the Table interface for a FeatureLayer?)

To follow up on this: the `struct table` type is now removed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/dadfa12e10c6d0719c32e855cbb9415753e1b3a2. Along the way, we [caught a bug](https://github.com/yeesian/ArchGDAL.jl/runs/2706849393) that was introduced in https://github.com/yeesian/ArchGDAL.jl/pull/160 and fixed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/d2ae1bd1bff8a2378e871ca681deb3aed8031122.

@evetion for vis.

src/tables.jl
```diff
@@ -1,73 +1,74 @@
 """
-Constructs `Table` out of `FeatureLayer`, where every row is a `Feature` consisting of Geometry and attributes.
-```
-ArchGDAL.Table(T::Union{IFeatureLayer, FeatureLayer})
-```
+A tabular representation of a `FeatureLayer`
```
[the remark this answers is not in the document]
yeesian: Oh thanks I just saw this message and responded in kind in the other thread.

docs/src/images.md
```diff
@@ -0,0 +1,53 @@
...
+## Reading from Rasterbands
+We can also read from individual raster bands:
+```@example rasters
+AG.imread(AG.getband(dataset, 1))
+```
+Or equivalently,
+```@example rasters
+AG.imread(dataset, 1)
+```
+It will interpret the color channel (for RGB) correctly there is one. E.g.
```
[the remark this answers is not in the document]
yeesian: Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/a8f831e214c074805927506b82ce1af99965211f, thanks!

docs/src/images.md
```diff
@@ -0,0 +1,53 @@
...
+## Working with Colors
+Operations on colors behave as you think they might:
+```@example rasters
+AG.imread(dataset, 2) + AG.imread(dataset, 3)
+```
+and
+```@example rasters
+0.5 * AG.imread(dataset, 1) + AG.imread(dataset, 3)
+```
+See [Colors.jl](http://juliagraphics.github.io/Colors.jl/stable/) for more on what you can do.
```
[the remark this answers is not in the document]
yeesian: Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/a8f831e214c074805927506b82ce1af99965211f, thanks!

src/raster/images.jl
```diff
@@ -0,0 +1,259 @@
+function imview(colortype::Type{<:ColorTypes.Colorant}, imgvalues...)
+    return ImageCore.PermutedDimsArray(
```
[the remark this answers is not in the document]
yeesian: Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/5107fe5fa14b92496eeca2cc58736c7f49cd2653, thanks!

src/raster/images.jl
```diff
@@ -0,0 +1,259 @@
...
+        imview(GPI_RGB, zerovalues, imgvalues, zerovalues)
+    elseif gci == GCI_BlueBand
+        zerovalues = zeros(eltype(imgvalues), size(imgvalues))
+        imview(GPI_RGB, zerovalues, zerovalues, imgvalues)
+    else
+        error(
+            """
+      Unknown GCI: $gci. Please file an issue at
+      https://github.com/yeesian/ArchGDAL.jl/issues if it should be supported.
+      """,
```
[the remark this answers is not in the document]
yeesian: There isn't a good reason, thanks for the catch! Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/5107fe5fa14b92496eeca2cc58736c7f49cd2653.

src/raster/images.jl
```diff
@@ -0,0 +1,259 @@
...
+    return ImageCore.PermutedDimsArray(
+        ImageCore.colorview(
+            colortype,
+            (ImageCore.normedview(img) for img in imgvalues)...,
+        ),
+        (2, 1),
+    )
+end
+
+function imview(gci::GDALColorInterp, imgvalues::Matrix)
```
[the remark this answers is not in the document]
yeesian: Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/5107fe5fa14b92496eeca2cc58736c7f49cd2653, thanks!

src/raster/images.jl
```diff
@@ -0,0 +1,259 @@
...
+end
+
+function imread(dataset::AbstractDataset, indices)
+    gpi, idxs = _paletteindices(dataset, indices)
+    return imview(gpi, (read(getband(dataset, i)) for i in idxs)...)
+end
+
+imread(dataset::AbstractDataset) = imread(dataset, 1:nraster(dataset))
+
+function imread(filename::String)
```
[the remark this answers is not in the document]
yeesian: Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/5107fe5fa14b92496eeca2cc58736c7f49cd2653, thanks!

src/constants.jl
```diff
@@ -0,0 +1,236 @@
...
+
+@enum(
+    OGRFieldSubType,
+    OFSTNone = 0,
+    OFSTBoolean = 1,
+    OFSTInt16 = 2,
+    OFSTFloat32 = 3,
+    OFSTJSON = 4,
+    OFSTMaxSubType = 5, # 4
+)
```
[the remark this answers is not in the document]
yeesian: > How is this file currently made?

This file is manually created/maintained, mostly from me looking up the corresponding `src/common.jl` file in GDAL.jl.

>  CEnum allows multiple fields to have the same value.

I think that's desired behavior for GDAL.jl, but awkward when it came to type-checking enums in this package (I forgot the specific example I had in mind, sorry; but it feels weird that enums aren't mutually exclusive). And yeah, I think the values for `wkb*` can overshoot the bounds of valid values for enums.

For all those reasons, I figured it's better to maintain an independent set of enums that

(i) [behaves like enums](https://github.com/yeesian/ArchGDAL.jl/blob/5107fe5fa14b92496eeca2cc58736c7f49cd2653/src/constants.jl), and
(ii) [can be converted to GDAL enums](https://github.com/yeesian/ArchGDAL.jl/blob/5107fe5fa14b92496eeca2cc58736c7f49cd2653/src/types.jl#L228-L534) when they're being passed to GDAL.jl (based on this [convert macro](https://github.com/yeesian/ArchGDAL.jl/blob/5107fe5fa14b92496eeca2cc58736c7f49cd2653/src/utils.jl#L1-L46) which replaces the old gdalenum macro)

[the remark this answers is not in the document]
visr: Ok I understand how this is working now, thanks. What do you mean with "behaves like enums", do CEnums not behave like enums? I think I would make the choice to stick with the original values to avoid confusion, though I understand your reasoning here as well.

src/tables.jl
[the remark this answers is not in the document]
visr: Ok, sounds good, thanks for the extra info!

src/constants.jl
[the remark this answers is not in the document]
yeesian: > What do you mean with "behaves like enums"

Oh, I was thinking mostly about having distinct values. I'm starting to remember that it might have to do with e.g. using them as keys in a dictionary and accidentally overriding the entries, resulting in unintuitive behavior:

```julia
julia> Dict(GDAL.GCI_YCbCr_CrBand => "a", GDAL.GCI_Max => "b")
Dict{GDAL.GDALColorInterp, String} with 1 entry:
  GCI_YCbCr_CrBand => "b"
```

Between warning people to use `string(...)` for CEnums when using them as keys versus having enums that enforce distinct values, I opted for the latter.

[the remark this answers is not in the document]
yeesian: > I think I would make the choice to stick with the original values to avoid confusion, though I understand your reasoning here as well.

Happy to write it up in the developer documentation if it might become a source of confusion/contention -- since the value of `GDAL.GCI_Max` might change from one version to the next, my reasoning was that we'd want to be using the "names" of enums rather than knowing about (much less using) the corresponding values that are used to model them as enums.

There are some cases ([[1]](https://github.com/yeesian/ArchGDAL.jl/blob/5107fe5fa14b92496eeca2cc58736c7f49cd2653/src/types.jl#L551-L559), [[2]](https://github.com/yeesian/ArchGDAL.jl/blob/5107fe5fa14b92496eeca2cc58736c7f49cd2653/src/types.jl#L536-L541)) where the underlying values are of interest, so the CEnum approach in GDAL.jl is still very useful to keep. But since that's already provided by GDAL.jl, I figured it'll be complementary/helpful for end-users to have a package which models the corresponding Enum (rather than CEnum) behavior and facilitate bi-directional conversion.

src/ogr/feature.jl
yeesian: Switched to `missing` in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/83b70259519581b1ca8070284f5c56d2f3d9e053, and opened https://github.com/yeesian/ArchGDAL.jl/issues/192 to ensure that the codebase is consistent in semantics after this PR.

src/tables.jl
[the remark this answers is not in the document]
yeesian: Captured this discussion in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/7351750705dcc178fb8ebffd2a7645713e69e65d.

src/constants.jl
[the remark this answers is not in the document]
visr: Great, sounds good. Instead of dev docs I would also be happy with a short comment above all the enums that their value is different from GDAL but they map correctly with at-convert.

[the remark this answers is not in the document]
yeesian: Documented this discussion in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/9c7069cbd30b4bf61fbe59901577a18c1950301f.

src/ogr/feature.jl
yeesian: Leaving this comment thread unresolved for you to go through it. Feel free to let me know if I might have missed anything.

src/raster/rasterband.jl
yeesian: Just to give an update here: I'm changing the function to return `missing` rather than `nothing` in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/abe94cdea5ff20233fb96d519ac25c87f8b3c9b1 based on the discussion in https://github.com/yeesian/ArchGDAL.jl/issues/192#issuecomment-851678784.

@visr for vis.

src/constants.jl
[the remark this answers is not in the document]
yeesian: > Instead of dev docs I would also be happy with a short comment above all the enums that their value is different from GDAL but they map correctly with at-convert.

Added in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/d5a5885057a53499965c83839c6aba5912e67191. Leaving this comment thread open for you to resolve or let me know if I missed anything

README.md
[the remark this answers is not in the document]
yeesian: Done in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/29c4a943a7cf1b6888561fa4774a8eceae5f07b7.

src/raster/rasterband.jl
yeesian: Leaving this comment thread unresolved for you to go through it. Feel free to let me know if I might have missed anything.

visr (approved):
Looks good. Big step to 1.0!

README.md
```diff
@@ -29,3 +32,70 @@ To test if it is installed correctly,
...
+To manage the dependencies of this package, we work with [environments](https://pkgdocs.julialang.org/v1/environments/):
+
+1. Navigate to the directory corresponding to the package:
+
+```julia
+shell> cd /Users/yeesian/.julia/dev/ArchGDAL
+/Users/yeesian/.julia/dev/ArchGDAL
+```
+
+2. Activate the environment corresponding to `Project.toml`):
```
evetion, line 64: Nitpick; This misses the step of actually starting Julia. Which you could do with `julia --project`

docs/make.jl
```diff
@@ -17,15 +19,15 @@ makedocs(
         "GDAL Datasets" => "datasets.md",
         "Feature Data" => "features.md",
         "Raster Data" => "rasters.md",
+        "Working with Images" => "images.md",
         "Tables Interface" => "tables.md",
         "Geometric Operations" => "geometries.md",
         "Spatial Projections" => "projections.md",
         # "Working with Spatialite" => "spatialite.md",
```
evetion, line 26: Maybe remove this commented out section? Edit: uncomment this, there seems to be a spatialite section.

docs/src/considerations.md
```diff
@@ -15,3 +15,82 @@ Here's a collection of references for developers who are interested:
...
+* For >2D arrays (multiple rasterbands), if they have a palette interpretation (or combination of color interpretations) that resolves to a valid colortype, ArchGDAL.jl uses the palette interpretation. If they don't have a palette interpretation, we throw an error.
+
+In general, `read()` will return `Array{UInt8}`, and `imread()` will return `Array{<:Colorant}`.
+
+## Tables.jl Interface
+
+The interface is implemented in [`src/tables.jl`](https://github.com/yeesian/ArchGDAL.jl/blob/master/src/tables.jl). The current API from GDAL makes it row-based in the conventions of Tables.jl. Therefore,
+
+* `ArchGDAL.Feature` meets the criteria for an [`AbstractRow`](https://tables.juliadata.org/dev/#Tables.AbstractRow-1) based on https://github.com/yeesian/ArchGDAL.jl/blob/a665f3407930b8221269f8949c246db022c3a85c/src/tables.jl#L31-L58.
+* `ArchGDAL.FeatureLayer` meets the criteria for an `AbstractRow`-iterator based on the previous bullet and meeting the criteria for [`Iteration`](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration) in https://github.com/yeesian/ArchGDAL.jl/blob/a665f3407930b8221269f8949c246db022c3a85c/src/base/iterators.jl#L1-L18
```
evetion, line 96: We could add a note that because there can be multiple layers in a file, there can be multiple tables.

docs/src/considerations.md
```diff
@@ -15,3 +15,82 @@ Here's a collection of references for developers who are interested:
...
+To differentiate 2d arrays of colors from 3d arrays with band as the third dimension:
+
+* For 2D arrays (a single rasterband), if they have a color interpretation, we use the color interpretation. If they don't have a color interpretation, ArchGDAL.jl defaults to "Grey".
+* For >2D arrays (multiple rasterbands), if they have a palette interpretation (or combination of color interpretations) that resolves to a valid colortype, ArchGDAL.jl uses the palette interpretation. If they don't have a palette interpretation, we throw an error.
+
+In general, `read()` will return `Array{UInt8}`, and `imread()` will return `Array{<:Colorant}`.
+
+## Tables.jl Interface
+
+The interface is implemented in [`src/tables.jl`](https://github.com/yeesian/ArchGDAL.jl/blob/master/src/tables.jl). The current API from GDAL makes it row-based in the conventions of Tables.jl. Therefore,
```
evetion, line 93: Note that this interface is for `vector` stuff only. Unlike the colors, which is `raster` stuff.

src/base/display.jl
```diff
@@ -1,13 +1,20 @@
-function Base.show(io::IO, drv::Driver)
-    drv.ptr == C_NULL && (return print(io, "NULL Driver"))
+function Base.show(io::IO, drv::Driver)::IO
+    if drv.ptr == C_NULL
+        print(io, "NULL Driver")
+        return io
+    end
```
evetion, line 5: This could be an `else`, makes the return single as well.

src/base/display.jl
```diff
@@ -1,13 +1,20 @@
...
+    return io
 end
 
-function Base.show(io::IO, dataset::AbstractDataset)
-    dataset.ptr == C_NULL && (return print(io, "NULL Dataset"))
+function Base.show(io::IO, dataset::AbstractDataset)::IO
+    if dataset.ptr == C_NULL
+        print(io, "NULL Dataset")
+        return io
+    end
```
evetion, line 14: Again, possible `else`, but with all the stuff underneath here, maybe let it be?

src/base/display.jl
```diff
@@ -1,13 +1,20 @@
-function Base.show(io::IO, drv::Driver)
-    drv.ptr == C_NULL && (return print(io, "NULL Driver"))
+function Base.show(io::IO, drv::Driver)::IO
+    if drv.ptr == C_NULL
+        print(io, "NULL Driver")
+        return io
+    end
     print(io, "Driver: $(shortname(drv))/$(longname(drv))")
+    return io
```
evetion, line 7: Why do we return `io`? Show methods in the language often don't return a thing (`nothing` if you want to be explicit). Same with other `show` methods below. Don't forget about the code annotation if you remove it.

src/base/display.jl
```diff
@@ -31,43 +38,58 @@ function Base.show(io::IO, dataset::AbstractDataset)
         println(io, "\nNumber of feature layers: $nlayers")
         ndisplay = min(nlayers, 5) # display up to 5 layers
```
evetion, line 39: That 5 could be a constant outside of this function? Note that we can also check for compactness with `compact = get(io, :compact, false)` and we could print all elements when given a plaintext mimetype, like `::MIME"text/plain"` as second argument to show. See this [thread](https://discourse.julialang.org/t/show-and-showcompact-on-custom-types/8493/10).

src/base/display.jl
```diff
@@ -143,80 +169,117 @@ function Base.show(io::IO, layer::AbstractFeatureLayer)
...
     else
         print(io, "Spatial Reference System: $projstr")
     end
+    return io
 end
 
-function Base.show(io::IO, geom::AbstractGeometry)
-    geom.ptr == C_NULL && (return print(io, "NULL Geometry"))
+function Base.show(io::IO, geom::AbstractGeometry)::IO
+    if geom.ptr == C_NULL
```
evetion, line 255: Is `===` an option to use here? Could be faster according to the [docs](https://docs.julialang.org/en/v1/manual/performance-tips/#Checking-for-equality-with-a-singleton). If yes, this occurs all over the codebase, would require a search/replace.

.JuliaFormatter.toml
```diff
@@ -0,0 +1,8 @@
+# Configuration file for JuliaFormatter.jl
+# For more information, see: https://domluna.github.io/JuliaFormatter.jl/stable/config/
```
evetion, line 2: I hope we can someday put all these config things into the Project.toml, would clean up nicely.

src/base/iterators.jl
```diff
@@ -1,39 +1,52 @@
...
 
-function Base.iterate(obj::BlockIterator, iter::Int=0)
+function blocks(raster::AbstractRasterBand)::BlockIterator{Int64}
+    return blocks(Int64, raster)
+end
+
+function Base.iterate(
+    obj::BlockIterator{T},
+    iter::T = 0,
+)::Union{Nothing,Tuple{Tuple{Tuple{T,T},Tuple{T,T}},T}} where {T}
```
evetion, line 49: That's one of the more interesting types I've seen written out.

src/base/iterators.jl
```diff
@@ -47,43 +60,78 @@ function Base.iterate(obj::BlockIterator, iter::Int=0)
...
+windows(raster::AbstractRasterBand)::WindowIterator{Int64} =
+    windows(Int64, raster)
+
+function Base.iterate(
+    obj::WindowIterator{T},
+    iter::T = 0,
+)::Union{Nothing,Tuple{NTuple{2,UnitRange{T}},T}} where {T<:Integer}
     handle = obj.blockiter
     next = Base.iterate(handle, iter)
     next == nothing && return nothing
```
evetion, line 100: This could be an `===`, same check occurs more often in the codebase.

evetion (approved):

evetion:
Wow, this is some undertaking, good work!

For the next time I'd advise (also myself in other projects) to split the formatting into a separate PR that can be merged quickly and only then start the breaking changes. I've stopped reviewing after scrolling through unending green changes that only changed whitespace.

src/base/display.jl
yeesian: Opened #202 for it.

.JuliaFormatter.toml
yeesian: Yeah that'll be nice :)

src/base/iterators.jl
yeesian: Opened #202 for it

src/base/iterators.jl
yeesian: 😛

src/base/display.jl
yeesian: Opened #203 for it

src/base/display.jl
yeesian: https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/aecd4d89c1ef7e892e7e953001ff71913be18b24

docs/src/considerations.md
yeesian: Thank you! https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/b53179d0d133d510a3427bc4cefb4a4b9003bbd9

docs/src/considerations.md
yeesian: Yeah good callout :) https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/4faa97979e5317727ca9f3a3d8a85ddfa88f071d

docs/make.jl
yeesian: Added a note for it in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/2fd5aeec811f4abd159de4c4a818407914e4b9eb, thanks!

README.md
yeesian: Yeah that's a good point -- I did it in the REPL's shell mode, so I forgot documenting this step for others. Updated in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/26a8010f1e904bda8ce345792aac1d29297f5af5.

src/base/display.jl
yeesian: Yeah the [pattern](https://danp.net/posts/reducing-go-nesting/) was more for "check NULL, early return", so I'll let it be, thanks though!

src/base/display.jl
yeesian: > Why do we return `io`? Show methods in the language often don't return a thing 

Thanks! I recall trying to look it up and not getting any guidance on what the convention was (show/print implementations often calling show/print). Done in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/c615a992d155209fd546ba447aab7333c8570374.

yeesian:
> For the next time I'd advise (also myself in other projects) to split the formatting into a separate PR that can be merged quickly and only then start the breaking changes. I've stopped reviewing after scrolling through unending green changes that only changed whitespace.

Yeah I'm really sorry about this, and appreciate you taking the time to review and provide feedback still. I didn't realize it was needed at the start, and only came upon its importance midway (when I found myself spending a lot of time formatting code semi-arbitrarily and it was starting to weigh on me). It was exhausting work and I was worried about losing momentum/motivation, so I decided to continue with it rather than abandoning the effort.

yeesian:
@rafaqz there's been quite a lot of commits since https://github.com/yeesian/ArchGDAL.jl/pull/179#issuecomment-841599165:

> I like the direction here, and I'll start using this as my working branch to iron out any problems.
>
> Besides the return value on `getnodatavalue` introducing a bug, this doesn't seem to break anything in GeoData.jl.

I don't think any of them would change your experience of working with the latest version of this branch, but feel free to let me know if that isn't the case for you.

src/raster/images.jl
```diff
@@ -0,0 +1,264 @@
...
+    return if gpi == GPI_Gray
+        imview(ColorTypes.Gray, imgvalues)
+    else
+        error(
+            """
+            Unsupported GPI: $gpi. Please file an issue at
+            https://github.com/yeesian/ArchGDAL.jl/issues if it should be supported.
+            """,
+        )
+    end
```
visr, line 45: I run into this error with `GPI_RGB`. In the other methods below `GPI_RGB` does seem to be supported however. Is there an elseif missing here or is it more complicated than that?
```julia
using ArchGDAL
url = "https://neo.sci.gsfc.nasa.gov/archive/geotiff/CERES_INSOL_M/CERES_INSOL_M_2020-07.TIFF"
ArchGDAL.imread("/vsicurl/" * url)
```

```
ERROR: LoadError: Unsupported GPI: GPI_RGB. Please file an issue at
https://github.com/yeesian/ArchGDAL.jl/issues if it should be supported.

Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:33
 [2] imview(gpi::ArchGDAL.GDALPaletteInterp, imgvalues::Matrix{UInt8})
   @ ArchGDAL C:\Users\visser_mn\.julia\dev\ArchGDAL\src\raster\images.jl:39
 [3] imread(dataset::ArchGDAL.Dataset, indices::UnitRange{Int64})
   @ ArchGDAL C:\Users\visser_mn\.julia\dev\ArchGDAL\src\raster\images.jl:250
 [4] imread
   @ C:\Users\visser_mn\.julia\dev\ArchGDAL\src\raster\images.jl:253 [inlined]
 [5] #184
   @ C:\Users\visser_mn\.julia\dev\ArchGDAL\src\raster\images.jl:257 [inlined]
 [6] read(f::ArchGDAL.var"#184#185", args::String; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ArchGDAL C:\Users\visser_mn\.julia\dev\ArchGDAL\src\context.jl:267
 [7] read
   @ C:\Users\visser_mn\.julia\dev\ArchGDAL\src\context.jl:265 [inlined]
 [8] imread(filename::String)
   @ ArchGDAL C:\Users\visser_mn\.julia\dev\ArchGDAL\src\raster\images.jl:256
```

yeesian: Oh I haven't learnt how to handle "single channel" matrices -- since RGB channels are typically three matrices, rather than a single matrix?

```
julia> using ArchGDAL

julia> url = "https://neo.sci.gsfc.nasa.gov/archive/geotiff/CERES_INSOL_M/CERES_INSOL_M_2020-07.TIFF"
"https://neo.sci.gsfc.nasa.gov/archive/geotiff/CERES_INSOL_M/CERES_INSOL_M_2020-07.TIFF"

julia> ds = ArchGDAL.read("/vsicurl/" * url)
GDAL Dataset (Driver: GTiff/GeoTIFF)
File(s):
  /vsicurl/https://neo.sci.gsfc.nasa.gov/archive/geotiff/CERES_INSOL_M/CERES_INSOL_M_2020-07.TIFF

Dataset (width x height): 1440 x 720 (pixels)
Number of raster bands: 1
  [GA_ReadOnly] Band 1 (Palette): 1440 x 720 (UInt8)
```

Looking into how it should have been handled

rafaqz:
@yeesian thanks I'll have to update and try it out. 

Also FYI this PR will be a reliable way to replace all open datasets nested anywhere in a return value object. https://github.com/JuliaObjects/Accessors.jl/pull/23

It may produce errors with types that are not compatible with ConstructionBase.jl, but only in the case that they would contain a pointer to a closed file anyway, which is a worse error. This error would also be immediate rather than delayed, so you could use a `try catch` block and a message to either define `constructorof` for the object, or not return an open dataset. 

I can make a PR here when it's done if that helps.

src/raster/images.jl
yeesian: > Is there an elseif missing here or is it more complicated than that?

Sorry it took me this long to respond here: it's addressed in https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/317272f99f8372203de67868b87eb2fb66f88bab (which requires us to expose `GDAL.GDALColorEntry` and continue supporting some of the colortable methods). The implementation is less than ideal, but I want to just capture the API and get things working first.

yeesian: Sorry for the trouble: re-requesting your review for https://github.com/yeesian/ArchGDAL.jl/commit/317272f99f8372203de67868b87eb2fb66f88bab, and to confirm that the example you gave in https://github.com/yeesian/ArchGDAL.jl/pull/179#discussion_r647698681 is working for you now.

yeesian:
@rafaqz gentle ping to follow-up on https://github.com/yeesian/ArchGDAL.jl/pull/179#issuecomment-860337138

rafaqz:
Ok this works fine for GeoData.jl, with a few tweaks for the new default nodata value and uppercase const variables.

I'm not sure about the choice to return `missing` when there is no `nodata` value. I found `nothing` more appropriate as it suggests that there simply is no nodata value. ArchGDAL is not also returning an array of `Union{T,Missing}`. NCDatasets.jl does return arrays where the nodata value has been converted to `missing`, and that's what a value of `missing` suggests to me.

yeesian:
@rafaqz regarding https://github.com/yeesian/ArchGDAL.jl/pull/179#issuecomment-864701896, the nice thing about `missing` is for the following reason:

```julia
julia> [1 2;
       missing 4] + ones(2,2)
2×2 Matrix{Union{Missing, Float64}}:
 2.0       3.0
  missing  5.0
```
In those cases where there is no nodata value, it looks like a more reasonable behavior compared to `nothing`:
```julia
julia> [1 2;
       nothing 4] + ones(2,2)
ERROR: MethodError: no method matching +(::Nothing, ::Float64)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:560
  +(::Base.TwicePrecision, ::Number) at twiceprecision.jl:267
  +(::ColorTypes.AbstractGray{T} where T, ::Number) at /Users/yeesian/.julia/packages/ColorVectorSpace/71ig5/src/ColorVectorSpace.jl:285
  ...
```
The fact that NCDatasets returns `missing` seems more like an argument for going with `missing` (rather than `nothing`).

> ArchGDAL is not also returning an array of `Union{T,Missing}`.

I've opened #192 to make sure I systematically scan the package for NULL semantics later, but just want to make sure we're okay with going down that path first.

rafaqz:
>The fact that NCDatasets returns missing seems more like an argument for going with missing (rather than nothing).

I would argue the exact opposite... NCDatasets returns an array that actually contains `missing` values. ArchGDAL doesn't. In terms of handling these things in GeoData.jl they have a totally different meaning.

I don't understand this example either:
```julia
[1 2;       
nothing 4] + ones(2,2)
```

There are no nothing values actually in the raster when there is no nodata value. It's all numbers.

rafaqz:
To be clear I'm talking only about the result of :
```julia
ArchGDAL.getnodatavalue(band)
```

When there is actually not nodata in the array.

rafaqz:
A problem with returning `missing` is in many cases there is actually already a nodata value in the raster that isn't recorded for whatever reason - and when we return missing we then have 2 missing values active in the same raster. Returning `nothing` ensures that the user has to manually specify the missing value instead of trusting that the existing missing value was detected. 

I had problems with files from R raster where something like this had happened and multiple missing values were present in the file, which of course broke a lot of things. I think it's better to throw an error than let this happen unintentionally.

yeesian:
Thanks for correcting my misunderstanding with https://github.com/yeesian/ArchGDAL.jl/pull/179#issuecomment-864710188 and https://github.com/yeesian/ArchGDAL.jl/pull/179#issuecomment-864711101. I see where you're coming from and agree with you now. I'll describe what I now understand (let me know if I'm still misunderstanding):

1. The nodata value isn't meant to be used as a "sentinel value" in the raster output. If ArchGDAL wants to provide a convenience function for returning `Array{Union{T, Missing}}`, the treatment of missing data should be based on the appropriate use of `getmaskband()` (https://trac.osgeo.org/gdal/wiki/rfc15_nodatabitmask for details).
2. Even if ArchGDAL does provide a convenience function for returning `Array{Union{T, Missing}}`, the proper place for using `missing` values is in the array being constructed, and not the value being returned from `getnodatavalue(band)`.

For those reasons, I've changed back from `missing` to `nothing` in this PR now: https://github.com/yeesian/ArchGDAL.jl/pull/179/commits/800bf2c89bc418597f45eea50ee6cbb82b169730.

rafaqz:
> 1. The nodata value isn't meant to be used as a "sentinel value" in the raster output. If ArchGDAL wants to provide a convenience function for returning Array{Union{T, Missing}}, the treatment of missing data should be based on the appropriate use of getmaskband() (https://trac.osgeo.org/gdal/wiki/rfc15_nodatabitmask for details).

Not exactly... I think `nodata` can be a sentinal value, but not in the case that there is no missing value at all in the object, or more accurately when GDAL doesn't know about one.

I basically agree with everything written here in the current implementation:
https://github.com/yeesian/ArchGDAL.jl/blob/1dc8ca8360f2e6a2ee2319fd63c3ba973034f8a6/src/raster/rasterband.jl#L162-L188

> 2. Even if ArchGDAL does provide a convenience function for returning Array{Union{T, Missing}}, the proper place for using missing values is in the array being constructed, and not the value being returned from getnodatavalue(band).

I dont know really :)

visr (approved):

rafaqz (approved):

PR #207, opened 2021-06-21
looser ImageCore and ColorTypes deps

rafaqz:
There are still packages with older dependencies, and these should work equally well here?


yeesian:
Thank you!

Issue #210, opened 2021-06-29
ArchGDAL.Table not defined in v0.7


juliohm:
I am trying to update downstream packages but they are complaining that `ArchGDAL.Table` doesn't exist. The latest stable documentation says it should exist still. Could you please help identify the issue?

visr:
Hmm, it seems the stable docs have not been updated since January, so something is wrong there.
Dev docs are updated however: https://yeesian.com/ArchGDAL.jl/dev/tables/

`ArchGDAL.Table` was not needed anymore, since we could just implement the Tables interface directly on `AbstractFeatureLayer`.

So this:
https://github.com/JuliaEarth/GeoTables.jl/blob/79c9ffb22724cde58f51dfbb77a48ebaedbf0048/src/GeoTables.jl#L37-L38

can be condensed to `table = AG.getlayer(file, layer)`.

juliohm:
Thanks, I am trying to update GADM.jl here: https://github.com/JuliaGeo/GADM.jl/blob/f04d7c2d4c63ac9ffa14927b5f1404ae8f2a98c3/src/GADM.jl#L108

I understand that in the new version the layer is already a table so I replaced the line `table = ArchGDAL.Table(layer)` with `table = layer`. There are other errors popping up now. Is there a list of breaking changes and recommendations to update the code?

visr:
The top post of https://github.com/yeesian/ArchGDAL.jl/pull/179 has a list of breaking changes. I just added a note about `Table` to it, since that was added later.

yeesian:
Fixed by https://github.com/yeesian/ArchGDAL.jl/commit/6b545fae4a7c4bbc1114a96fac80628030ea7f6e.

PR #211, opened 2021-06-30
Add DOCUMENTER_KEY as ssh deploy key to tagbot

felixcremer:
This should enable TagBot to also trigger documentation builds for the tagged versions.
This fix the problems of the generation of stable docs as noticed in #210.
I assume, that the DOCUMENTER_KEY is already generated, because the development version of the docs are working fine.

See https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions


yeesian:
Thank you!

PR #216, opened 2021-07-05
Enable push_preview for deploying docs

yeesian:


visr (approved):
Looks like it worked: https://yeesian.com/ArchGDAL.jl/previews/PR216/

Unrelated, but perhaps it's good to update the url in the repo about settings, right now it points to yeesian.github.io/archgdal.jl/latest/. It still works, but latest is no longer being used, it's dev and stable.

yeesian:
> It still works, but latest is no longer being used, it's dev and stable.

Updated to `stable`, thanks! (I am aware of #210, and am hoping something like #217 might address it)

PR #217, opened 2021-07-05
Update versions.js

yeesian:
It seems that the gh-pages branch is very non-standard and it's unclear to me how to fix it. Just trying things out one-at-a-time. See https://github.com/yeesian/ArchGDAL.jl/issues/210#issuecomment-870678574 for context.


visr:
I don't think this is complete. See https://github.com/yeesian/ArchGDAL.jl/tree/gh-pages. There are no folders for those versions, so adding them to a list won't help. v0.6 and v0.7 would also need to be added as symlinks that point to the builds of the latest corresponding patch.

Not sure this is the perfect example, but you could look at https://github.com/Deltares/Wflow.jl/tree/gh-pages for reference.

yeesian:
In that case, I think we can do it for v0.7, but not v0.6?

visr:
Yeah perhaps that would work, we can try. I haven't manually worked with this branch before though, so indeed we have to try and see (or ask on #documentation).

yeesian:
(For posterity: it seems that the right way of doing this is via https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#Documentation-Versions. I've just created [`v0.7.1+doc1`](https://github.com/yeesian/ArchGDAL.jl/releases/tag/v0.7.1%2Bdoc1) and the run in https://github.com/yeesian/ArchGDAL.jl/actions/runs/1002254221 seems to have worked: https://github.com/yeesian/ArchGDAL.jl/commit/6b545fae4a7c4bbc1114a96fac80628030ea7f6e.

PR #218, opened 2021-07-05
Workflow for cleaning up preview and history.

yeesian:
By following https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#gh-pages-Branch.

### DIFF
--- a/.github/workflows/DocCleanup.yml
+++ b/.github/workflows/DocCleanup.yml
@@ -1,0 +1,26 @@
+name: Doc Preview Cleanup
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


yeesian:
It became relevant now that we have previews per #216.

visr (approved):
So this will delete the preview for PR n when PR n closes?

So if you merge this, this one will still be available:
https://yeesian.com/ArchGDAL.jl/previews/PR216/
But this preview should be removed:
https://yeesian.com/ArchGDAL.jl/previews/PR218/

Let's see :)

PR #219, opened 2021-07-05
Update to latest GDAL

yeesian:


yeesian:
My bad, this PR came from a misunderstanding of https://pkgdocs.julialang.org/v1/compatibility/

visr:
~~Why is this needed? Is it no longer compatible with GDAL.jl 1.1, for instance since you need some new functions from https://github.com/JuliaGeo/GDAL.jl/pull/109?~~

Issue #177, opened 2021-03-05
Handle NULL values in features correctly


evetion:
If I create the following geopackage (shown with ogrinfo)

```bash
Geometry Column = geom
date: DateTime (0.0)
test: Integer(Boolean) (0.0)
name: String (255.0)
number: Integer (0.0)
OGRFeature(null):1
  date (DateTime) = (null)         <-- note these nulls
  test (Integer(Boolean)) = 0
  name (String) = test
  number (Integer) = 1
  POINT (0.029535864978903 0.578059071729958)

OGRFeature(null):2
  date (DateTime) = 2021/03/12 00:00:00
  test (Integer(Boolean)) = 1
  name (String) = (null)         <-- note these nulls
  number (Integer) = (null)    <-- note these nulls
  POINT (-0.355836849507736 0.20393811533052)
```
and apply `getfield` to it, ArchGDAL fails with:
```julia
ERROR: Failed to fetch datetime at index 0
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] asdatetime(::ArchGDAL.Feature, ::Int32) at /Users/evetion/.julia/packages/ArchGDAL/j9NPL/src/ogr/feature.jl:318
 [3] getfield(::ArchGDAL.Feature, ::Int32) at /Users/evetion/.julia/packages/ArchGDAL/j9NPL/src/ogr/feature.jl:378
 [4] getfield(::ArchGDAL.Feature, ::Symbol) at /Users/evetion/.julia/packages/ArchGDAL/j9NPL/src/ogr/feature.jl:384
```

It seems that `isfieldset` doesn't trigger correctly for NULL DateTime fields. If we actually fill it, we get a result, but ArchGDAL uses `getdefault`, which returns actual data values (`name` and `number` should both be `null`). These should be either `nothing` or `missing`.

```julia
(date = Dates.DateTime("2021-03-12T00:00:00"), test = 1, name = "", number = 0, geom = Geometry: POINT (-0.355836849507736 0.20393811533052))
```

On a last note, we don't parse the width of integers, so we don't parse the `test=1` as the `bool` it is.

Ping to @vernimmen who ran into this when opening a GeoPackage with NULLs via GeoDataFrames.jl.

yeesian:
Sorry for being late to respond here: is it possible to provide a small working example here?

evetion:
Sure, I've uploaded a small test file.

```julia
using ArchGDAL

[...]

PR #220, opened 2021-07-06
Add possible subtypes

evetion:
In order to write slightly more exotic types as `Int16`, `Bool` and `Float32`, GDAL has subtypes. However, the main type must still be set and currently ArchGDAL doesn't support these types.

I pirated it myself in https://github.com/evetion/GeoDataFrames.jl/blob/master/src/io.jl#L14, this is the upstream fix.

For the actual usage of these subtypes (in writing) see https://github.com/evetion/GeoDataFrames.jl/blob/master/src/io.jl#L103. Note that funnily enough, this enables write support of these types earlier than read support (#177).

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -303,8 +303,11 @@ eval(
     @convert(
         OGRFieldType::DataType,
         OFTInteger::Int32,
+        OFTInteger::Int16,
+        OFTInteger::Bool,
         OFTIntegerList::Vector{Int32},
         OFTReal::Float64,
+        OFTReal::Float32,
         OFTRealList::Vector{Float64},
         OFTString::String,
         OFTStringList::Vector{String},
--- a/src/types.jl
+++ b/src/types.jl
@@ -302,12 +302,12 @@ eval(
 eval(
     @convert(
         OGRFieldType::DataType,
-        OFTInteger::Int32,
-        OFTInteger::Int16,
         OFTInteger::Bool,
+        OFTInteger::Int16,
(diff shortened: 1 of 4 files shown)


src/types.jl
```diff
@@ -303,8 +303,11 @@ eval(
     @convert(
         OGRFieldType::DataType,
         OFTInteger::Int32,
+        OFTInteger::Int16,
+        OFTInteger::Bool,
```
yeesian, line 307: To give an example (see my other comment): I suspect the inclusion of these two lines might mangle the conversion of `OFTInteger` to `Int32` because of the `Dict`-based implementation of the `convert` macro.

yeesian (changes requested):
The current `convert` macro is based on dictionaries: https://github.com/yeesian/ArchGDAL.jl/blob/7f9157dd988c8d7c3ffe7ecdf6ad039b471d358d/src/utils.jl#L32-L44

For this to work, we need to modify the `convert` macro to implement the `convert` functions individually like you did in https://github.com/evetion/GeoDataFrames.jl/blob/3d2fb76cf066a0b2c0027b4e291a644401db0f35/src/io.jl#L14-L16.

src/types.jl
evetion: Good point, I missed the reverse one. So because of the `Dict`, the macro has become order dependent.

yeesian: Thanks!

yeesian (approved):
Thanks!

PR #227, opened 2021-08-19
Generalize exception type

timholy:
In preparation for https://github.com/JuliaLang/julia/pull/41885

### DIFF
--- a/test/test_array.jl
+++ b/test/test_array.jl
@@ -171,7 +171,7 @@ const AG = ArchGDAL;
             nbands = 0,
             dtype = UInt8,
         ) do dataset
-            @test_throws ArgumentError AG.RasterDataset(dataset)
+            @test_throws Union{ArgumentError,MethodError} AG.RasterDataset(dataset)
             @test_throws DimensionMismatch AG._common_size(dataset)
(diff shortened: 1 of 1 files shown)


test/test_array.jl
```diff
@@ -139,7 +139,7 @@ const AG = ArchGDAL;
         AG.readraster("ospy/data4/aster_write.img"; flags = AG.OF_UPDATE) do ds
             @test sprint(print, ds) == """
             GDAL Dataset (Driver: HFA/Erdas Imagine Images (.img))
-            File(s):
```
yeesian, line 142: the tests are whitespace sensitive within the strings, so I suspect this will fail

yeesian:
(It seems that the workflows needed to be manually triggered for PRs by first-time contributors; also sorry for pushing a commit earlier that clashed with your fix and squash)

Thanks!

visr:
> It seems that the workflows needed to be manually triggered for PRs by first-time contributors

Under settings / actions you can change this to be only for new accounts.

yeesian:
Ah thanks! Changed it to be only for new accounts

PR #230, opened 2021-08-27
New convert macro using a `const Dict` outside the `convert` function…

### LINKED ISSUE #221: Allow for multi-way enum conversions
Copied from a comment in https://github.com/yeesian/ArchGDAL.jl/pull/220:

The current `convert` macro is based on dictionaries: https://github.com/yeesian/ArchGDAL.jl/blob/7f9157dd988c8d7c3ffe7ecdf6ad039b471d358d/src/utils.jl#L32-L44

For this to work, we need to modify the `convert` macro to implement the `convert` functions individually like in https://github.com/evetion/GeoDataFrames.jl/blob/3d2fb76cf066a0b2c0027b4e291a644401db0f35/src/io.jl#L14-L16.

mathieu17g:
… and a `convert` function of the form `convert(::Type{type1}, ::Type{subtype2}) = subtype1` when `type1` is an `Enum` type and `type2` is `DataType`

Fixes #221 

Hello @yeesian, I have a proposal for `@convert` macro evolution. 

## Modifications
I modified to items:
1) Moved `fwd_map` and `rev_map` to a `const` outside the `convert` function
2) Used the `convert` signature found by @evetion when one type is an `Enum` and the other is `DataType`

It does not work for:
```julia
(
    GDALDataType::ImageCore.Normed,
    GDT_Byte::ImageCore.N0f8,
    GDT_UInt16::ImageCore.N0f16,
    GDT_UInt32::ImageCore.N0f32,
)
```
It fails at compile time although both old and new `@convert` macro expands into similar code. I have not figured out why, yet. There seem to be an issue with declaring a `const Dict` with `ImageCore.Normed` types in it.  
Therefore I kept the old `@convert` macro for these types conversion, renaming it `@convert_old` and using it for the above conversions.

## Performance
The new `@convert` macro brings a x2 speed up on conversion to Tables:

With the master version:
```julia
julia> ds = get_road_ds(;clean=false)
       layer = AG.getlayer(ds, 0)
       @benchmark Tables.columns($layer)
BenchmarkTools.Trial: 21 samples with 1 evaluation.
 Range (min … max):  227.551 ms … 254.846 ms  ┊ GC (min … max): 6.85% … 10.86%
 Time  (median):     241.586 ms               ┊ GC (median):    8.71%
 Time  (mean ± σ):   242.752 ms ±   5.826 ms  ┊ GC (mean ± σ):  8.74% ±  0.97%

                             █  ▁        ▁   ▁                   
  ▆▁▁▁▁▁▁▁▁▁▁▁▁▁▆▁▁▁▁▁▁▁▆▁▁▁▁█▆▁█▆▁▁▁▁▁▁▆█▁▆▆█▁▁▁▆▆▁▁▁▁▁▁▁▁▁▁▁▆ ▁
  228 ms           Histogram: frequency by time          255 ms <

 Memory estimate: 133.47 MiB, allocs estimate: 1427522.
```

With the new `@macro` macro
```julia
julia> ds = get_road_ds(;clean=false)
       layer = AG.getlayer(ds, 0)
       @benchmark Tables.columns($layer)
BenchmarkTools.Trial: 41 samples with 1 evaluation.
 Range (min … max):   94.682 ms … 264.763 ms  ┊ GC (min … max): 0.00% … 28.53%
 Time  (median):     106.135 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   122.680 ms ±  45.326 ms  ┊ GC (mean ± σ):  5.07% ±  7.51%

    █▃▁                                                          
  ▇▇███▆▆▆▆▇▄▁▄▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▄▁▆ ▁
  94.7 ms          Histogram: frequency by time          265 ms <

 Memory estimate: 8.98 MiB, allocs estimate: 373860.
```

## Additionnal question
Is the call to `eval` necessary in types.jl after calling the `@convert` macro ?

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -255,7 +255,7 @@ eval(
 )
 
 eval(
-    @convert(
+    @convert_old(
         GDALDataType::ImageCore.Normed,
         GDT_Byte::ImageCore.N0f8,
         GDT_UInt16::ImageCore.N0f16,
--- a/src/types.jl
+++ b/src/types.jl
@@ -255,7 +255,7 @@ eval(
 )
 
 eval(
-    @convert_old(
+    @convert(
         GDALDataType::ImageCore.Normed,
         GDT_Byte::ImageCore.N0f8,
         GDT_UInt16::ImageCore.N0f16,
--- a/src/types.jl
+++ b/src/types.jl
@@ -235,317 +235,281 @@ mutable struct ColorTable
     ptr::GDAL.GDALColorTableH
 end
 
-eval(
-    @convert(
-        GDALDataType::GDAL.GDALDataType,
-        GDT_Unknown::GDAL.GDT_Unknown,
-        GDT_Byte::GDAL.GDT_Byte,
-        GDT_UInt16::GDAL.GDT_UInt16,
-        GDT_Int16::GDAL.GDT_Int16,
-        GDT_UInt32::GDAL.GDT_UInt32,
-        GDT_Int32::GDAL.GDT_Int32,
-        GDT_Float32::GDAL.GDT_Float32,
-        GDT_Float64::GDAL.GDT_Float64,
-        GDT_CInt16::GDAL.GDT_CInt16,
-        GDT_CInt32::GDAL.GDT_CInt32,
-        GDT_CFloat32::GDAL.GDT_CFloat32,
-        GDT_CFloat64::GDAL.GDT_CFloat64,
-        GDT_TypeCount::GDAL.GDT_TypeCount,
-    )
+@convert(
+    GDALDataType::GDAL.GDALDataType,
+    GDT_Unknown::GDAL.GDT_Unknown,
+    GDT_Byte::GDAL.GDT_Byte,
+    GDT_UInt16::GDAL.GDT_UInt16,
+    GDT_Int16::GDAL.GDT_Int16,
+    GDT_UInt32::GDAL.GDT_UInt32,
+    GDT_Int32::GDAL.GDT_Int32,
+    GDT_Float32::GDAL.GDT_Float32,
+    GDT_Float64::GDAL.GDT_Float64,
+    GDT_CInt16::GDAL.GDT_CInt16,
+    GDT_CInt32::GDAL.GDT_CInt32,
+    GDT_CFloat32::GDAL.GDT_CFloat32,
+    GDT_CFloat64::GDAL.GDT_CFloat64,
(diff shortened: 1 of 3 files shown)


src/types.jl
```diff
@@ -235,317 +235,281 @@ mutable struct ColorTable
...
-        OFTTime::Dates.Time,
-        OFTDateTime::Dates.DateTime,
-        OFTInteger64::Int64,
-        OFTInteger64List::Vector{Int64},
-    )
+@convert(
+    OGRFieldType::DataType,
+    OFTInteger::Bool,
+    OFTInteger::Int16,
+    OFTInteger::Int32,  # default type comes last
```
yeesian, line 298: (optional) Given that we're using dictionaries for representing these mappings, it has always made me a little uncomfortable that there's no good way of asserting "default type comes last" for mappings that are not bijective.
* Should it be included in the docstring?
* Should we have unit tests for them?
* Should we organize the implementation such that it becomes clear how we're enforcing this by construction?

yeesian (approved):
I started an earlier review and saw that you've basically addressed everything in your subsequent commits! :)

The only remaining comment I have is optional and would not block this PR -- but I'm curious to hear your thoughts on it.

src/types.jl
```diff
@@ -235,317 +235,281 @@ mutable struct ColorTable
...
+    OFTIntegerList::Vector{Int32},
+    OFTReal::Float32,
+    OFTReal::Float64,  # default type comes last
+    OFTRealList::Vector{Float64},
+    OFTString::String,
+    OFTStringList::Vector{String},
+    OFTWideString::Nothing,
+    OFTWideStringList::Nothing,
```
yeesian, line 306: (Just a comment that's unrelated to this PR)

This caught my eye (#231) as I was looking for examples where "default type comes last" is happening for the reverse mapping, and it looks like its the case in the current implementation too:

```julia
julia> convert(ArchGDAL.OGRFieldType, Nothing)
OFTWideStringList::OGRFieldType = 7
```

src/types.jl
yeesian: After thinking more about it, I realized these are requests (outside of the scope of this PR) rather than questions, so I've opened an issue for it in #232 instead.

PR #233, opened 2021-08-28
Deprecate OFTWideString

### LINKED ISSUE #231: Implement datatype conversions for OFTWideStrings
We [currently map them to `nothing`](https://github.com/yeesian/ArchGDAL.jl/blob/f4406a55c5e2d42e0617e5e686d87f71d376d4be/src/types.jl#L314-L315), which doesn't look correct to me.

yeesian:
(I wouldn't consider this a breaking change since we've never properly supported that datatype, but let me know if I am wrong on that front.)

Fixes #231.


visr (approved):
Looks good to me.

PR #236, opened 2021-08-29
Return the default geotransform rather than throwing an error.

### LINKED ISSUE #190: Define default behavior for geotransform
I learnt from https://github.com/evetion/GeoArrays.jl/blob/6d7fb9bd05c3a2a2ca58129cbd5427a4bf491b92/src/geoutils.jl#L1-L12 and https://github.com/rafaqz/GeoData.jl/blob/ef0d458af45999942754aeebfbc5c8154351f40e/src/sources/gdal.jl#L176-L177 that https://github.com/yeesian/ArchGDAL.jl/blob/14b92dac074eb70244ac64570b64ae74b539b3c7/src/dataset.jl#L629-L631

yeesian:
Fixes #190.

~~@visr There's a small drop in test coverage due to the drop in test coverage for [`cplerr`](https://github.com/yeesian/ArchGDAL.jl/blob/f98a7cdce3482691ab526e3fbbf796d23aec35ed/src/utils.jl#L137-L143). However, I haven't been able to come up with a scenario in which I'll actually get `CE_Failure` without GDAL.jl throwing an `GDAL.GDALError`. (Which I think is a good thing!)~~ (Update: fixed by Felix in https://github.com/yeesian/ArchGDAL.jl/pull/236#discussion_r698527429)

### DIFF
--- a/src/dataset.jl
+++ b/src/dataset.jl
@@ -714,7 +714,8 @@ function getgeotransform!(
     @assert length(transform) == 6
     result = GDAL.gdalgetgeotransform(dataset.ptr, pointer(transform))
     if result != GDAL.CE_None
-        transform .= (0, 1, 0, 0, 0, 1)
+        # The default geotransform.
+        transform .= (0.0, 1.0, 0.0, 0.0, 0.0, 1.0)
     end
     return transform
 end
--- a/src/dataset.jl
+++ b/src/dataset.jl
@@ -713,7 +713,9 @@ function getgeotransform!(
 )::Vector{Cdouble}
     @assert length(transform) == 6
     result = GDAL.gdalgetgeotransform(dataset.ptr, pointer(transform))
-    @cplerr result "Failed to get geotransform"
+    if result != GDAL.CE_None
+        transform .= (0,1,0,0,0,1)
+    end
     return transform
 end
 
--- a/src/dataset.jl
+++ b/src/dataset.jl
@@ -714,7 +714,7 @@ function getgeotransform!(
     @assert length(transform) == 6
     result = GDAL.gdalgetgeotransform(dataset.ptr, pointer(transform))
     if result != GDAL.CE_None
-        transform .= (0,1,0,0,0,1)
+        transform .= (0, 1, 0, 0, 0, 1)
     end
     return transform
 end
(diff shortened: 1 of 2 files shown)


src/dataset.jl
```diff
@@ -713,7 +713,9 @@ function getgeotransform!(
 )::Vector{Cdouble}
     @assert length(transform) == 6
     result = GDAL.gdalgetgeotransform(dataset.ptr, pointer(transform))
-    @cplerr result "Failed to get geotransform"
+    if result != GDAL.CE_None
+        transform .= (0,1,0,0,0,1)
```
evetion: Nitpick, making these floats is more explicit.

test/test_dataset.jl
```diff
@@ -50,8 +50,7 @@ const AG = ArchGDAL;
         AG.getlayer(dataset1, 0) do layer1
             @test AG.nfeature(layer1) == 4
         end
-        # the following test is to cover an example of `macro cplerr(code, message)`
```
evetion, line 53: Is there still a test that covers the `cplerr` macro now?

evetion (approved):

test/test_dataset.jl
yeesian: Unfortunately there isn't. That's why the unit test coverage has gone down by a bit. I haven't been able to find an example of a scenario where we get `CE_Failure` without GDAL.jl throwing a `GDAL.GDALError`. Happy to add such a test in if someone discovers it.

src/dataset.jl
yeesian: Done in https://github.com/yeesian/ArchGDAL.jl/pull/236/commits/71a01e8682cd03e434f5eb7a237c976babd95c7d.

test/test_dataset.jl
felixcremer: You can use the setproj! function:
```julia
        @test_throws ErrorException AG.setproj!(dataset1, "nonsensestring")
```

yeesian: Thank you! https://github.com/yeesian/ArchGDAL.jl/pull/236/commits/330a3c951d690b9a9a8f7e4317759210f8f54f4e

PR #238, opened 2021-08-31
Return `missing` if the field is set but null.

### LINKED ISSUE #177: Handle NULL values in features correctly

yeesian:
Fixes #177 .

I think this aligns the behavior a lot better across the different file formats (see the test sets in test/test_tables.jl)

This behavior is implemented through `getfield(feature, i)` which makes use of `getfieldtype(feature, i)`. That way, we are aligned in behavior for field subtypes even for e.g. displaying the fields.

Because a lot hinges on distinguishing between whether a field is set versus whether it is null (see https://github.com/Toblerity/Fiona/issues/460#issue-240013079), I have also added support for `isfieldnull()`, `isfieldsetandnotnull()`, and `setfieldnull!()`.

### DIFF
--- a/src/ogr/feature.jl
+++ b/src/ogr/feature.jl
@@ -149,6 +149,50 @@ function unsetfield!(feature::Feature, i::Integer)::Feature
     return feature
 end
 
+"""
+    isfieldnull(feature::Feature, i::Integer)
+
+Test if a field is null.
+
+### Parameters
+* `feature`: the feature that owned the field.
+* `i`: the field to test, from 0 to GetFieldCount()-1.
+
+### Returns
+`true` if the field is null, otherwise `false`.
+"""
+isfieldnull(feature::Feature, i::Integer)::Bool =
+    Bool(GDAL.ogr_f_isfieldnull(feature.ptr, i))
+
+"""
+    isfieldsetandnotnull(feature::Feature, i::Integer)
+
+Test if a field is set and not null.
+
+### Parameters
+* `feature`: the feature that owned the field.
+* `i`: the field to test, from 0 to GetFieldCount()-1.
+
+### Returns
+`true` if the field is null, otherwise `false`.
+"""
+isfieldsetandnotnull(feature::Feature, i::Integer)::Bool =
+    Bool(GDAL.ogr_f_isfieldsetandnotnull(feature.ptr, i))
+
+"""
+    setfieldnull!(feature::Feature, i::Integer)
+
+Clear a field, marking it as null.
+
+### Parameters
+* `feature`: the feature that owned the field.
+* `i`: the field to set to null, from 0 to GetFieldCount()-1.
+"""
+function setfieldnull!(feature::Feature, i::Integer)::Feature
+    GDAL.ogr_f_setfieldnull(feature.ptr, i)
+    return feature
+end
+
 # """
 #     OGR_F_GetRawFieldRef(OGRFeatureH hFeat,
 #                          int iField) -> OGRField *
@@ -403,12 +447,14 @@ const _FETCHFIELD = Dict{OGRFieldType,Function}(
 )
 
 function getfield(feature::Feature, i::Integer)
-    return if isfieldset(feature, i)
-        _fieldtype = gettype(getfielddefn(feature, i))
+    return if !isfieldset(feature, i)
+        getdefault(feature, i)
+    elseif isfieldnull(feature, i)
+        missing
+    else
+        _fieldtype = getfieldtype(getfielddefn(feature, i))
         _fetchfield = get(_FETCHFIELD, _fieldtype, getdefault)
         _fetchfield(feature, i)
-    else
-        getdefault(feature, i)
     end
 end
 
--- a/src/base/display.jl
+++ b/src/base/display.jl
@@ -149,7 +149,7 @@ function Base.show(io::IO, layer::AbstractFeatureLayer)::Nothing
     nfielddisplay = min(n, 5)
     for i in 1:nfielddisplay
         fd = getfielddefn(featuredefn, i - 1)
-        display = "     Field $(i - 1) ($(getname(fd))): [$(gettype(fd))]"
+        display = "     Field $(i - 1) ($(getname(fd))): [$(getfieldtype(fd))]"
         if length(display) > 75
             println(io, "$display[1:70]...")
             continue
--- a/src/ogr/fielddefn.jl
+++ b/src/ogr/fielddefn.jl
@@ -72,6 +72,26 @@ function setsubtype!(fielddefn::FieldDefn, subtype::OGRFieldSubType)::FieldDefn
 end
 
 """
+    getfieldtype(fielddefn::AbstractFieldDefn)
+
+Returns the type or subtype (if any) of this field.
+
+### Parameters
+* `fielddefn`: handle to the field definition.
+
+### Returns
+The field type or subtype.
+"""
+function getfieldtype(fielddefn::AbstractFieldDefn)::Union{OGRFieldType, OGRFieldSubType}
+    fieldsubtype = getsubtype(fielddefn)
+    return if fieldsubtype != OFSTNone
+        fieldsubtype
+    else
+        gettype(fielddefn)
+    end
+end
+
+"""
     getjustify(fielddefn::AbstractFieldDefn)
 
 Get the justification for this field.
--- a/test/test_feature.jl
+++ b/test/test_feature.jl
@@ -157,6 +157,10 @@ const AG = ArchGDAL;
                 AG.setfield!(feature, 9, Int16(1))
                 AG.setfield!(feature, 10, Int32(1))
                 AG.setfield!(feature, 11, Float32(1.0))
+                for i in 1:AG.nfield(feature)
+                    @test !AG.isfieldnull(feature, i-1)
+                    @test AG.isfieldsetandnotnull(feature, i-1)
+                end
                 @test sprint(print, AG.getgeom(feature)) == "NULL Geometry"
                 AG.getgeom(feature) do geom
                     @test sprint(print, geom) == "NULL Geometry"
@@ -169,6 +173,10 @@ const AG = ArchGDAL;
                 AG.addfeature(layer) do newfeature
                     AG.setfrom!(newfeature, feature)
                     @test AG.getfield(newfeature, 0) == 1
+                    for i in 1:AG.nfield(newfeature)
+                        @test !AG.isfieldnull(newfeature, i-1)
+                        @test AG.isfieldsetandnotnull(newfeature, i-1)
+                    end
                     @test AG.getfield(newfeature, 1) ≈ 1.0
                     @test AG.getfield(newfeature, 2) == Int32[1, 2]
                     @test AG.getfield(newfeature, 3) == Int64[1, 2]
@@ -209,6 +217,13 @@ const AG = ArchGDAL;
                         @test AG.getfield(newfeature, 0) == 45
                         @test AG.getfield(newfeature, 1) ≈ 18.2
                         @test AG.getfield(newfeature, 5) == String["foo", "bar"]
+
+                        @test AG.isfieldsetandnotnull(newfeature, 5)
+                        AG.setfieldnull!(newfeature, 5)
+                        @test !AG.isfieldsetandnotnull(newfeature, 5)
+                        @test AG.isfieldset(newfeature, 5)
+                        @test AG.isfieldnull(newfeature, 5)
+                        @test ismissing(AG.getfield(newfeature, 5))
                     end
                     @test AG.nfeature(layer) == 1
                 end
--- a/test/test_fielddefn.jl
+++ b/test/test_fielddefn.jl
@@ -10,14 +10,19 @@ const AG = ArchGDAL;
             AG.setname!(fd, "newname")
             @test AG.getname(fd) == "newname"
             @test AG.gettype(fd) == AG.OFTInteger
+            @test AG.getfieldtype(fd) == AG.OFTInteger
             AG.settype!(fd, AG.OFTDate)
             @test AG.gettype(fd) == AG.OFTDate
+            @test AG.getfieldtype(fd) == AG.OFTDate
             AG.settype!(fd, AG.OFTInteger)
             @test AG.getsubtype(fd) == AG.OFSTNone
+            @test AG.getfieldtype(fd) == ArchGDAL.OFTInteger
             AG.setsubtype!(fd, AG.OFSTInt16)
             @test AG.getsubtype(fd) == AG.OFSTInt16
+            @test AG.getfieldtype(fd) == AG.OFSTInt16
             AG.setsubtype!(fd, AG.OFSTBoolean)
             @test AG.getsubtype(fd) == AG.OFSTBoolean
+            @test AG.getfieldtype(fd) == AG.OFSTBoolean
             AG.setsubtype!(fd, AG.OFSTNone)
             @test AG.getjustify(fd) == AG.OJUndefined
             AG.setjustify!(fd, AG.OJLeft)
--- a/test/test_tables.jl
+++ b/test/test_tables.jl
@@ -488,7 +488,7 @@ using Tables
                 withmissingfield::Bool,
                 withmixedgeomtypes::Bool,
                 reference_geotable::NamedTuple,
-            )::Bool
+            )
                 map_on_test_dataset(
                     drvshortname,
                     geomfamilly;
@@ -497,18 +497,14 @@ using Tables
                     withmixedgeomtypes = withmixedgeomtypes,
                 ) do ds
                     layer = AG.getlayer(ds, 0)
-                    return all([
-                        keys(Tables.columntable(layer)) ==
-                        reference_geotable.names,
-                        eltype.(values(Tables.columntable(layer))) ==
-                        reference_geotable.types,
-                        tupleoftuples_equal(
-                            columntablevalues_toWKT(
-                                values(Tables.columntable(layer)),
-                            ),
-                            reference_geotable.values,
+                    @test keys(Tables.columntable(layer)) == reference_geotable.names
+                    @test eltype.(values(Tables.columntable(layer))) == reference_geotable.types
+                    @test tupleoftuples_equal(
+                        columntablevalues_toWKT(
+                            values(Tables.columntable(layer)),
                         ),
-                    ])
+                        reference_geotable.values,
+                    )
                 end
             end
 
@@ -524,24 +520,21 @@ using Tables
             function test_layer_to_table(
                 layer::AG.AbstractFeatureLayer,
                 reference_geotable::NamedTuple,
-            )::Bool
-                return all([
-                    keys(Tables.columntable(layer)) == reference_geotable.names,
-                    eltype.(values(Tables.columntable(layer))) ==
-                    reference_geotable.types,
-                    tupleoftuples_equal(
-                        columntablevalues_toWKT(
-                            values(Tables.columntable(layer)),
-                        ),
-                        reference_geotable.values,
+            )
+                @test keys(Tables.columntable(layer)) == reference_geotable.names
+                @test eltype.(values(Tables.columntable(layer))) == reference_geotable.types
+                @test tupleoftuples_equal(
+                    columntablevalues_toWKT(
+                        values(Tables.columntable(layer)),
                     ),
-                ])
+                    reference_geotable.values,
+                )
             end
 
             @testset "Conversion to table for ESRI Shapefile driver" begin
                 ESRI_Shapefile_test_reference_geotable = (
                     names = (Symbol(""), :id, :name),
-                    types = (Union{Missing,ArchGDAL.IGeometry}, Int64, String),
+                    types = (Union{Missing,ArchGDAL.IGeometry}, Union{Missing,Int64}, String),
                     values = (
                         Union{Missing,String}[
                             "POLYGON ((0 0,0 1,1 1))",
@@ -549,11 +542,11 @@ using Tables
                             missing,
                             "POLYGON ((0 0,-1 0,-1 1))",
                         ],
-                        [1, 2, 3, 0],
+                        [1, 2, 3, missing],
                         ["polygon1", "multipolygon1", "emptygeom", "emptyid"],
                     ),
                 )
-                @test test_layer_to_table(
+                test_layer_to_table(
                     "ESRI Shapefile",
                     "polygon",
                     true,
@@ -569,7 +562,7 @@ using Tables
                             Missing,
                             ArchGDAL.IGeometry{ArchGDAL.wkbLineString},
                         },
-                        Int64,
+                        Union{Missing,Int64},
                         String,
                     ),
                     values = (
@@ -579,11 +572,11 @@ using Tables
                             missing,
                             "LINESTRING (5 6,6 7,7 8)",
                         ],
-                        [1, 2, 3, 0],
+                        [1, 2, 3, missing],
                         ["line1", "line2", "emptygeom", "emptyid"],
                     ),
                 )
-                @test test_layer_to_table(
+                test_layer_to_table(
                     "ESRI Shapefile",
                     "line",
                     true,
@@ -612,7 +605,7 @@ using Tables
                         ["polygon1", "multipolygon1", "emptygeom", "emptyid"],
                     ),
                 )
-                @test test_layer_to_table(
+                test_layer_to_table(
                     "GeoJSON",
                     "polygon",
                     true,
@@ -642,7 +635,7 @@ using Tables
                         ["line1", "line2", "emptygeom", "emptyid"],
                     ),
                 )
-                @test test_layer_to_table(
+                test_layer_to_table(
                     "GeoJSON",
                     "line",
                     true,
@@ -678,7 +671,7 @@ using Tables
                         ["line1", "multiline1", "emptygeom", "emptyid"],
                     ),
                 )
-                @test test_layer_to_table(
+                test_layer_to_table(
                     "GML",
                     "line",
                     true,
@@ -691,7 +684,7 @@ using Tables
             @testset "Conversion to table for GPKG driver" begin
                 GPKG_test_reference_geotable = (
                     names = (:geom, :id, :name),
-                    types = (Union{Missing,ArchGDAL.IGeometry}, Int64, String),
+                    types = (Union{Missing,ArchGDAL.IGeometry}, Union{Missing,Int64}, String),
                     values = (
                         Union{Missing,String}[
                             "LINESTRING (1 2,2 3,3 4)",
@@ -699,11 +692,11 @@ using Tables
                             missing,
                             "LINESTRING (5 6,6 7,7 8)",
                         ],
-                        [1, 2, 3, 0],
+                        Union{Missing,Int64}[1, 2, 3, missing],
                         ["line1", "multiline1", "emptygeom", "emptyid"],
                     ),
                 )
-                @test test_layer_to_table(
+                test_layer_to_table(
                     "GPKG",
                     "line",
                     true,
@@ -727,7 +720,7 @@ using Tables
                         ["", "", ""],
                     ),
                 )
-                @test test_layer_to_table(
+                test_layer_to_table(
                     "KML",
                     "line",
                     true,
@@ -751,7 +744,7 @@ using Tables
                         ["emptyid", "multiline1", "line1"],
                     ),
                 )
-                @test test_layer_to_table(
+                test_layer_to_table(
                     "FlatGeobuf",
                     "line",
                     true,
@@ -762,7 +755,6 @@ using Tables
             end
 
             @testset "Conversion to table for CSV driver" begin
-                @test begin
                     AG.read(
                         joinpath(@__DIR__, "data/multi_geom.csv"),
                         options = [
@@ -797,11 +789,10 @@ using Tables
                                 ["Mumbai", "New Delhi"],
                             ),
                         )
-                        return test_layer_to_table(
+                        test_layer_to_table(
                             multigeom_test_layer,
                             CSV_multigeom_test_reference_geotable,
                         )
-                    end
                 end
             end
 


src/ogr/feature.jl
```diff
@@ -403,12 +447,14 @@ const _FETCHFIELD = Dict{OGRFieldType,Function}(
 )
 
 function getfield(feature::Feature, i::Integer)
-    return if isfieldset(feature, i)
-        _fieldtype = gettype(getfielddefn(feature, i))
+    return if !isfieldset(feature, i)
+        getdefault(feature, i)
```
evetion, line 451: This still seems wrong to me. If the field is not set it is `nothing`. It should not be displayed at all. 

See also https://gdal.org/development/rfc/rfc67_nullfieldvalues.html

However, I can imagine we make it `missing` as well, which would be a bit more user friendly and it is one of the options presented in the above RFC.

evetion (changes requested):
For reference https://gdal.org/development/rfc/rfc67_nullfieldvalues.html

src/ogr/feature.jl
yeesian: That's great feedback, thanks!

Let me go through the following RFCs and functions (and any other corresponding associated functions):
* https://gdal.org/development/rfc/rfc53_ogr_notnull_default.html
  * `getdefault(...)`
  * `setdefault!(...)`
  * `setnullable!(...)`
  * `isdefaultdriverspecific(...)`
  * `validate()`
* https://gdal.org/development/rfc/rfc67_nullfieldvalues.html
  * `getfield(feature, i)`
  * `isfieldnull(feature, i)`
  * `isfieldsetandnotnull(feature, i)`
  * `setfieldnull!(feature, i)`

I will
* provide a corresponding discussion under https://yeesian.com/ArchGDAL.jl/stable/considerations/ as part of this PR,
* update the PR title and so on to reflect the design decision, before
* updating the implementation and tests to reflect it.

yeesian: Here's what I have as an early draft:

```
+-------+---------+---------+
|       |  null   | notnull |
+-------+---------+---------+
| set   | missing | value   |
+-------+---------+---------+
| unset | nothing | nothing |
+-------+---------+---------+
```

The design decision reflects that
* a value can either "not exist" or "be unknown", but not both.
* "set" means that a value "exists" (even if it might be unknown): the corresponding value we use for it is `missing`.
* "null" means that a value is "unknown" (presuming that it exists): the corresponding value we use for it is `nothing`.

evetion: In RFC 67, under [Compatibility](https://gdal.org/development/rfc/rfc67_nullfieldvalues.html#compatibility-issues), it says

>All code, in GDAL source code, and in calling external code, that currently uses OGRFeature::IsFieldSet() / OGR_F_IsFieldSet() should also be updated to used IsFieldNull() / OGR_F_IsFieldNull(), **either to act exactly as in the unset case**, or add a new appropriate behavior.

Which would mean that we could use `missing` also in the unset case as well. 

While I think `nothing` is semantically more correct, I fear that having a `Union{Nothing, Missing, ValueType}` is not pragmatic. We should test this with a few special files with both unset as null fields.

yeesian: > We should test this with a few special files

Let me know If there are any cases you wish to call out or files you'd like to ensure gets tested.

mathieu17g: I would say from my reading of the [RFC 67](https://github.com/OSGeo/gdal/blob/master/gdal/doc/source/development/rfc/rfc67_nullfieldvalues.rst) and the GDAL source files containing nMarker1, that the logic is the following: `OGRUnsetMarker` and `OGRNullMarker`are mutually exclusive.

I propose the following mapping:

**nMarker1/2/3**|**Field has default value ?**|**Returned value**
:-----:|:-----:|:-----:
OGRUnsetMarker|true|default value
OGRUnsetMarker|false|missing
OGRNullMarker|true / false|nothing
other value|true / false|value

yeesian: Thanks Mathieu! What do you think of the following proposal?

```
+-------+---------+--------------+
|       |  null   |   notnull    |
+-------+---------+--------------+
| set   | missing |    value     |
+-------+---------+--------------+
| unset | nothing | getdefault() |
+-------+---------+--------------+
```
where `getdefault()` should return `missing` if there is no default value.

I think that way,
* "NULL" has flavours of `missing` and `nothing` depending on whether the field is set.
* "NOTNULL" is consistent with "Note that adding a field with a NOT NULL constraint on a non-empty layer is generally impossible, unless a DEFAULT value is associated with it" and "Fields with NOT NULL constraints are sometimes accompanied with a DEFAULT clause so as to be able to create a new feature without filling all fields, while maintaining integrity" in https://gdal.org/development/rfc/rfc53_ogr_notnull_default.html.
* Fields that are set can propagate `missing` value behavior (rather than throwing an error on `nothing`). Users can use [`isfieldset()` and `unsetfield!(...)`](https://github.com/yeesian/ArchGDAL.jl/blob/d177575acd70775434f4ace21b9313573823469c/src/ogr/feature.jl#L126-L150) to check for this behavior and control it to return `nothing` (rather than `missing`) if they want.
* Fields that might be unset have a `FillUnsetWithDefault()` method in GDAL (that should be rarely used; per https://gdal.org/development/rfc/rfc53_ogr_notnull_default.html).

> I would say from my reading of the RFC 67 and the GDAL source files containing nMarker1, that the logic is the following: OGRUnsetMarker and OGRNullMarkerare mutually exclusive.

It's not obvious to me that `OGRUnsetMarker ` and `OGRNullMarkerare` are mutually exclusive -- can you explain how you came to the conclusion?

yeesian: @evetion I'm afraid the proposal is a complex one, but it is not trying to be more complex than it needs to be. From what I've read of the GDAL RFCs, I think that `missing`, `nothing` or `getdefault()` can be sensible behavior depending on the circumstance.

evetion: From my reading of the RFCs, `getdefault()` is only used for **insertion**, when a field can't be null, but isn't given. This code is about reading, so it shouldn't come into play at all.

So your original proposal seems right:
| read field | null    | notnull |
|------------|---------|---------|
| set        | missing | value   |
| unset      | nothing | nothing |

mathieu17g: @yeesian 

> > I would say from my reading of the RFC 67 and the GDAL source files containing nMarker1, that the logic is the following: OGRUnsetMarker and OGRNullMarke rare mutually exclusive.
> 
> It's not obvious to me that OGRUnsetMarker and OGRNullMarkerare are mutually exclusive -- can you explain how you came to the conclusion?

Well, my understanding of OGR cpp code may be superficial (1st time I looked into it), but here it is for comments:

1. An `ÒGRField` is defined by an [Union](https://github.com/OSGeo/gdal/blob/814434983f73c2917ceb36ee8a72fa3225ce646c/gdal/ogr/ogr_core.h#L730-L788)
2. In this Union, there is as [`Set`](https://github.com/OSGeo/gdal/blob/814434983f73c2917ceb36ee8a72fa3225ce646c/gdal/ogr/ogr_core.h#L770-L774) struct containing the three markers `nMarker1`, `nMarker2`, `nMarker3`
3. If one of the members (other than the `Set` struc member) of an `OGRField` is set, `nmarker2` and `nmarker3` are set to `0`
When `nMarker1` is set, `nMarker2` and `nMarker3` are set equal to `nMarker1`. 
`nMarker1` is set to [two possible values](https://github.com/OSGeo/gdal/blob/814434983f73c2917ceb36ee8a72fa3225ce646c/gdal/ogr/ogr_core.h#L715-L728) by two functions:
    - `OGRUnsetMarker` by [`OGR_RawField_SetUnset`](https://github.com/OSGeo/gdal/blob/b4c3e127edda6cbb0c8e3f37d287e624c7d84395/gdal/ogr/ogrfeature.cpp#L6968-L6973)
    - `OGRNullMarker`by [`OGR_RawField_SetNull`](https://github.com/OSGeo/gdal/blob/b4c3e127edda6cbb0c8e3f37d287e624c7d84395/gdal/ogr/ogrfeature.cpp#L6992-L6997)

Therefore since Unset state and Null state for an OGRField are implemented by two different values of one unique variable, the two states are mutually exclusive

evetion: @mathieu17g Did you switch the meanings of `missing` and `nothing` by accident? IMHO they should be switched. The Julia manual states: `missing is equivalent to NULL in SQL`, while `[nothing is] used by convention when there is no value to return`.

Then you'd get.

Read Field | null | notnull
-- | -- | --
set | missing | value
unset | n/a | nothing


Write Field | nullable | notnullable
-- | -- | --
nothing | unset | unset
missing | null | default value
value | value | value

mathieu17g: @evetion: Well my proposal was intended but mostly based on intuition.
Following your remark, I have dug a bit more in Julia's documentation about missing and nothing values and SQL null usage, I have found some arguments to support my convention proposition:
- `OGRField.Set.nMarker1 = OGRUnsetMarker`  to Julia `missing` and 
- `OGRField.Set.nMarker1 = OGRNullMarker` to Julia `nothing`

But none of these arguments is definitive and one may find some other arguments supporting the opposite convention. So since, we have to decide to be able to move on, **both conventions proposals are fine for me**. And if later, we encounter a similar convention issue in another Julia package with convincing argument, we will be able to easily change the convention in ArchGDAL.

---

If you have spare time, here are below, a few additional observations based on vague pieces of docs found on the topic in Julia's documentation

### In Julia's [documentation page on missing values](https://docs.julialang.org/en/v1/manual/missing/#missing):

> Julia provides support for representing missing values in the statistical sense, that is for situations where no value is available for a variable in an observation, but a valid value theoretically exists. Missing values are represented via the missing object, which is the singleton instance of the type Missing. **missing is equivalent to NULL in SQL and NA in R**, and behaves like them in most situations.

One interpretation could be:
- NULL in SQL is associated with no value provided for a field when the record was written. So when one does not set a value in a record, it is always set to NULL
- The equivalent in OGR is unset since a feature fields are initiated with the OGRField set to OGRUnsetMarker in the [basic constructor](https://github.com/OSGeo/gdal/blob/b4c3e127edda6cbb0c8e3f37d287e624c7d84395/gdal/ogr/ogrfeature.cpp#L99-L107).
- The `OGRNullMarker` is never set by default in OGR, it will be set on purpose with the function `SetFieldNull` on a particular OGRField instance

**So my understanding of this part of Julia documentation is that it could also be in favor of associating `OGRField.Set.nMarker1 = OGRUnsetMarker`  to Julia `missing`**

### In Julia's documentation page on Function and in the paragraph ["Returning nothing"](https://docs.julialang.org/en/v1/manual/functions/#Returning-nothing)

> For **functions that do not need to return a value** (functions used only for some side effects), the Julia convention is to return the value nothing

Here with `ArchGDAL.getfield` the main purpose is not any side effect, but to return a supposedly existing value of the field

I have not found your reference:

> [nothing is] used by convention when there is no value to return

### In Julia's documents FAQ's paragraph ["Nothingness and missing values"](https://docs.julialang.org/en/v1/manual/faq/#Nothingness-and-missing-values)

> For situations where a value `x` of type `T` exists only sometimes, the `Union{T, Nothing}` type can be used for function arguments, object fields and array element types as the equivalent of [Nullable](https://en.wikipedia.org/wiki/Nullable_type), [Option](https://en.wikipedia.org/wiki/Nullable_type) or [Maybe](https://en.wikipedia.org/wiki/Nullable_type) in other languages. 

When I follow the link to the wikipedia page, the explinations given there are confusing since they never refer to existence of something similar to Julia's dual "missingness" and "nothingness"
The rest of the paragraph does not allow to  decide.

yeesian: It seems like we're getting close to consensus here. @evetion, will you be okay if I update my original proposal in https://github.com/yeesian/ArchGDAL.jl/pull/238#discussion_r706615531 and yours in https://github.com/yeesian/ArchGDAL.jl/pull/238#discussion_r707369926 to the following (changes are marked with an asterisk):

```
+-------+---------+----------+
| Read  |  null   | notnull  |
+-------+---------+----------+
| set   | missing |  value   |
+-------+---------+----------+
| unset | nothing | missing* |
+-------+---------+----------+

+---------+-------------+--------------+
|  Write  |  nullable   | notnullable  |
+---------+-------------+--------------+
| nothing |    unset    |     unset    |
+---------+-------------+--------------+
| missing |     null    | getdefault() |
+---------+-------------+--------------+
|  value  |    value    |    value     |
+---------+-------------+--------------+

```

Namely, we
1. There is no change to your proposal when writing fields.
2. When reading fields, we return `missing` instead of `nothing` for fields that are `(unset, notnull)` case.

I'll start reflecting this discussion and decision in the documents, so there's still time to make changes to intended behavior.

evetion: @yeesian I'm fine with your proposed change, it seems more pragmatic to me. @mathieu17g Thanks for all the thought you put into the discussion, it has made things much better and more explicit.

yeesian: @mathieu17g you have saved me from potential hours of head-scratching with https://github.com/yeesian/ArchGDAL.jl/pull/238#discussion_r707040320: they are mutually exclusive and I have not been able to construct any cases of fields that are null and unset, thank you!

mathieu17g: You’re welcome :-)

yeesian: Hey mathieu, the latest implementation in this PR (see https://github.com/yeesian/ArchGDAL.jl/pull/238/commits/958496206e31085aba53cb3ce5cb83aa09160a73) is based on https://github.com/yeesian/ArchGDAL.jl/pull/238#discussion_r708786099 and (i) assumes that the default value is only for writing fields and not when reading fields, (ii) returns `missing` rather than `nothing` for unset values. So I think it's the following table:

```
+-----------------+
| Read            | 
|-----------------|-----------+
| set and notnull |   value   |
|-----------------+-----------+
| everything else | `missing` |
|-----------------+-----------+
```

If you might feel strongly about returning `nothing` (rather than `missing`) for the `unset` case, I'm happy to revisit the implementation. Otherwise, my current thinking is to reserve `nothing` for the impossible case where we read a field that is both `unset` and `null`.

mathieu17g: > If you might feel strongly about returning nothing (rather than missing) for the unset case, I'm happy to revisit the implementation. Otherwise, my current thinking is to reserve nothing for the impossible case where we read a field that is both unset and null.

I think that:
- reading a field that is both **UNSET** and **NULL**, is not possible in GDAL's current implementation since those states correspond to different values of the same `Set struct` variable of an `OGRField` instance 
- we would loose the symmetry between reading and writing as, in my PR #243, I was about to map `nothing` to **NULL** and `missing` to **UNSET**, to give the ability to set either of those two states from a table source

But, if your are not convinced, no issue, I will map both `nothing` and `missing` to `UNSET`

yeesian: Thanks for re-kindling the debate @mathieu17g, and sorry it took me this long to think through it deeper: I agree with your reasoning and have reverted the code to the behavior described in https://github.com/yeesian/ArchGDAL.jl/pull/238#discussion_r699802843, namely,

```
+-------+---------+---------+
| Read |  null   | notnull |
+-------+---------+---------+
| set   | missing | value   |
+-------+---------+---------+
| unset | nothing | nothing |
+-------+---------+---------+

+---------+-------------+--------------+
|  Write  |  nullable   | notnullable  |
+---------+-------------+--------------+
| nothing |    unset    |     unset    |
+---------+-------------+--------------+
| missing |     null    | getdefault() |
+---------+-------------+--------------+
|  value  |    value    |    value     |
+---------+-------------+--------------+
```

where `getdefault()` returns `nothing` if there is no default value (which would cause it to be "unset" for notnullable columns).

---

That way, we
1. have a sufficiently rich set of Julian values (using both `missing` and `nothing`) to be able to reconstruct the state of a FeatureLayer (with the possible states of "unset" and "null").
2. support both (i) users who want to "fail early": use notnullable fields and error on `nothing`, and (ii) users who want to continue data processing with missing values: use nullable fields and propagate `missing` values.

yeesian: @mathieu17g @evetion I have updated the PR with https://github.com/yeesian/ArchGDAL.jl/pull/238/commits/08ffc271a64de12f352d12cb0eb4de3426ef1518 and documented it in https://yeesian.com/ArchGDAL.jl/previews/PR238/considerations/#Missing-and-Null-Semantics-in-GDAL -- thank you both so much! Can you have a look and let me know if this thread is good to be resolved?

mathieu17g (commented):
I have made some suggestions on the src part and put on hold the test part's review, waiting for your feedback

src/ogr/feature.jl
```diff
@@ -439,6 +504,20 @@ field types may be unaffected.
...
+end
+
+function setfield!(feature::Feature, i::Integer, value::Missing)::Feature
+    if isnullable(getfielddefn(feature, i))
+        setfieldnull!(feature, 1)
+    else
+        setfield!(feature, i, getdefault(feature, i))
+    end
+    return feature
+end
```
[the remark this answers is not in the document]
yeesian: > My understanding of nullable case handling in GDAL (based on RFC 57), is that it is delegated to the low level driver.

Unfortunately that doesn't seem to always be the case: e.g. using the recommendation will cause the following test to fail: https://github.com/yeesian/ArchGDAL.jl/blob/08ffc271a64de12f352d12cb0eb4de3426ef1518/test/test_feature.jl#L159

src/ogr/feature.jl
```diff
@@ -402,13 +455,25 @@ const _FETCHFIELD = Dict{OGRFieldType,Function}(
...
+        _fieldtype = getfieldtype(getfielddefn(feature, i))
         _fetchfield = get(_FETCHFIELD, _fieldtype, getdefault)
         _fetchfield(feature, i)
+    elseif isfieldset(feature, i)
+        missing
     else
-        getdefault(feature, i)
+        nothing
     end
 end
```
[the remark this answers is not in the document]
yeesian: > Since we want to distinguish unset and null state, shouldn't we use isfieldset and isfieldnull instead ?

Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/238/commits/323cbdb7719a1ad62affec6d0462fd0de7d18d7b.

> And since, we should not have any unknown fieldtype, shouldn't we raise an error in case of unknown type instead

Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/238/commits/5c7b677e5350fe581a1e5ec9559d9ac91d302908.

docs/src/considerations.md
```diff
@@ -96,3 +96,34 @@ The interface is implemented in [`src/tables.jl`](https://github.com/yeesian/Arc
...
+
+This reflects that
+* writing `nothing` will cause the field to be unset.
+* writing `missing` will cause the field to be null. In the cause of a notnullable field, it will take the default value (see https://gdal.org/development/rfc/rfc53_ogr_notnull_default.html for details). If there is no default value, `getdefault()` will return `nothing`, causing the field to be unset.
+* writing a value will behave in the usual manner.
+
+For additional references, see
+* https://docs.julialang.org/en/v1/manual/faq/#faq-nothing
+* https://gdal.org/development/rfc/rfc53_ogr_notnull_default.html
+* https://gdal.org/development/rfc/rfc67_nullfieldvalues.html
```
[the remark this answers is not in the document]
yeesian: It is an oversight, thank you! Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/238/commits/364f171a390f1073810e0cef3f2f0852e4be7e34.

src/ogr/feature.jl
```diff
@@ -149,6 +149,59 @@ function unsetfield!(feature::Feature, i::Integer)::Feature
...
+    isfieldsetandnotnull(feature::Feature, i::Integer)
+
+Test if a field is set and not null.
+
+### Parameters
+* `feature`: the feature that owned the field.
+* `i`: the field to test, from 0 to GetFieldCount()-1.
+
+### Returns
+`true` if the field is null, otherwise `false`.
```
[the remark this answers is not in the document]
yeesian: Fixed in https://github.com/yeesian/ArchGDAL.jl/pull/238/commits/092bed4a4ddaf4d08696ddbf4c2d655f1e0b24c5, thanks!

mathieu17g (approved):

visr:
Nice work here! I wasn't aware of unset vs null in GDAL before. I believe null is much more common in practice?

I made this GeoJSONSeq file with both unset and null values:

```
{ "type": "Feature", "properties": { "a": 1, "b": null }, "geometry": { "type": "Point", "coordinates": [ 1.0, 1.0 ] } }
{ "type": "Feature", "properties": {}, "geometry": { "type": "Point", "coordinates": [ 2.0, 2.0 ] } } 
```

If I run this through ogr with `ogr2ogr -f GeoJSONSeq unset-gdal-translate.geojson unset.geojson`, the result is the same, e.g. unset and null are both kept. If I import the file in QGIS however, it shows this:

image

And indeed if I export it from QGIS I get nulls back in field that were unset before:

```
{ "type": "Feature", "properties": { "a": 1, "b": null }, "geometry": { "type": "Point", "coordinates": [ 1.0, 1.0 ] } }
{ "type": "Feature", "properties": { "a": null, "b": null }, "geometry": { "type": "Point", "coordinates": [ 2.0, 2.0 ] } } 
```

Probably it makes more sense to keep the distinction here just like `ogr2ogr` does, this being a GDAL wrapper. Though I do understand @evetion's point about making everything missing possibly being more user friendly.

evetion:
@visr This is all about displaying a non-tabular (sparse) dataset like your geojson in a tabular interface. In this case, QGIS doesn't have a representation of unset, just NULL. Julia can be more expressive, although I wonder whether our Tables interface likes the mixing of `missing` and `nothing`.

visr:
Yeah exactly. I can imagine it being less ideal in the Tables interface. Though these functions are also used outside of that, in the sparse setting, where it does make sense to keep the distinction.

I think the interface can technically handle both just fine, it is just not very conventional. Though I don't know how common unset is in practice, and it's relatively easy to convert `nothing` to `missing`, if it causes problems.

src/ogr/feature.jl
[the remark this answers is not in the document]
mathieu17g: My mistake.

Furthermore I had not understood that `setfield!(::Feature, ::Integer, ::AbstractString)` was calling `GDAL.ogr_f_setfieldstring`, calling itself GDAL `OGRFeature::SetField` which takes care of the conversion from a string pointer to the `FieldDefn` type

mathieu17g:
@evetion and @visr, in the draft PR #243 , I'm about to try to handle it for a conversion from a table source to an `IFeatureLayer`.  
I was just waiting for this PR #238  to be merged.

In the other way, since we rely on [Tables.jl/src/fallbacks.jl](https://github.com/JuliaData/Tables.jl/blob/51a6ad4f4a8305b1d79ea1b8f607e4d0f13f2ebc/src/fallbacks.jl) `buildcolumns(::Nothing, rowitr::T) where {T}` which iterates through all features building column types by "widening" them with `promote_type` each time it encounters a new type in field values, it should work as follow if we have an `FieldDefn` of `OGRFieldType` `OFTInteger64`  and features with UNSET or NULL fields for this `FieldDefn`:
```julia
julia> T = Int64; @show T
       T = promote_type(T, typeof(nothing)); @show T
       T = promote_type(T, typeof(missing)); @show T
T = Int64
T = Union{Nothing, Int64}
T = Union{Missing, Nothing, Int64}
```

visr:
@mathieu17g thanks for showing that indeed technically it works (#243 is a great addition by the way).

I think for the tables interface specifically, the concerns about `nothing` are:
- In SQL / JuliaData / Tables.jl packages generally only use `missing` and not `nothing`, potentially leading to confusion or errors if our tables are used in functions that only expect `missing` or data.
- `Union{Missing, Nothing, Int64}` may not enable small union optimizations that `Union{Missing, Int64}` will have (would need to be checked).

Hence for tables I'm less enthousiastic about `nothing` than for just `getfield`. Though I'm not sure what's best either. If we go with `nothing` at worst we force people to explicitly convert to `missing`, which is not too bad of a price to pay to make the data perfectly round trippable and consistent with GDAL itself. Though if you know a way to optionally do the `nothing` to `missing` conversion during table construction instead of afterwards, potentially avoiding an expensive copy, it would be interesting as well.

yeesian:
@evetion @visr is this PR good to go from your perspective? I'm okay with iterating further on feedback or for you to have more time to review if you'd like -- else I'll merge it to keep things moving for Mathieu per https://github.com/yeesian/ArchGDAL.jl/pull/238#issuecomment-927720530

mathieu17g:
> Hence for tables I'm less enthousiastic about `nothing` than for just `getfield`. Though I'm not sure what's best either. If we go with `nothing` at worst we force people to explicitly convert to `missing`, which is not too bad of a price to pay to make the data perfectly round trippable and consistent with GDAL itself. 

Yes, and if we need to convert all `nothing` values to `missing` later, it won't be difficult

>Though if you know a way to optionally do the `nothing` to `missing` conversion during table construction instead of afterwards, potentially avoiding an expensive copy, it would be interesting as well.

This should be easily done in [`getcolumn`](https://github.com/yeesian/ArchGDAL.jl/blob/8f820c77c790d6166f03f17f145a4ffe474933de/src/tables.jl#L12-L32) function which is called for each field and geometry value by Tables.jl/src/fallback.jl `Tables.columns` function
For the `Tables.rows`, it can be done by implementing a slightly modified version of `Tables.rows` fallback function
```julia
function Tables.rows(x::AbstractFeatureLayer)
    cols = Tables.columns(x)
    return Tables.RowIterator(cols, Int(Tables.rowcount(cols)))
end
``` 

> potentially avoiding an expensive copy

I didn't get what you have in mind. `Tables.columns` fallback constructor always make a copy.  
I had not considered the option of not making a copy. Maybe it would allow to modify a dataset from a table (e.g. DataFrame). Brilliant idea!

docs/src/considerations.md
```diff
@@ -96,3 +96,34 @@ The interface is implemented in [`src/tables.jl`](https://github.com/yeesian/Arc
...
 * `ArchGDAL.AbstractDataset` might contain multiple layers, and might correspond to multiple tables. The way to construct tables would be to get the layers before forming the corresponding tables.
 
+## Missing and Null Semantics in GDAL
+
+When reading the fields of a feature using `getfield(feature, i)`, ArchGDAL observes the following behavior:
+
+| Field | null    | notnull |
+|-------|---------|---------|
+| set   | missing | value   |
+| unset | nothing | nothing |
```
visr, line 106: Shouldn't the unset/null case be n/a? Because it cannot happen.

src/ogr/feature.jl
```diff
@@ -402,13 +455,33 @@ const _FETCHFIELD = Dict{OGRFieldType,Function}(
...
+        missing
     else
-        getdefault(feature, i)
+        _fieldtype = getfieldtype(getfielddefn(feature, i))
+        try
+            _fetchfield = _FETCHFIELD[_fieldtype]
+            _fetchfield(feature, i)
+        catch e
+            if e isa KeyError
+                error("$_fieldtype not implemented in getfield, please report an issue to https://github.com/yeesian/ArchGDAL.jl/issues")
```
visr, line 480: This is helpful message, but I wonder if a try-catch in such an often called method might slow it down? The error without try catch will be worse, but not terrible.

visr (approved):
I'm fine with merging this. Left two comments.

visr:
@mathieu17g we can discuss copies and such in a different issue. I still wanted to create one, because I saw that converting a layer to DataFrame for a large vector file took minutes.

docs/src/considerations.md
yeesian: > Because it cannot happen.

In the (impossible) event that it does, do we prefer receiving a `nothing` or having an error being raised?

src/ogr/feature.jl
yeesian: Yeah I was thinking of [turning the enums into types so that we can dispatch on them](https://discourse.julialang.org/t/style-recommendation-for-enum-as-type/46078/5), but that one commit grew and grew in scope, so I ended up keeping it out of this PR, with the rationale that performance optimizations can be pursued down the line once we've nailed the API.

docs/src/considerations.md
visr: Hmm I see that since we have `if !isfieldset(feature, i) return nothing` in the code that it would return `nothing`. But since we know it is not possible, putting `nothing` could be confusing to people, that may wrongly believe they have to check for this condition. Hence I think n/a is better. For the same reason it's not worth checking for this condition in `getfield` to raise an error.

src/ogr/feature.jl
visr: Ok indeed it would only be a performance optimization, which could be done later.

docs/src/considerations.md
yeesian: > But since we know it is not possible, putting `nothing` could be confusing to people, that may wrongly believe they have to check for this condition.

That makes sense, reflected in https://github.com/yeesian/ArchGDAL.jl/pull/238/commits/a7710c24a4101e10f79d79cc01ef6ae8fb09623e, thanks!

src/ogr/feature.jl
yeesian: Resolving this thread since it's run its course. Feel free to open new comments / issues if there's anything unresolved.

src/ogr/feature.jl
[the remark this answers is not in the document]
yeesian: I think this thread is resolved. Feel free to open a new comment or issue if that's not the case.

src/ogr/feature.jl
yeesian: Opened https://github.com/yeesian/ArchGDAL.jl/issues/246 to track it.

yeesian:
> @mathieu17g we can discuss copies and such in a different issue. I still wanted to create one, because I saw that converting a layer to DataFrame for a large vector file took minutes.

I'm interested in this too! Feel free to open an issue for it or have the discussion in #243

mathieu17g (approved):
Thorough! :-)

yeesian:
I'll be merging this PR after #245 is merged if I don't see any further objections by then.

PR #240, opened 2021-09-01
Handle non-bijective data-type conversions more robustly

### LINKED ISSUE #232: Handle non-bijective data-type conversions more robustly
Given that we're using dictionaries for the code to convert between data types, there's no good way of [asserting "default type comes last" for mappings that are not bijective](https://github.com/yeesian/ArchGDAL.jl/blob/f4406a55c5e2d42e0617e5e686d87f71d376d4be/src/types.jl#L305-L310).

Opening this ticket to
* Include it in the docstring for the `convert` macro.
* Add unit tests for them.
* Rewrite the implementation to [enforce it by construction](https://discourse.julialang.org/t/initialising-a-dictionary-with-duplicate-keys/2290/2).

mathieu17g:
Fixes #232
Assert "default type comes last" for mappings that are not bijective
- [x] Include it in the docstring for the convert macro
- [x] Add unit tests for them.
- [x] Rewrite the implementation to enforce it by construction

I have enriched the `@convert` macro's docstring with some explanations and added a note on "default type comes last"

---
Before going further, @yeesian could you elaborate on:
>  asserting "default type comes last" for mappings that are not bijective.

Should we enforce it with something like:
1) enhance the readability of the macro call
```julia
@convert(
    OGRFieldType::DataType,
    OFTInteger::(Bool, Int16, Int32, ) # default type comes last
    OFTInteger64::Int64,
)
```
and assert that all type ids of type 1 and all type ids / subtypes of type 2 are unique?

2) Or is it enough to assert that all type id <-> type id / subtype mappings are unique in `args`? 
3) Or something else?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c9ce4bd3-c3d5-55b8-8973-c0e20141b8c3"
 keywords = ["GDAL", "IO"]
 license = "MIT"
 desc = "A high level API for GDAL - Geospatial Data Abstraction Library"
-version = "0.7.3"
+version = "0.7.4"
 
 [deps]
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ desc = "A high level API for GDAL - Geospatial Data Abstraction Library"
 version = "0.7.4"
 
 [deps]
+CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 DiskArrays = "3c3547ce-8d99-4f5e-a174-61eb10b00ae3"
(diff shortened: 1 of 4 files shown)


yeesian:
I think we don't need to enhance the readability of the macro call, since the current one also works in the reverse direction.

I was thinking more about writing https://github.com/yeesian/ArchGDAL.jl/blob/18578bb0d27b3aa0db707c50dfb097aeb69f8765/src/utils.jl#L60-L75 as a loop as suggested in https://discourse.julialang.org/t/initialising-a-dictionary-with-duplicate-keys/2290/2. That way, the behavior is deterministic rather than relying on the current implementation of `Dict` (which could change).

mathieu17g (commented):
Tests fails in Julia 1.3 because there were no ImmutableDict method taking a list of Pair arguments. It has been [fixed](https://github.com/JuliaLang/julia/pull/36143#issue-427808934) in Julia 1.6 and back ported to Julia 1.5

I will add a copy of the fix in utils.jl for Julia versions < v"1.5"

mathieu17g (commented):
Still does not work, I will fallback to the initial helper constructor based on a loop for Julia version `< v"1.5"`

yeesian:
This is great! Thank you so much (as always)!

Issue #215, opened 2021-07-05
Model and handle relationships between GDAL objects systematically


yeesian:
See e.g. https://github.com/yeesian/ArchGDAL.jl/pull/214#issuecomment-873856342 for context.

Issue #152, opened 2020-11-09
Construct ArchGDAL.Table via named tuples or DataFrame


zerefwayne:
I have a use case of filtering rows out of an ArchGDAL table and then return ArchGDAL datatype. It would be helpful to construct an ArchGDAL.Table using an array of named tuples or even a DataFrame from DataFrames.jl.

cc: @juliohm

juliohm:
@zerefwayne I suggest starting a draft PR in GADM.jl and linking to this issue. It is hard to know exactly what you are trying to achieve otherwise. :+1:

yeesian:
We currently support https://tables.juliadata.org/stable/#Implementing-the-Interface-(i.e.-becoming-a-Tables.jl-source)-1 in [`src/tables.jl`](https://github.com/yeesian/ArchGDAL.jl/blob/master/src/tables.jl).

It will be nice to support the creation of an ArchGDAL.Dataset (or FeatureLayer) from a corresponding table via https://tables.juliadata.org/stable/#Using-the-Interface-(i.e.-consuming-Tables.jl-compatible-sources)-1: here's an [example](https://github.com/yeesian/ArchGDAL.jl/blob/9cc8713063a537f10cc2801907073f7cfd30c1c4/test/test_ospy_examples.jl#L92-L119) of how it can be done.

evetion:
Shameless plug (I'm the author), I've made a first implementation over at https://github.com/evetion/GeoDataFrames.jl.

mathieu17g:
@yeesian (cc: @evetion) I'm could work on this by creating IFeatureLayer constructor for table types (creating an associated a "Memory" driver Dataset).

Writing to a file, could then be handled by a write, leaving GDAL handling the driver restrictions as pointed in https://github.com/yeesian/ArchGDAL.jl/issues/147#issuecomment-714186909

yeesian:
That'll be really awesome @mathieu17g! I also agree that the analogous GDAL object is a feature layer and have updated the title to reflect it.

PR #243, opened 2021-09-08
Table to IFeatureLayer

mathieu17g:
Here is a 1st draft to fix issue #152 

I have implemented the conversion from table to an `IFeatureLayer` in a memory dataset.
It is done via an `IFeatureLayer` constructor but it can be moved to a `createlayer` method.

Here are some modifications that has / may need to be implemented:
- [ ] add tests
- [ ] (MAYBE) in column type conversion, move from try-catch with convert to if-else with collections (to be defined)
- [ ] handle the case where `names === nothing` or `types === nothing` in the `Table.Schema`
- [ ] once PR #238 is merged, set field/geomfield values `=== nothing` to null

Other additions:
- added `convert` methods between `OGRwkbGeometryType` and `IGeometry{OGRwkbGeometryType}`. Necessary in column types conversion since it is based on try-catch on `convert` methods
- used `GDAL.ogr_gfld_setname` to set first geomfield name, since `setname!` is not implemented for `FeatureDefn`. set name!(::FeatureDefn, ::String) may require to fix issue #215 first 
- fixed `@ogrerr`: message has to be escaped. This fix may also be necessary in other similar macros (not tested)
- added `convert` methods between `OGRFieldType` and `DataType` for vectors, `Int8` and `Float16`

Here is a test on multi_geom.csv test file:
```julia
ulia> ds = AG.read("multi_geom.csv",
           options = [
               "GEOM_POSSIBLE_NAMES=point,linestring",
               "KEEP_GEOM_COLUMNS=NO",
           ],
       )
       layer = AG.getlayer(ds, 0)
       @show layer
       df = DataFrame(layer)
       @show df; println()
       @show describe(df); println()
       new_layer = AG.IFeatureLayer(df)
       @show new_layer
       new_df = DataFrame(new_layer)
       @show new_df; println()
       @show describe(new_df); println()
layer = Layer: multi_geom
  Geometry 0 (point): [wkbUnknown]
  Geometry 1 (linestring): [wkbUnknown]
     Field 0 (id): [OFTString], 5.1, 5.2
     Field 1 (zoom): [OFTString], 1.0, 2.0
     Field 2 (location): [OFTString], Mumbai, New Delhi

df = 2×5 DataFrame
 Row │ point               linestring               id      zoom    location
     │ IGeometr…           IGeometry…               String  String  String
   1 │ Geometry: wkbPoint  Geometry: wkbLineString  5.1     1.0     Mumbai
   2 │ Geometry: wkbPoint  Geometry: wkbLineString  5.2     2.0     New Delhi

describe(df) = 5×7 DataFrame
 Row │ variable    mean     min     median   max        nmissing  eltype
     │ Symbol      Nothing  Union…  Nothing  Union…     Int64     DataType
─────┼─────────────────────────────────────────────────────────────────────────────────────
   1 │ point                                                   0  IGeometry{wkbPoint}
   2 │ linestring                                              0  IGeometry{wkbLineString}
   3 │ id                   5.1              5.2               0  String
   4 │ zoom                 1.0              2.0               0  String
   5 │ location             Mumbai           New Delhi         0  String

new_layer = Layer: 
  Geometry 0 (point): [wkbPoint]
  Geometry 1 (linestring): [wkbLineString]
     Field 0 (id): [OFTString], 5.1, 5.2
     Field 1 (zoom): [OFTString], 1.0, 2.0
     Field 2 (location): [OFTString], Mumbai, New Delhi

new_df = 2×5 DataFrame
 Row │ point               linestring               id      zoom    location
     │ IGeometr…           IGeometry…               String  String  String
   1 │ Geometry: wkbPoint  Geometry: wkbLineString  5.1     1.0     Mumbai
   2 │ Geometry: wkbPoint  Geometry: wkbLineString  5.2     2.0     New Delhi

describe(new_df) = 5×7 DataFrame
 Row │ variable    mean     min     median   max        nmissing  eltype
     │ Symbol      Nothing  Union…  Nothing  Union…     Int64     DataType
─────┼─────────────────────────────────────────────────────────────────────────────────────
   1 │ point                                                   0  IGeometry{wkbPoint}
   2 │ linestring                                              0  IGeometry{wkbLineString}
   3 │ id                   5.1              5.2               0  String
   4 │ zoom                 1.0              2.0               0  String
   5 │ location             Mumbai           New Delhi         0  String

```


mathieu17g:
OK as PR #245, I will close this PR and open a new one.
Currently, I have rebase to abort locally that blocks me. I'm digging into it

mathieu17g:
I have added a 1st draft for GeoInterface geometries handling

```julia
julia> using ArchGDAL; const AG = ArchGDAL
       using LibGEOS
       using DataFrames

julia> df = DataFrame(AG.getlayer(AG.read("tmp/road/road.shp"), 0))
       df = transform(
           df, 
           Symbol("") => 
           (x -> passmissing(LibGEOS.readgeom ∘ ArchGDAL.toWKT).(x)) => 
           Symbol("")
       )
       df[1:5, :]
5×8 DataFrame
 Row │                                    gid    roadcode     plod    absd     plof    absf     analyse             ⋯
     │ Abstract…?                         Int32  String       String  Float64  String  Float64  String              ⋯
─────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ LineString(Ptr{Nothing} @0x00007…      1  01A903905CD  DB1         0.0  FB1         0.0  2 - Bretelles autor ⋯
   2 │ LineString(Ptr{Nothing} @0x00007…      2  01A903905CD  DB2         0.0  FB2         0.0  2 - Bretelles autor
   3 │ LineString(Ptr{Nothing} @0x00007…      3  01A903905CD  DB3         0.0  FB3         0.0  2 - Bretelles autor
   4 │ LineString(Ptr{Nothing} @0x00007…      4  01A903905CD  DB4         0.0  FB4         0.0  2 - Bretelles autor
   5 │ LineString(Ptr{Nothing} @0x00007…      5  01A903910CD  DB1         0.0  FB1         0.0  2 - Bretelles autor ⋯
                                                                                                     1 column omitted

julia> AG.IFeatureLayer(df; name="roads")
Layer: roads
  Geometry 0 (): [wkbUnknown], LINESTRING (874644.4...), ...
     Field 0 (gid): [OFTInteger], 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, ...
     Field 1 (roadcode): [OFTString], 01A903905CD, 01A903905CD, 01A903905CD, ...
     Field 2 (plod): [OFTString], DB1, DB2, DB3, DB4, DB1, DB2, DB1, DB2, ...
     Field 3 (absd): [OFTReal], 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, ...
     Field 4 (plof): [OFTString], FB1, FB2, FB3, FB4, FB1, FB2, FB1, FB2, ...
...
 Number of Fields: 7
```

src/tables.jl
```diff
@@ -47,3 +47,352 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+3. create layer named `name` in a MEMORY dataset geomfields and fields types inferred from `rows`'s column types
+4. populate layer with `rows` values
+
+"""
+function _fromtable(
+    sch::Tables.Schema{names,types},
+    rows;
+    name::String,
+    parseWKT::Bool,
+    parseWKB::Bool,
```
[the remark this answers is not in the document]
mathieu17g: > > 3. Part of the benefit of the fieldtypes arg is also the ability to skip the try-catch block for inferring the field type:
> 
> The try-catch block should not be time consuming, since it only work on the schema. But I can add a `fieldtypes` as `Union{Nothing, Dict{Int, GeometryOrFieldType}, Dict{String, GeometryOrFieldType}}`

I have added the `fieldtypes` kwarg option, without any specific tests yet in commit 125a076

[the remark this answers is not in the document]
mathieu17g: I have added tests on `geomcols` kwarg and corrected some issues raised by those tests in commit 9e90dc14595c1236dd4b4ead44e91ab2b07625a8
-> 10 pts increase on PR coverage.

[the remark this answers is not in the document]
mathieu17g: GDAL.jl seems to be broken (failing to precompile) after an update of geos from 3.9 to 3.10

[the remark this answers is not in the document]
mathieu17g: I openned an issue in GDAL.jl: https://github.com/JuliaGeo/GDAL.jl/issues/126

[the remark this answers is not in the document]
mathieu17g: I have added tests on the `fieldtypes` kwarg. 

Pkg.test is 100% OK and Coverage is 100% when tested locally. So when JuliaGeo/GDAL.jl#126 is fixed, CI for this PR should be OK

@visr or @yeesian do you know how to test error messages as it can be done with other messages by mixing `@test` with `@test_logs`. 
I have not figured out how to mix `@test_throws` and `@test_logs`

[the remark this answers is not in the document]
mathieu17g: > @visr or @yeesian do you know how to test error messages as it can be done with other messages by mixing @test with @test_logs.
> I have not figured out how to mix @test_throws and @test_logs

I found a simple way in https://discourse.julialang.org/t/how-to-test-a-default-error-raising-methods-message/19224
`@test_throws ErrorException("message") ex`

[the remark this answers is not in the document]
mathieu17g: @yeesian it should be ready for review now

src/tables.jl
```diff
@@ -47,3 +47,585 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+                if e isa MethodError
+                    error(
+                        "Cannot convert column \"$colname\" (type $coltype) to neither IGeometry{::OGRwkbGeometryType} or OGRFieldType and OGRFieldSubType",
+                    )
+                else
+                    throw(e)
+                end
+            end
+        end
+    end
```
yeesian, line 170: I tried, but the highlighted section is the only part of this function that I was able to comprehend. I.e. I'm afraid I found it too challenging to follow L172-282 of this file in this PR (at the time of this writing).

I can see the value of having a data-driven inference of geometries from WKT/WKB columns, but I cannot in good faith say that I'll know how to maintain it. For unblocking this PR, can we document it in a feature request for now, so the code will still be available if we want to go down this path?

src/tables.jl
```diff
@@ -47,3 +47,585 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+    layer_name::String,
+    geomcols::Union{Nothing,Vector{String},Vector{Int}} = nothing, # Default value set as a convinience for tests
+    fieldtypes::Union{
+        Nothing,
+        Dict{Int,Union{OGRwkbGeometryType,Tuple{OGRFieldType,OGRFieldSubType}}},
+        Dict{
+            String,
+            Union{OGRwkbGeometryType,Tuple{OGRFieldType,OGRFieldSubType}},
+        },
+    } = nothing, # Default value set as a convinience for tests
```
yeesian, line 461: nit: convinience -> convenience

src/tables.jl
```diff
@@ -47,3 +47,585 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+2. split `rows`'s columns into geometry typed columns and field typed columns
+3. create layer named `name` in a MEMORY dataset geomfields and fields types inferred from `rows`'s column types
+4. populate layer with `rows` values
+
+"""
+function _fromtable(
+    sch::Tables.Schema{names,types},
+    rows;
+    layer_name::String,
+    geomcols::Union{Nothing,Vector{String},Vector{Int}} = nothing, # Default value set as a convinience for tests
```
yeesian, line 453: nit: convinience -> convenience

src/tables.jl
```diff
@@ -47,3 +47,585 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+    } = nothing, # Default value set as a convinience for tests
+)::IFeatureLayer where {names,types}
+    # Test coherence of `geomcols` and `fieldtypes` and normalize them with indices for schema names
+    (spgeomcols, spfieldtypes) = _coherencecheckandnormalizationofkwargs(
+        geomcols,
+        fieldtypes,
+        string.(sch.names),
+    )
+
+    # Infer geometry and field types
```
yeesian, line 470: I think we can remove this comment

src/tables.jl
```diff
@@ -47,3 +47,585 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+    AGtypes = _infergeometryorfieldtypes(sch, rows, spgeomcols, spfieldtypes)
+
+    # Create layer
```
yeesian, line 473: I think we can remove this comment

src/tables.jl
```diff
@@ -47,3 +47,585 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+end
+
+"""
+    IFeatureLayer(table; kwargs...)
+
+Construct an IFeatureLayer from a source implementing Tables.jl interface
+
+## Keyword arguments
+- `layer_name::String = ""`: name of the layer
+- `geomcols::Union{Nothing, Vector{String}, Vector{Int}} = nothing`: if `geomcols` is different from nothing, only the specified columns (by names or number) will be converted to geomfields
```
yeesian, line 551: Have we considered not having this as part of the function?

* Rationale: If users knew what the `geomcols` are, they can convert before calling this function.
* In this proposal: the `fieldtypes` keyword arg should be the only way in which users can hint/coerce. 
* This might avoid the need to have a `_coherencecheckandnormalizationofkwargs(...)`

src/tables.jl
```diff
@@ -47,3 +47,585 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+        },
+    },
+    colnames = Vector{String},
+)::Tuple{
+    Union{Nothing,Vector{Int}},
+    Union{
+        Nothing,
+        Dict{Int,Union{OGRwkbGeometryType,Tuple{OGRFieldType,OGRFieldSubType}}},
+    },
+}
```
yeesian, line 315: To the extent I can, I'd prefer to avoid creating/supporting an API that requires such a function. (Nonetheless, I want to say I really really appreciate the work that went into this function.)

yeesian (changes requested):
Thanks! I think the overall flow is there, but the amount of coherence checks and inference and so on makes me very hesitant about the API. I've suggested some changes to see if they might help keep things simple.

src/ogr/geometry.jl
```diff
@@ -1636,3 +1640,28 @@ for (f, rt) in ((:create, :IGeometry), (:unsafe_create, :Geometry))
         end
     end
 end
+
+# Conversion from GeoInterface geometry
+# TODO handle the case of geometry collections
```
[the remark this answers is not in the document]
mathieu17g: I have added the conversion for `GeoInterface.AbstractGeometryCollection` in commit 5cd1f16
Note that I could not make use of `GDAL.ogr_g_addgeometrydirectly`. I did not figure out how avoid the crash I had systematically with it.

I have also:
- Moved from `@generated` to normal functions
- Differentiated display between `IGeometry` and `Geometry` -> to be able to distinguish them when using REPL. If it suits you I will make an issue to generalize to the other objects having both interactive version and unsafe version
- Added tests on compact display for `AbstracGeometry` (e.g in use in DataFrames)

src/ogr/geometry.jl
```diff
@@ -1636,3 +1640,28 @@ for (f, rt) in ((:create, :IGeometry), (:unsafe_create, :Geometry))
...
+@generated function convert(
```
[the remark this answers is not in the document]
mathieu17g: I have moved from `@generated` to normal functions in commit 5cd1f16

src/tables.jl
mathieu17g: Well, I'm still around :-) And I could deactivate the WKT/WKB column detection by default via an `auto::Bool=false` kwarg which have it value `true` value mutually exclusive with `fieldtypes != nothing`

But here is an explanation, that I hope will clarify the code
#### 1. L145-L170: Parse stage 1 (based on schema) -> Identify column types for later layer population stage
#### 2. L172-L258: Parse stage 2 (based on data) for column that may be WKT or WKB columns and therefore identified as IGeometry columns for later layer population stage
```
For each column C of type string (which may be a WKT column) or binary (which may be a WKB column)
    1. Iterate though elements E of C until one E cannot be converted to IGeometry (except when = nothing or missing)
    2. If all element of C can be converted to IGeometry, mark C as a IGeometry column for later layer population stage
       Else leave it marked as a Field column 
```
You told me in a previous PR that you were not fond of ∈, ∉, ∩, ∪, ⊆, ⊈, ... I naturally use them since it . If it helps to maintain of the code, I can of course replace them by their plain function names.
#### 3. L260-L285 (induced by the `geomcols` kwarg addition): Enforce IGeometry columns to be as user specified and ensure that it is coherent with the actual data for later layer population stage. Raise an error otherwise, with more information than would be available at layer population stage (simple `convert MethodError`)
```
If there is a list of columns marked as IGeometry by the user (via `geomcols` kwarg)
    1. If there is one column in this list that has not been identified as an IGeometry column at parse stage 1 or 2 ERROR  
    2. For each column C identified as IGeometry as parse stage 1 or 2 and outside of this list
           If C can be converted to a Field column mark it so for later layer population stage
           Else ERROR
```
By the way, I have added another restriction on columns to be parsed at stage 2, by excluding columns specified in `fieldtypes` kwarg. See commit d08a47c

src/tables.jl
mathieu17g: Ok thanks, corrected in commit ad14efd

src/tables.jl
mathieu17g: Ok thanks, corrected in commit ad14efd

src/tables.jl
mathieu17g: OK

src/tables.jl
mathieu17g: > Rationale: If users knew what the `geomcols` are, they can convert before calling this function.

It will always be faster for the user not to have to convert :-)

---

> In this proposal: the `fieldtypes` keyword arg should be the only way in which users can hint/coerce.
It is a lot easier:
- to type a list of `Int` or `String` for `geomcols` than 
- to type in a `Dict` a list of `Int` or `String` PLUS a list `IGeometry{OGRwkbGeometryType}` or `Tuple{OGRFieldType, OGRFieldSubType}`  for `fieldtypes` kwarg

Therefore I would rather drop the `fieldtypes` kwarg.

---

> This might avoid the need to have a `_coherencecheckandnormalizationofkwargs(...)`

You would still need part of what is done in this function either you drop `geomcols` or `fieldtypes`.
But the function would be simpler if you drop `fieldtypes` than if you drop `geomcols
There 4 parts in it (gathering the comments within) :
| Part | Description | `fieldtypes`dropped | `geomcols` dropped |
|---|:--|---|---|
|1|Test coherence of `geomcols` and normalize it with indices of schema names |kept|dropped|
|2|Test coherence `fieldtypes` with schema names, and normalize it to a `Dict{Int, ...}` with indices of schema names|dropped|kept|
|3|Test coherence of `spfieldtypes` and `spgeomcols`|dropped|dropped|
|4|Test coherence of `OGRFieldType` and `OGRFieldSubType` in `fieldtypes` kwarg|dropped|kept|

So it is 3 drops for `fieldtypes` dropping against 2 drops for `geomcols` dropping: `geomcols` dropping wins :-)

src/tables.jl
mathieu17g: That was what I was fearing when I asked:

> But is it worth the extra complexity ?

in https://github.com/yeesian/ArchGDAL.jl/pull/243#discussion_r738021766 :-)

But now, it is done, I think it will be helpful for most, and I will help you maintain it. We could just drop `fieldtypes` kwarg for which personally I see no usage.

src/tables.jl
yeesian: I'll need more time to work through your code and comments -- thank you in advance for your initiative and patience!

src/tables.jl
```diff
@@ -47,3 +47,587 @@ function schema_names(featuredefn::IFeatureDefnView)
...
+"""
+function _infergeometryorfieldtypes(
+    sch::Tables.Schema{names,types},
+    rows,
+    spgeomcols::Union{Nothing,Vector{String},Vector{Int}},
+    spfieldtypes::Union{
+        Nothing,
+        Dict{Int,Union{OGRwkbGeometryType,Tuple{OGRFieldType,OGRFieldSubType}}},
+    },
+) where {names,types}
```
yeesian, line 142: Can we start with a simple implementation where this is
```julia
_infergeometryorfieldtypes(sch::Tables.Schema{names,types}, rows) where {names,types}
```
?

The behavior will be based on the current implementation with `spgeomcols=nothing` and `spfieldtypes=nothing`.

Let's introduce the functionality for `spgeomcols` in a follow-up PR (we can drop `spfieldtypes` in that PR per your recommendation).

src/tables.jl
yeesian: You have convinced me about the utility of the WKT/WKB parsing, and the lack of utility of `spfieldtypes`. But this PR is really hard to review for correctness.

I'll be able to approve this PR if we stash the WKT/WKB comments & code in an issue/follow-up PR.

src/tables.jl
yeesian: I agree it will be helpful for most, and we can drop the `fieldtypes` kwarg, thank you for demonstrating it

yeesian (changes requested):

mathieu17g:
@yeesian I dropped `fieldtypes` kwarg in src/ and test/ to check if it still works properly before stashing all the WKT/WKB parsing and `geomcols` kwarg

mathieu17g:
@yeesian, in commit e8560e4, I dropped geomcols kwarg in src/ and test/

src/tables.jl
mathieu17g: Done in commit e8560e4

src/tables.jl
mathieu17g: Dropped `spgeomcols` in commit e8560e4.
I have duplicated the branch before the drop, to be able to make an issue/PR, once the current PR is merged

src/ogr/geometry.jl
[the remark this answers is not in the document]
yeesian: Thank you!

yeesian (approved):

yeesian:
Wanted to say I like the new API a lot, thank you so much for working through all the review comments!

src/types.jl
```diff
@@ -321,11 +351,29 @@ end
...
-    OFSTFloat32::Float32,
-    OFSTJSON::String,
+    OFSTNone::Int8,
+    OFSTNone::Int32,
+    OFSTBoolean::Vector{Bool},
+    OFSTBoolean::Bool, # default type comes last
+    OFSTInt16::Vector{Int16},
+    OFSTInt16::Int16, # default type comes last
+    OFSTNone::Vector{Int32},
+    OFSTInt16::Float16,
```
evetion, line 361: This (and also `Vector{Float16}`) is not correct for OFSTInt16 AFAIK

Issue #241, opened 2021-09-06
CPLDestroyMutex: Error = 16 (Device or resource busy)


evetion:
This is something that happens on the exit of many tests over at GeoArrays, but I see the same here in the tests:
https://github.com/yeesian/ArchGDAL.jl/runs/3502488899?check_suite_focus=true#step:6:67

I'd say this is something that shouldn't happen, but I'm not sure. @visr any thoughts?

visr:
Not directly. Something is perhaps not closed in the right order or something like that. Do you have a small example that we can start testing this out?

mathieu17g:
The error message appears with `Pkg.test("ArchGDAL")` but not with `include("runtests.jl")` from the test directory

visr:
I think we have that, and it's enabled by default. Evey time you get a `GDALError` it is a Julia Exception thrown by an error in GDAL, see https://github.com/JuliaGeo/GDAL.jl/blob/master/src/error.jl.

If you can reliably trigger the issue, perhaps the quickest way to locate it is with some print statements in between the tests. Not sure why there is a difference between how you run the tests, unless the session and hence the handles are kept alive in one case but not the other.

mathieu17g:
I have browsed the code of GDAL.jl/src/error.jl and GDAL.jl/gen/epilogue.jl, it is a really nice work you did there.

By commenting all `include`s and uncommenting then one by one in ArchGDAL/test/runtests.jl and running Pkg.test("ArchGDAL"), here are my findings:

- `include("test_gdalutilities.jl")`: that's where the GDAL error is written to stderr "CPLDestroyMutex: Error = 16 (Resource busy)" -> I will look after it
- `include("test_rasterattrtable.jl")`: produces a warning "There is no linear binning information." -> maybe it should be handled
- `include("test_ospy_examples.jl")`: produces a timing output. Is it necessary to keep the `@time` that must linger within ?
-  `include("test_images.jl")`: Maybe add GDAL import at the beginning instead of relying on other test files

visr:
Thanks!

Ok so probably the issue is in test_gdalutilities.jl. The `@time` can probably be removed, and indeed good to add a GDAL import to test_images.jl. For the warning, perhaps you can test/remove it by testing with https://docs.julialang.org/en/v1/stdlib/Test/#Test.@test_warn? Would be nice to not see it.

mathieu17g:
> Ok so probably the issue is in test_gdalutilities.jl

[...]

Issue #124, opened 2020-06-03
Update the documentation on Projections


yeesian:
We should update https://github.com/yeesian/ArchGDAL.jl/blob/master/docs/src/projections.md with the recent updates with GeoFormat and #120.

visr:
Docs haven't been building since mentioned PR, because we have `strict = true`, so these warnings were turned into errors:

```
┌ Error: 4 docstrings potentially missing:
│
│     Base.convert :: Tuple{Type{var"#s134"} where var"#s134"<:GeoFormatTypes.GeoFormat,Union{Type{GeoFormatTypes.FormatMode}, GeoFormatTypes.FormatMode},GeoFormatTypes.GeoFormat}
│     Base.convert :: Tuple{Type{var"#s134"} where var"#s134"<:ArchGDAL.AbstractGeometry,GeoFormatTypes.AbstractWellKnownText}
│     Base.convert :: Tuple{Type{var"#s134"} where var"#s134"<:GeoFormatTypes.AbstractWellKnownText,ArchGDAL.AbstractGeometry}
│     Base.convert :: Tuple{Type{var"#s134"} where var"#s134"<:GeoFormatTypes.GeoFormat,Union{GeoFormatTypes.CRS,Type{GeoFormatTypes.CRS}},GeoFormatTypes.GeoFormat}
└ @ Documenter.DocChecks C:\Users\visser_mn\.julia\packages\Documenter\bRylW\src\DocChecks.jl:66
```

I'm not sure how to fix them. The docstrings that were actually there in `convert.jl` caused these errors (not their absence). I had to choose between setting `strict = false` and allowing issues to creep in, or keep it strict, and disable the docstrings by putting a newline between them and the function definitions. I went for the latter for now. So @rafaqz if you wonder where your `Base.convert` docstrings went, they are still in the file, but not as docstrings.

Actually we probably should update our setup such that CI is red if the docs don't build, to prevent getting stuck with old builds.

visr:
@yeesian in the new `Documenter.jl` a canonical url should be given, which I believe also helps SEO; which do you prefer:
- `canonical = "https://yeesian.com/ArchGDAL.jl"`
- `canonical = "https://yeesian.github.io/ArchGDAL.jl"`
I assume yeesian.com since that is what it jumps to right now, but just making sure.

yeesian:
[...]

PR #245, opened 2021-09-28
Modification of tests on GDAL utilities

mathieu17g:
Fixes issue #241 and some other issues in GDAL utilities tests

[] `include("test_gdalutilities.jl")`: that's where the GDAL error is written to stderr "CPLDestroyMutex: Error = 16 (Resource busy)" 
[] `include("test_rasterattrtable.jl")`: produces a warning "There is no linear binning information." -> maybe it should be handled
[] `include("test_ospy_examples.jl")`: produces a timing output. Is it necessary to keep the `@time` that must linger within ?
[] `include("test_images.jl")`: Maybe add GDAL import at the beginning instead of relying on other test files


mathieu17g (commented):
@visr, I have narrowed the issue to one test on `AG.gdalinfo` throwing an error
As evoked in GDAL `GDALInfo()`'s [docstring](https://github.com/OSGeo/gdal/blob/1c9669a596bec859270eafb10b0d6ff311897cc7/gdal/apps/gdalinfo_lib.cpp#L206-L225), isn't it the result of GDAL `GDALInfo()` that has to be freed  ?

mathieu17g (commented):
Putting `GDAL.gdalinfo` within `AG.gdalinfo`, in a try... finally block to free `options`, does not solve the issue either

mathieu17g (commented):
Maybe it has something to do with AG.gdalinfo's options arg being overwritten inside the method -> No :-(

mathieu17g (commented):
Tried to test errored `AG.gdalinfo` with `@test` instead of `@test_throws`: issue not solved

mathieu17g:
Maybe the issue turns around catching `GDAL.GDALError` ? Here is a test with and without catching the error:
- Catching the error
```julia
julia> using ArchGDAL; const AG = ArchGDAL
ArchGDAL

julia> cd("/Users/Mathieu/.julia/dev/ArchGDAL/test/")

julia> AG.read("data/utmsmall.tif") do ds_small
           try
               AG.gdalinfo(
                   ds_small,
                   ["-novalidoption"],
               )
           catch e
               string(e)
           end
       end
"GDAL.GDALError(GDAL.CE_Failure, 6, \"Unknown option name '-novalidoption'\")"

julia> exit()
CPLDestroyMutex: Error = 16 (Resource busy)

[Opération terminée]
``` 
- Without catching the error:
```julia
julia> using ArchGDAL; const AG = ArchGDAL
ArchGDAL

julia> cd("/Users/Mathieu/.julia/dev/ArchGDAL/test/")

julia> AG.read("data/utmsmall.tif") do ds_small
           AG.gdalinfo(ds_small, ["-novalidoption"])
       end
ERROR: GDALError (CE_Failure, code 6):
	Unknown option name '-novalidoption'

Stacktrace:
 [1] gdaljl_errorhandler(class::GDAL.CPLErr, errno::Int32, errmsg::Cstring)
   @ GDAL ~/.julia/packages/GDAL/ycHfN/src/error.jl:36
 [2] gdalinfooptionsnew(papszArgv::Vector{String}, psOptionsForBinary::Ptr{Nothing})
   @ GDAL ~/.julia/packages/GDAL/ycHfN/src/GDAL.jl:18417
 [3] gdalinfo(dataset::ArchGDAL.Dataset, options::Vector{String})
   @ ArchGDAL ~/.julia/packages/ArchGDAL/zkx2f/src/utilities.jl:15
 [4] #1
   @ ./REPL[3]:2 [inlined]
 [5] read(f::var"#1#2", args::String; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ArchGDAL ~/.julia/packages/ArchGDAL/zkx2f/src/context.jl:267
 [6] read(f::Function, args::String)
   @ ArchGDAL ~/.julia/packages/ArchGDAL/zkx2f/src/context.jl:265
 [7] top-level scope
   @ REPL[3]:1

julia> exit()
CPLDestroyMutex: Error = 16 (Resource busy)

[Opération terminée]
```
-> No link with catching the error

mathieu17g:
No more ideas.
I clean up and mark the PR as final with the three corrections.
I will put the last tests in the issue for later investigations.

test/test_gdalutilities.jl
```diff
@@ -286,4 +233,4 @@ end
         @test AG.height(ds_warped) == 100
         @test AG.shortname(AG.getdriver(ds_warped)) == "MEM"
     end
-end
+end
```
yeesian, line 236: Might be good to include a newline at the end of the file in the future -- but I think it'll get cleaned up by JuliaFormatter, so it doesn't matter too much.

yeesian (approved):
Yeah I don't think it's necessary to keep the time macro, and your other observations are all on point, thank you!

visr (approved):
Looks like a good improvement, thanks!

> would it be interesting to activate some debug on GDAL (if feasible)?

Woud be! Unfortunately I don't really have other ideas for how to achieve this. You could try in https://github.com/JuliaGeo/GDAL.jl/blob/master/src/error.jl to be stricter, and raise warnings or errors on CE_Warning, now it is set to CE_Failure.

mathieu17g:
> You could try in https://github.com/JuliaGeo/GDAL.jl/blob/master/src/error.jl to be stricter, and raise warnings or errors on CE_Warning, now it is set to CE_Failure.

Nothing comes out

Maybe in the DriverManager finalization. I'll have a look...

mathieu17g (commented):
@visr and @yeesian, in the current code there is no real link between `ArchGDAL.DRIVER_MANAGER` and GDAL `DriverManager`

If you have the same reading I suggest to either:
1. try the modifications of this commit. They have, in prerequisite, the creation in GDAL.jl of (I don't know if it can be done) : 
  - type `GDAL.GDALDriverManager` 
  - function `GDAL.getgdaldrivermanager()` calling GDAL `GetGDALDriverManager()`

Plus maybe in the `destroy(::DriverManager)`:
a. deregister the custom error handler function by restoring the default with `GDAL.seterrorhandler(C_NULL)`
b. ...? 

2. just drop the mapping of the drivermanager
**EDIT**: tested -> no positive consequence on the issue

mathieu17g (commented):
Deregistering `gdaljl_errorhandler` C function at exit works :-)
@visr, Maybe it could be done in GDAL.jl `__init__()` function 

@visr, @yeesian, what about the proposition around the mapping of GDAL `DriverManager` in ArchGDAL:
- suppression?
- modification?
- leave it as it is?

yeesian (approved):
Regarding your question: my vote is to drop the driver manager mapping: this PR looks good to me once we've deleted all the redundant comments. I double-checked with the original commit in https://github.com/yeesian/ArchGDAL.jl/commit/b140d51ee144a037470ab5e5b5ddfa38f293bbb6 and think this is a much cleaner way of managing the drivers -- thank you for investigating!

yeesian:
Might be worth also updating the PR title to reflect proper finalization of driver management (by deregistering the error handler).

visr:
I don't fully understand. Now GDAL.jl deregisters the custom error handler on exit, which looks like a good thing.

But why does this PR remove the call to `GDAL.gdaldestroydrivermanager()`? Isn't that the counterpart to `GDAL.gdalallregister()` that is called when ArchGDAL loads? Should `gdaldestroydrivermanager` be called `onexit` here?

src/ArchGDAL.jl
```diff
@@ -38,19 +38,9 @@ include("tables.jl")
...
-        return drivermanager
-    end
-end
-
-const DRIVER_MANAGER = Ref{DriverManager}()
-
 function __init__()
-    DRIVER_MANAGER[] = DriverManager()
+    GDAL.gdalallregister()
+    atexit(() -> GDAL.cplseterrorhandler(C_NULL))
```
yeesian, line 43: I think per @visr's comment, `GDAL.gdaldestroydrivermanager()` should be called in `atexit()`, so this line should be something like 

```
atexit(() -> (GDAL.gdaldestroydrivermanager(); GDAL.cplseterrorhandler(C_NULL)))
```

mathieu17g: Yes you are right. I'll make the change

mathieu17g: I guess without the `GDAL.cplseterrorhandler(C_NULL)` part since it is now handled in GDAL.jl

mathieu17g:
Now we should be good to go :-)

src/ArchGDAL.jl
```diff
@@ -40,7 +40,7 @@ include("convert.jl")
 
 function __init__()
     GDAL.gdalallregister()
-    atexit(() -> GDAL.cplseterrorhandler(C_NULL))
+    atexit(() -> GDAL.gdaldestroydrivermanager())
```
visr, line 43: This can be simplified to atexit(GDAL.destroydrivermanager) since that is already a zero arg function.

mathieu17g: I overlooked it in the previous commit, but adding `() -> GDAL.gdaldestroydrivermanager()` with `atexit` in ArchGDAL brings back the issue "CPLDestroyMutex: Error = 16 (Resource busy)".

It may be linked to the fact that `atexit` adds functions in FIFO mode and that the custom error handler should be deregistered before calling `GDAL.gdaldestroydrivermanager()` 

If we want to keep `GDAL.gdaldestroydrivermanager()`, the two cleanup function should be called sequentially either in GDAL.jl or a client package as ArchGDAL

mathieu17g: Futhermore, as found [here](https://github.com/OSGeo/gdal/blob/3c518a6cac746b97f367578314e0bf784f78b861/gdal/apps/gdalasyncread.cpp#L558-L559) we could use `GDALDumpOpenDatasets(stderr)` to spot leaked dataset in tests

mathieu17g: The function `GDALDestroy` found [here](https://github.com/OSGeo/gdal/blob/3c518a6cac746b97f367578314e0bf784f78b861/gdal/gcore/gdaldllmain.cpp#L51-L100), hints to a more thorough cleanup and is called after `GDALDestroyDriverManager`call in some GDAL apps: ogr2ogr, gdalinfo, ogrinfo

Still, it seems that the custom error handler should not be called after as for `GDAL.gdaldestroydrivermanager()`

mathieu17g: I have tested a few combination. It seems that registering `gdaldestroy` at the beginning of GDAL.jl/gen/epilogue.jl `__init__()` function solves everything. We do not need to deregister the custom error handler in that same function, and since GDAL `GDALDestroy()` call `GDALDestroyDriverManager()`, we do not have to register its Julia wrapped version at exit.

It may be interresting to also register `() -> gdaldumpopendatasets("/dev/stderr")` after `gdaldestroy` to allow the user to identify unclosed datasets. But I have not found the way to pass a C pointer to `stderr` that works.

Shall I make a PR in GDAL.jl ?

mathieu17g: Well, I have made PR #124 in GDAL for comment.

And I have prepared, in case @visr accepts the latter PR, a commit below with no cleanup and but a few comments, in ArchGDAL.jl/src/ArchGDAL.jl

mathieu17g (commented):
Comments could have been:
```julia
    # As gdaldestroy() is registered first in Julia exit hooks 
    # (and then called last) in GDAL.jl/gen/epilogue.jl __init__() function, 
    # we don't have to:
    # - deregister of gdaljl_errorhandler (used to raise an error in Julia 
    #   when a GDAL error with level CEFailure is raised)
    # - destroy of the GDALDriverManager 
```
But the discussion in this PR should be enough for later reference to explain what cleanup should have been done if not handled in GDAL.jl

src/ArchGDAL.jl
yeesian: > Well, I have made PR #124 in GDAL.jl for comments.

(I'm assuming you're referring to https://github.com/JuliaGeo/GDAL.jl/pull/124 -- github links that to https://github.com/yeesian/ArchGDAL.jl/issues/124 instead)

Sounds good to me. We'll need to wait for that PR to be merged and a new version of GDAL released, before updating the dependency on GDAL in https://github.com/yeesian/ArchGDAL.jl/blob/8f820c77c790d6166f03f17f145a4ffe474933de/Project.toml#L23

mathieu17g: Yes, I was referring to JuliaGeo/GDAL.jl#124

I had not seen that GDAL dependencies was specified up to the sub sub version. You are quite conservative :-)
I was using the master version when testing locally this PR and the related ArchGDAL

yeesian: All sloppiness traces to me, and all good credit goes to @visr really :)

visr: Haha not at all! @mathieu17g note that this compatability allows all 1.y.z GDAL.jl versions. The reason we used a patch version had to do with compatibility of the underlying build, see 47cc5564d23a7cde0591311908270444bc60643a.

mathieu17g:
Oops, I rebased to master and got all those new commit reappearing in this pull request.
@yeesian I hope I rebased the right way...

yeesian:
Hmm, might it be easier to bail on this PR and start a new one? Sorry for the churn

yeesian:
Thank you!

PR #248, opened 2021-10-05
Setup of a first PkgBenchmark.jl benchmark suite

mathieu17g:
Fixes https://github.com/yeesian/ArchGDAL.jl/issues/193#issue-892504558

Here is a first step with a PkgBenchmark.jl benchmark suite benchmarking shapefile reading and converting to table with GDAL.jl (w or w/o vsizip) and Shapefile.jl

```julia
julia> using PkgBenchmark

julia> using ArchGDAL

julia> benchmarkpkg(ArchGDAL)
PkgBenchmark: Running benchmarks...
PkgBenchmark: using benchmark tuning data in /Users/Mathieu/.../dev/ArchGDAL/benchmark/tune.json

(1/1) benchmarking "shapefile_to_table"...
  (1/3) benchmarking "frenchroads_with_GDAL.jl_via_vsizip"...
  done (took 5.725868093 seconds)
  (2/3) benchmarking "frenchroads_with_GDAL.jl"...
  done (took 5.672228116 seconds)
  (3/3) benchmarking "frenchroads_with_Shapefile.jl"...
  done (took 5.685400644 seconds)
done (took 17.703220046 seconds)
Benchmarking 100%|█████████████████████████████████████████████████████| Time: 0:00:19
Benchmarkresults:
    Package: /Users/Mathieu/OneDrive - Altice Campus SFR/code_julia/radio_sites_near_motorways/dev/ArchGDAL
    Date: 5 Oct 2021 - 10:47
    Package commit: 1e5fe5
    Julia commit: ae8452
    BenchmarkGroup:
        1-element BenchmarkTools.BenchmarkGroup:
          tags: []
          "shapefile_to_table" => 3-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "frenchroads_with_GDAL.jl_via_vsizip" => Trial(304.691 ms)
                  "frenchroads_with_GDAL.jl" => Trial(194.496 ms)
                  "frenchroads_with_Shapefile.jl" => Trial(66.557 ms)
```

Next step: integrate it with Github actions with BenchmarkCI.jl

PR #249, opened 2021-10-08
Modified benchmark.yml to benchmark last commit against master, instead of against itself as has to be done in jumpstart PR #248

mathieu17g:
In commit #248, I had to restrict benchmark of  the last commit to itself since the benchmark configuration as not present in master before merge.

Now every new commit should be benchmarked to master


yeesian:
Thank you!

PR #268, opened 2021-12-18
support imview for Unsigned besides UInt8

visr:
The single argument normedview is only for UInt8.
This assumes that the full range will be used to go from 0 to 1.

https://juliaimages.org/stable/function_reference/#ImageCore.normedview

I ran into this with a UInt16 landsat image, and with this change I can `imread` it. Though I see that the tests only cover UInt8.

EDIT: added a second commit to make imread work on RasterDataset


yeesian (approved):
Thank you!

PR #270, opened 2022-01-08
update tests for GDAL 3.4

visr:
This will fail for now, since https://github.com/JuliaGeo/GDAL.jl/pull/131 is not yet merged and tagged. But locally this works using that branch.

GDAL.jl is dropping julia 1.5 and older support in that PR. As long as ArchGDAL doesn't use features that are only available in GDAL 3.4, it is fine to keep the current GDAL and julia compat in Project.toml, though then perhaps this PR should be made in a way that it will pass tests on julia 1.3 as well.


yeesian:
I think it's okay to move to GDAL 3.4 when that happens. It'll have to be a new release, so that users of previous GDAL versions can still be pinned to it if they want?

visr:
Yeah indeed, that will be easiest for us as well. GDAL.jl will get a minor (non breaking) release.

visr:
I pushed a commit that bumps the ArchGDAL minor version to v0.8.0. Since ArchGDAL is pre-v1 it is my understanding that it needs to be a breaking/minor release.

Though if you prefer to include other changes in v0.8.0 we could hold off on this as well.

evetion (approved):

yeesian:
I do not have any other plans for v0.8, so this PR looks good to me :)

PR #271, opened 2022-01-14
import ArchGDAL as AG

visr:
Purely cosmetic change / increasing the consistency.

The AG abbreviation was already used almost everywhere in the tests. This changes the last few, and changes the import over to `import ArchGDAL as AG`. Since in the docs there is mixed usage, I didn't change it there, sticking to `using ArchGDAL; const AG = ArchGDAL` which gets both ArchGDAL and AG in the namespace. Same for GeoFormatTypes.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,7 +1,6 @@
 using BenchmarkTools
 using ZipFile
-using ArchGDAL;
-const AG = ArchGDAL;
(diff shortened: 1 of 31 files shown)


meggart (approved):
Nice! Looks much cleaner when consistently using the new syntax.

yeesian (approved):
Thank you!

PR #273, opened 2022-01-18
Update DiskArrays

meggart:
This supersedes #272 , where tests fail because of this `eachchunk` function.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -19,7 +19,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 [compat]
 CEnum = "0.4"
 ColorTypes = "0.10, 0.11"
-DiskArrays = "0.2.4"
+DiskArrays = "0.2.4, 0.3"
 GDAL = "1.3"
 GeoFormatTypes = "0.3"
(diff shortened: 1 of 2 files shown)


visr:
Nice, tests pass benchmark fails, but it seems that would be resolved by this PR:

```
ERROR: LoadError: Unsatisfiable requirements detected for package DiskArrays [3c3547ce]:
 DiskArrays [3c3547ce] log:
 ├─possible versions are: 0.1.0-0.3.0 or uninstalled
 ├─restricted to versions 0.2.4-0.2 by ArchGDAL [c9ce4bd3], leaving only versions 0.2.4-0.2.13
 │ └─ArchGDAL [c9ce4bd3] log:
 │   ├─possible versions are: 0.8.0 or uninstalled
 │   └─ArchGDAL [c9ce4bd3] is fixed to version 0.8.0
 └─restricted to versions 0.3.0 by an explicit requirement — no versions left
 ```

visr (approved):

mathieu17g:
>it seems that would be resolved by this PR

Yes probably since benchmark does not have DiskArrays in its own project.toml

Issue #277, opened 2022-01-27
Remove DiskArrays 0.2.4 compat entry from Project.toml


maxfreu:
It seems like version 0.2.4 is not compatible with the latest ArchGDAL version. Cross posting from [here](https://github.com/rafaqz/Rasters.jl/issues/244)

>Looks like ArchGDAL has done a minor version bump of DiskArrays.jl, and not removed compat for versions before 0.3. That bug is entirely between ArchGDAL and version 0.2.4 of DiskArrays, which is no longer supported but is still included in the ArchGDAL Project.toml.
Please make an issue over there to remove compat with disk arrays 0.2.4, and to remove the current ArchGDAL version from the julia registry.

rafaqz:
Yanked: https://github.com/JuliaRegistries/General/pull/53421

PR #278, opened 2022-01-28
remove DiskArrays 0.2.4 compat

maxfreu:
I just removed the DiskArrays 0.2.4 compat entry in the Project toml, because ArchGDAL is not compatible with the old version. You would have to tag a new version and remove the broken one from the registry. Closes https://github.com/rafaqz/Rasters.jl/issues/244 and https://github.com/yeesian/ArchGDAL.jl/issues/277


rafaqz:
@yeesian is it ok if I merge this? Then we can make a new patch version and yank the last one from the registry.

PR #284, opened 2022-03-31
Enable more exotic types in reading and writing.

evetion:
As an extension of #220. I've taken some of the work of @mathieu17g from the WIP at #243 in terms of type conversions.

Adds support for reading again (which broke somewhere during the 0.8 release train)
- OFSTBoolean
- OFSTInt16
- OFSTFloat32

Make sure we can write
- Int8
- UInt8
- UInt16
- UInt32
- Float16

By widening the type (i.e. `UInt32` => `Int64`), which requires the correct conversion in both `OGRFieldType` as `OGRFieldSubType`.

- [x] Add tests

### DIFF
--- a/src/ogr/feature.jl
+++ b/src/ogr/feature.jl
@@ -567,52 +567,29 @@ function setfield!(feature::Feature, i::Integer, value::Missing)::Feature
     return feature
 end
 
-function setfield!(feature::Feature, i::Integer, value::Int32)::Feature
+function setfield!(
+    feature::Feature,
+    i::Integer,
+    value::Union{Bool,UInt8,Int8,UInt16,Int16,Int32},
+)::Feature
     GDAL.ogr_f_setfieldinteger(feature.ptr, i, value)
     return feature
 end
 
-function setfield!(feature::Feature, i::Integer, value::UInt16)
-    GDAL.ogr_f_setfieldinteger(feature.ptr, i, Int32(value))
-    return feature
-end
-
-function setfield!(feature::Feature, i::Integer, value::Int16)::Feature
-    GDAL.ogr_f_setfieldinteger(feature.ptr, i, value)
-    return feature
-end
-
-function setfield!(feature::Feature, i::Integer, value::Union{UInt8,Int8})
-    GDAL.ogr_f_setfieldinteger(feature.ptr, i, Int16(value))
-    return feature
-end
-
-function setfield!(feature::Feature, i::Integer, value::Bool)::Feature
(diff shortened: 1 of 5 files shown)


evetion:
I don't think the failing tests on nightly are related `MethodError: no method matching regenerateoverviews!(::ArchGDAL.IRasterBand{UInt8}, ::Vector{Matrix{UInt8}})`

src/ogr/feature.jl
```diff
@@ -218,6 +218,30 @@ end
...
+    asint16(feature::Feature, i::Integer)
+
+Fetch field value as integer 16 bit.
+
+### Parameters
+* `feature`: the feature that owned the field.
+* `i`: the field to fetch, from 0 to GetFieldCount()-1.
+"""
+asint16(feature::Feature, i::Integer)::Int16 =
+    convert(Int16, GDAL.ogr_f_getfieldasinteger(feature.ptr, i))
```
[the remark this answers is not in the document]
yeesian: Oh okay I think that's answered by `OFSTInt16 => asint16,` later further below, so resolving this comment

src/ogr/feature.jl
```diff
@@ -533,11 +572,21 @@ function setfield!(feature::Feature, i::Integer, value::Int32)::Feature
     return feature
 end
 
+function setfield!(feature::Feature, i::Integer, value::UInt16)
+    GDAL.ogr_f_setfieldinteger(feature.ptr, i, Int32(value))
```
yeesian, line 576: Out of curiosity: would it still work if we didn't convert `value` to `Int32` (i.e. would the ccall in `GDAL.ogr_f_setfieldinteger(...)` do it automatically)? Asking also because we don't seem to do it for `value::Int16` on L580.

The tests seem to indicate that we need to do appropriate widening / "type promotions" -- are we doing it here because the ccall conversions aren't the ones we expect? Similar questions for the conversions below.

yeesian (approved):
Thank you!

src/ogr/feature.jl
evetion: Good point, it seems to work fine without them indeed, I've now unified all the types for the `setfield!` operation. 

In the tests I wanted to make sure that in the Julia ecosystem the subtype is used, even though the GDAL API only sees it as metadata. GDAL does check whether the value fits in the subtype in `ogr_f_setfieldx`.

yeesian: Thank you!

yeesian (approved):

PR #286, opened 2022-03-31
Enable preperation of geometry

evetion:
Preparing geometry is useful when you repeat spatial operations on a geometry, as required, calculated, information needed for said operation is cached. See https://blog.cleverelephant.ca/2008/10/postgis-performance-prepared-geometry.html

The open question is if/where we check for the remaining GDAL `ogrhaspreparedgeometrysupport` method in all these calls. Maybe a warning on init? I'm not sure what happens when a non-Yggdrasil GDAL doesn't have prepgeom support.

### DIFF
--- a/src/ogr/geometry.jl
+++ b/src/ogr/geometry.jl
@@ -87,6 +87,18 @@ function destroy(geom::AbstractGeometry)::Nothing
 end
 
 """
+Destroy prepared geometry object.
+
+Equivalent to invoking delete on a prepared geometry, but it guaranteed to take place
+within the context of the GDAL/OGR heap.
+"""
+function destroy(geom::AbstractPreparedGeometry)::Nothing
+    GDAL.ogrdestroypreparedgeometry(geom.ptr)
+    geom.ptr = C_NULL
+    return nothing
+end
+
+"""
     clone(geom::AbstractGeometry)
 
 Returns a copy of the geometry with the original spatial reference system.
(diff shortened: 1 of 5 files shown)


src/ogr/geometry.jl
```diff
@@ -121,6 +133,18 @@ creategeom(geomtype::OGRwkbGeometryType)::IGeometry =
...
+"""
+    preparegeom(geom::AbstractGeometry)
+
+Create an prepared geometry of a geometry. This can speed up operations which interact
+with the geometry multiple times, by storing caches of calculated geometry information.
+"""
+preparegeom(geom::AbstractGeometry)::IPreparedGeometry =
+    IPreparedGeometry(GDAL.ogrcreatepreparedgeometry(geom.ptr))
+
+unsafe_preparegeom(geom::AbstractGeometry)::PreparedGeometry =
```
yeesian, line 145: Can we add a unit test for this?

src/ogr/geometry.jl
```diff
@@ -656,6 +683,9 @@ Returns `true` if g1 contains g2.
 contains(g1::AbstractGeometry, g2::AbstractGeometry)::Bool =
     Bool(GDAL.ogr_g_contains(g1.ptr, g2.ptr))
 
+contains(g1::AbstractPreparedGeometry, g2::AbstractGeometry)::Bool =
```
yeesian, line 686: Can we add a unit test for this?

yeesian (approved):
Thank you! LGTM apart from unit test coverage for some of the methods

rafaqz (approved):
This looks good to me. What "prepare" means is a bit vague - but as GDAL uses it I guess that's fine.

Issue #285, opened 2022-03-31
Failing tests on 1.9


evetion:
As seen in master and in recent PRs (#284), nightly fails with

`MethodError: no method matching regenerateoverviews!(::ArchGDAL.IRasterBand{UInt8}, ::Vector{Matrix{UInt8}})`
https://github.com/yeesian/ArchGDAL.jl/runs/5513853907?check_suite_focus=true#step%3A6%3A165=

yeesian:
Thank you for filing this; I looked at the code and it seems that everything's working as-intended, but please correct me if I'm wrong! 

In the meantime, let me doublecheck that failure on nightlies are not a blocker for merging PRs

evetion:
Ok, from what I can see, in the failing function `regenerateoverviews!(::T, ::Vector{<:ArchGDAL.AbstractRasterBand})` Julia normally dispatches fine on `Vector{<:ArchGDAL.AbstractRasterBand}`. In Julia 1.9 it doesn't and walks all the way up its supertypes to `Matrix` as `ArchGDAL.AbstractRasterBand <: AbstractDiskArray{T,2} <: <: AbstractArray{T,N}`.

Will file an upstream bug report.

yeesian:
I really appreciate your investigation and follow-up in https://github.com/JuliaLang/julia/issues/44821. The lack of stability in the common type for vectors seems slightly concerning. Should I consider setting up a system of promotion rules defined for the types in this package?

evetion:
It seems wise to add promotion rules indeed, but only for the (probably small amount) of cases where it's needed. In the test we triggered this by mixing both `RasterBand` and `IRasterBand`, the same is probably needed for Geometry, but I don't know more cases. I also don't expect a user to quickly mix these, as the public interface mostly works with the `I` variants.

PR #287, opened 2022-04-01
Fix nightly by setting the type of Vector to prevent unwanted promotion

evetion:
Partial fix for #285. Required due to change in https://github.com/JuliaLang/julia/pull/44096


yeesian:
Thank you!

PR #290, opened 2022-04-10
Implement new GeoInterface traits

evetion:
[GeoInterface 1.0](https://juliageo.org/GeoInterface.jl/stable/) makes use of Traits, this is a PR to update our pre-1.0 GeoInterface implementation.

Note that I had to add some methods to align with GeoInterface:
- Added `getm` method, previously not wrapped
- Added `getgeom` for LineStrings. Technically OGR doesn't expose this (and wants us to use getpoint), but our `ngeom` does have a switch for Lines.
- Added `is3d` and `ismeasured` to be used when getting the the third or fourth coordinate
- Added `keys` and `values` to a Feature

I've also added dependencies on `GeoInterfaceRecipes`, a plot package based on `GeoInterface`, which previously was integrated into `GeoInterface` and on `Extents`, a package defining bounding boxes as `NamedTuples`.

I also note that much in ArchGDAL doesn't support Z en M coordinates properly, for which I will create a separate issue. In short, many of our creation methods don't result in the correct type (just a Point instead of Point25D etc.). This makes it hard to correctly implement `getcoord` for the `m` dimension, as it could be both the third or fourth dimension, depending on the type.


evetion:
- [ ] Benchmark CI doesn't seem to update to latest GeoInterface (still using `v0.5.7`?)
- [ ] Plots recipes are required for the documentation to work @rafaqz

visr:
> Benchmark CI doesn't seem to update to latest GeoInterface (still using v0.5.7?)

That'd be because of Shapefile.jl

https://github.com/yeesian/ArchGDAL.jl/blob/e401e0079d97dac9d4755a4bc9dce345a9c5bf81/benchmark/Project.toml#L6

src/geointerface.jl
```diff
@@ -23,57 +35,223 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
...
+
+    function GeoInterface.asbinary(::GeometryTraits, a::AbstractGeometry)
+        return toWKB(geom)
+    end
+
+    function GeoInterface.astext(::GeometryTraits, a::AbstractGeometry)
+        return toWKT(geom)
+    end
+
+    function GeoInterface.convert(::Type{IGeometry}, type::GeometryTraits, geom)
```
visr, line 250: Base.convert?

evetion:
@rafaqz 

Can you take a look here at the `convert` options? I keep running into ambiguous things, as our suggetion for implementing `convert` overrides *everything*, including existing methods such as those for GeoFormatTypes.

```julia
  Expression: sprint(print, convert(AG.IGeometry, json)) == "Geometry: POINT (100 70)"
  MethodError: convert(::Type{ArchGDAL.IGeometry}, ::GeoFormatTypes.GeoJSON{String}) is ambiguous. Candidates:
    convert(::Type{T}, source::X) where {T<:ArchGDAL.AbstractGeometry, X<:GeoFormatTypes.GeoJSON} in ArchGDAL at /Users/evetion/code/ArchGDAL.jl/src/convert.jl:48
    convert(::Type{T}, geom::X) where {T<:ArchGDAL.IGeometry, X} in ArchGDAL at /Users/evetion/code/ArchGDAL.jl/src/geointerface.jl:262
  Possible fix, define
    convert(::Type{T}, ::X) where {T<:ArchGDAL.IGeometry, X<:GeoFormatTypes.GeoJSON}
```

rafaqz:
Ww need to dispatch both on `ArchGDAL.AbstractGeometry`. That will work.

evetion:
> We need to dispatch both on ArchGDAL.AbstractGeometry. That will work. Both first arguments need to be equally specific.

Well that's what I did first, but it segfaults preparedgeometry tests(!). Something to investigate, but otherwise, I'm leaning towards the idea that only IGeometries should be created from conversion, while _all_ geometries can function as a source. But maybe better as an issue than to keep changing more and more in this PR.

rafaqz:
Both can be on `IGeometry` as well I guess? I cant remember the distinctions honestly.

src/geointerface.jl
```diff
@@ -23,57 +35,282 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
...
+        return getcoorddim(geom)
+    end
+
+    function GeoInterface.getcoord(::GeometryTraits, geom::AbstractGeometry, i)
+        if i == 1
+            getx(geom, 0)
+        elseif i == 2
+            gety(geom, 0)
+        elseif i == 3  # M is an option here, but not properly supported by ArchGDAL yet
+            getm(geom, 0)
```
yeesian, line 127: Out of curiosity, when is it Z, versus M?

yeesian:
> I'm leaning towards the idea that only IGeometries should be created from conversion, while all geometries can function as a source

Yeah I think this might be a viable approach (or at least it was how I was mentally thinking about things without properly having articulated them anywhere)

src/geointerface.jl
evetion: Well I made a mistake here, I think we should call `getz` here. It could be `m`, but only when `is3d` is false and `ismeasured` true?

evetion:
Benchmark should be able to run when https://github.com/JuliaGeo/Shapefile.jl/pull/72 is merged and released.

src/geointerface.jl
```diff
@@ -23,57 +35,282 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
...
+
+    function GeoInterface.getgeom(
+        ::GeometryTraits,
+        geom::AbstractGeometry,
+        i::Integer,
+    )
+        return getgeom(geom, i - 1)
+    end
+
+    # Operations
```
yeesian, line 151: Can we add unit tests for the operations introduced from this line onwards? Other than that, I think it's fine that some of the new traits in `GeoInterface.geomtrait(...)` aren't being covered in this PR.

yeesian (approved):
Overall LGTM, thank you so much!

Might be good to update the top level comment of this PR with the changes.

src/geointerface.jl
yeesian: Addressed by https://github.com/yeesian/ArchGDAL.jl/pull/290/commits/96f9731f18daec06b1d9c07a98903917aa8fa8ce

src/geointerface.jl
```diff
@@ -1,3 +1,15 @@
+using Extents
+
+lookup_method = Dict(
```
rafaqz, line 3: This could be `const`

src/geointerface.jl
```diff
@@ -23,57 +35,282 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
...
+        GeoInterface.MultiSurfaceTrait,
+        GeoInterface.PolyhedralSurfaceTrait,
+        GeoInterface.TINTrait,
+        GeoInterface.TriangleTrait,
+    }
+
+    # Feature
+    GeoInterface.isfeature(feat::AbstractFeature) = true
+    function GeoInterface.properties(feat::AbstractFeature)
+        return (;
```
rafaqz, line 59: This is a little confusing, maybe refactor with named variables?

rafaqz (commented):
Few minor comments, otherwise LGTM

src/geointerface.jl
visr: `const Dict{DataType, Function}`

evetion:
Ok, I can't merge without review approval here. The benchmark suite seems to require a released ArchGDAL version to test against.

yeesian (approved):
LGTM modulo the remaining open comments, thank you so much for seeing this through!

src/geointerface.jl
yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/290/commits/3079fd31974a6ef5ccbf4241ee4b5bc2c4bfca2b

src/geointerface.jl
yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/290/commits/3079fd31974a6ef5ccbf4241ee4b5bc2c4bfca2b

src/geointerface.jl
yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/290/commits/e2892d0d5a8d9a725c0a051ec173237f49e1638e

src/geointerface.jl
yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/290/commits/a56959cddc9fbd0cf5a90f1f158e07fd7f37d7f3

PR #266, opened 2021-11-22
Enhance ArchGDAL OGR objects with parametric composite types

mathieu17g:
Here is a draft for review, demonstrating:
1. a way to dispatch on type in `getfield` (to fix issue #246) bring for a 20% performance gain on layer to table conversion
2. the use of `ogr_f_stealgeometry` for a gain of an additionnal 20%

<details><summary>Performance results</summary><p>
```julia
julia> using ArchGDAL
       const AG = ArchGDAL
       using BenchmarkTools
       using Tables
       using Shapefile

julia> ds = AG.read("../../tmp/road/road.shp")
       layer = AG.getlayer(ds, 0)
       fdp_layer = AG.getFDPlayer(ds, 0)
Layer: road
  Geometry 0 (): [wkbLineString], LINESTRING (874644.4...), ...
     Field 0 (gid): [OFTInteger], 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, ...
     Field 1 (roadcode): [OFTString], 01A903905CD, 01A903905CD, 01A903905CD, ...
     Field 2 (plod): [OFTString], DB1, DB2, DB3, DB4, DB1, DB2, DB1, DB2, ...
     Field 3 (absd): [OFTReal], 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, ...
     Field 4 (plof): [OFTString], FB1, FB2, FB3, FB4, FB1, FB2, FB1, FB2, ...
...
 Number of Fields: 7

julia> @benchmark Tables.columns(Shapefile.Table("../../tmp/road/road.shp")
       )
BenchmarkTools.Trial: 71 samples with 1 evaluation.
 Range (min … max):  65.360 ms … 79.845 ms  ┊ GC (min … max): 0.00% … 12.70%
 Time  (median):     68.829 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   70.556 ms ±  3.834 ms  ┊ GC (mean ± σ):  4.00% ±  4.75%

    ▂ ▂  ▂ █  █  █▂                    ▂   █  ▂                
  ▅▁█▅██▅███▁███▅██▅▁▅▁▁▅▁▁▅▁▁▁▁▅▅▁▁▅▁▅█▁▅██▁▅██▅▁▁▁▁▁█▅▁▁▁▁█ ▁
  65.4 ms         Histogram: frequency by time        78.2 ms <

[...]

PR #294, opened 2022-04-20
Added IFeature type. Fixed memory leakage for `Feature`s.

evetion:
Triggered by out of memory errors when working with a *lot* of GeoDataFrames.

It seems that `iterate` just allocates `Feature`s without anyone ever cleaning them. This PR introduces an `IFeature` (and its parent `AbstractFeature`), which should align nicely with the other types.

Since GeoDataFrames goes through quite some calls, it is a bit hard to find these things, but I expect this is not the only memory leak remaining.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -13,6 +13,9 @@ abstract type AbstractSpatialRef end
 abstract type AbstractDataset end
 # needs to have a `ptr::GDAL.GDALDatasetH` attribute
 
+abstract type AbstractFeature end
+# needs to have a `ptr::GDAL.OGRFeatureH` attribute
+
 abstract type AbstractFeatureDefn end
 # needs to have a `ptr::GDAL.OGRFeatureDefnH` attribute
 
@@ -142,8 +145,18 @@ mutable struct IFeatureLayer <: AbstractFeatureLayer
     end
 end
 
-mutable struct Feature
+mutable struct Feature <: AbstractFeature
     ptr::GDAL.OGRFeatureH
+end
+
+mutable struct IFeature <: AbstractFeature
+    ptr::GDAL.OGRFeatureH
+
+    function IFeature(ptr::GDAL.OGRFeatureH = C_NULL)
+        feature = new(ptr)
+        finalizer(destroy, feature)
+        return feature
+    end
 end
 
 mutable struct FeatureDefn <: AbstractFeatureDefn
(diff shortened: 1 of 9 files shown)


yeesian:
Thank you for catching this! LGTM and merging so that the more comprehensive scrubbing of the codebase isn't blocking this fix from being merged.

yeesian (approved):

src/types.jl
```diff
@@ -142,8 +145,18 @@ mutable struct IFeatureLayer <: AbstractFeatureLayer
...
+mutable struct Feature <: AbstractFeature
+    ptr::GDAL.OGRFeatureH
+end
+
+mutable struct IFeature <: AbstractFeature
     ptr::GDAL.OGRFeatureH
+
+    function IFeature(ptr::GDAL.OGRFeatureH = C_NULL)
+        feature = new(ptr)
+        finalizer(destroy, feature)
```
yeesian, line 157: I think we might need to be careful about the geometries associated with the feature (see https://gdal.org/api/python_gotchas.html#python-crashes-if-you-use-an-object-after-deleting-an-object-it-has-a-relationship-with, https://gdal.org/api/python_gotchas.html#certain-objects-contain-a-destroy-method-but-you-should-never-use-it).

Maybe it can be achieved by defining [`ownedby`](https://github.com/yeesian/ArchGDAL.jl/blob/7d7bba0f649981d71048c64a24f855948af7f7c3/src/types.jl#L176-L192) for the Geometries, doublechecking all the methods that relate geometries and features and making sure that they match the intended usage? (Leaving this as an open question)

yeesian (changes requested):

src/types.jl
evetion: FWIW, the tests run, including those of GeoDataFrames. These tests include Features with Geometries. Furthermore, I believe that the getgeom calls create a copy (hence the improvements by using `stealgeometry` over in #266).

evetion: From the documentation it is the other way around. A feature is owned by a layer:

> The returned feature becomes the responsibility of the caller to delete with [OGR_F_Destroy()](https://gdal.org/api/vector_c_api.html#ogr__api_8h_1a16fb8126aa932f4b4acdaca5ea99bbbb). It is critical that all features associated with an [OGRLayer](https://gdal.org/api/ogrlayer_cpp.html#classOGRLayer) (more specifically an [OGRFeatureDefn](https://gdal.org/api/ogrfeature_cpp.html#classOGRFeatureDefn)) be deleted before that layer/datasource is deleted.

So I'd say destroying the `Feature` as I do in this PR is absolutely fine, but we need to investigate what the hierarchy of owned objects is.

yeesian: Ohh okay thanks for clarifying, that makes sense! No need to hold up the rest of this PR for a hypothetical concern

yeesian (approved):
This looks good to me now! I won't auto-merge in case you have anything else you want to add it to it. Feel free to merge if there's nothing else.

Issue #301, opened 2022-05-19
How to access the points in a polygon?


maxfreu:
I just don't get it (and can't find the relevant passage in the docs), please help:

```julia
simplepolygon = ArchGDAL.createpolygon([(0.,0.), (0.,1.), (1.,1.)])
AG.getpoint(simplepolygon,1)  # CE Failure Code 6, incompatible geometry for operation
simplepolygon[1]  # no getindex
first(simplepolygon)  # no iterate
```
In the end I want to convert the polygons to their julia counterpart from Meshes.jl.

visr:
It looks like you can use `getgeom` for this:
https://github.com/yeesian/ArchGDAL.jl/blob/e401e0079d97dac9d4755a4bc9dce345a9c5bf81/src/geointerface.jl#L69-L72

> In the end I want to convert the polygons to their julia counterpart from Meshes.jl.

A completely renewed https://github.com/JuliaGeo/GeoInterface.jl/ was just released, with a pending PR in #290. If Meshes would also be open to add this interface (it has 0 dependencies), conversion could be a simple `convert`.

maxfreu:
Omg, just now I read that getgeom indexing starts at 0... And I wondered why I always get a `NULL Geometry`. Shouldn't it simply raise an out of bounds error? Anyway, thanks for the hint & I'm looking forward to the new implementation!

visr:
That would probably be a good improvement, yes. I assume it would be a cheap check.

maxfreu:
The initial issue has been solved, so I think this can be closed :)

PR #302, opened 2022-05-20
error when accessing geom or layer out of bounds

maxfreu:
`getgeom` or `getlayer` now print an error message instead of returning a `NULL Geometry`. The original behaviour can be obtained by adding an `@inbounds` annotation in contexts where `getgeom` or `getlayer` is inlined. I think this PR breaks code, which explicitly handles `NULL Geometry`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c9ce4bd3-c3d5-55b8-8973-c0e20141b8c3"
 keywords = ["GDAL", "IO"]
 license = "MIT"
 desc = "A high level API for GDAL - Geospatial Data Abstraction Library"
-version = "0.8.4"
+version = "0.8.5"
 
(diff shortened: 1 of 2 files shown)


maxfreu:
Oh I just noticed the string continuation problem, too, strange...

> Is there a reason you didn't go for the BoundsError you mentioned in https://github.com/yeesian/ArchGDAL.jl/issues/301?

Yes:
```julia
throw(BoundsError(layer, 123))  # ERROR: BoundsError: attempt to access ArchGDAL.IFeatureLayer at index [123]
```
I found that not specific enough and it is missing the hint to 0 based indexing. And I didn't find a way to incorporate a different message into the `BoundsError`.

maxfreu:
just made another branch for this, maybe I'll reopen it later

PR #305, opened 2022-06-08
Generalise geomtypes

### LINKED ISSUE #304: Can't create geometry with non-Float64 vectors
Encountered in #290, with creation of geometry from `GeoInterface.coordinates(geom)`, which are nested `Vector`s of `<:Real`.

While you can create points with Integers:

```julia
julia> import ArchGDAL as AG
julia> AG.createpoint([1,2])
Geometry: POINT (1 2)
```

You cannot do so with more complex geometries, even with LineString:
```julia
julia> AG.createlinestring([[1,2],[1,2]])
ERROR: MethodError: no method matching createlinestring(::Vector{Vector{Int64}})
```

Not even with other Float types:
```julia
julia> AG.createlinestring([Float32.([1,2]),Float32.([1,2])])
ERROR: MethodError: no method matching createlinestring(::Vector{Vector{Float32}})
```

rafaqz:
This PR allows creating points with any `Real` numbers.

It also cleans up the code generation code used for `createpoint` etc, simplifying and reducing indentation level.

Closes #304


yeesian:
Thank you for this! It seems the only remaining method to be covered is `createmultipolygon(::Vector{Vector{T} where T})`. Apart from that, the rest seem to be pertaining more to some of the GeoInterface methods to be implemented.

yeesian (approved):
The implementation looks much cleaner, and the drop in test coverage is a weakness of the tests rather than the implementation so this LGTM, thank you!

I appreciate some of the additional tests you've added! Additional increases in test coverage can happen in the future is isn't a blocker here from my POV.

rafaqz:
Yes sometimes when you delete code you delete more lines that were covered than lines that weren't, so it looks like you've reduced test coverage. But the amount of funcitonality covered hasn't actually changed.

Issue #303, opened 2022-05-21
`is3d` and `ismeasured` are missing


evetion:
Encountered in #290.

`is3d` and `ismeasured` are missing which makes finding out which coordinate dimensions are implemented difficult. They're easily implemented though:

```julia
"""
    is3d(geom::AbstractGeometry)

Returns `true` if the geometry has a z coordinate, otherwise `false`.
"""
is3d(geom::AbstractGeometry)::Bool = GDAL.ogr_g_is3d(geom.ptr) == 1

"""
    ismeasured(geom::AbstractGeometry)

Returns `true` if the geometry has a m coordinate, otherwise `false`.
"""
ismeasured(geom::AbstractGeometry)::Bool = GDAL.ogr_g_ismeasured(geom.ptr) == 1
```

However, a simple test gives the following:

```julia
julia> point =  AG.createpoint(100, 70, 1)
Geometry: POINT (100 70 1)
julia> is3d(point)
false
julia> AG.GDAL.ogr_g_is3d(point.ptr)
2
```
Not sure if related, but looking at the source (https://github.dev/OSGeo/gdal/blob/9e9f8b294418e5f2670dacd5cacb5d6cdeabb797/ogr/ogr_geometry.h#L405), I can see that the is3d flag being set is 2 in this case.

Because of this oddity, I haven't yet implemented it.

evetion:
Well, the functions `ogr_g_is3d` and similar ones don't return a boolean, but just the `&` result. `isempty` only worked because it's flag is `0x1` 😓.

evetion:
Implemented in https://github.com/yeesian/ArchGDAL.jl/pull/290

yeesian:
Leaving this open until #290 is merged.

Issue #158, opened 2020-12-27
make tests more robust to new GDAL / GEOS / PROJ


visr:
Right now when updating the minor version of the binaries, the ArchGDAL tests usually fail due to minor differences. See for instance #157.

It would be better to modify the tests to be less sensitive to these changes.

For PROJ, we already test with `isapprox` but probably need to introduce a larger absolute difference (`@test π ≈ 3.14 atol=0.01`). One millimeter differences are probably not interesting to us anyway, we just want to make sure PROJ is doing the projection.

For GEOS, the tests often rely on WKT string comparison. However the node ordering is subject to changes, as well as the geometry ordering in a GeometryCollection. Hence it is probably better to move away from WKT string comparison, and just test if the geometries are `ArchGDAL.equals`, like was done in https://github.com/JuliaGeo/LibGEOS.jl/pull/81.

PR #307, opened 2022-06-15
Also dispatch GI.isgeometry on Type.

evetion:
This fixes a small oversight from #290.


rafaqz:
Was wondering what was going on

rafaqz (approved):

rafaqz:
Well, theres still something odd going on with the tests.

evetion:
Yep, it's because GDAL v1.4 😢 @visr

rafaqz (approved):

rafaqz:
Pinning 1.3 works for me, that will let my other PRs get merged as well.

visr:
What was up with GDAL.jl v1.4? If it is just the regular #158 we shouldn't pin it. I tested ArchGDAL locally before making the GDAL release, I believe it was just that.

rafaqz:
I'm not sure its just that? There are a lot of `nothing` return values and a method error:

https://github.com/yeesian/ArchGDAL.jl/runs/6903399033?check_suite_focus=true#step:20:296

evetion:
See https://github.com/yeesian/ArchGDAL.jl/runs/6903398191?check_suite_focus=true

More tests were failing than just a tolerance it seemed, not sure why. And pinning untill we make a release that works with 1.4 seems right?

evetion:
Failures are due to https://github.com/OSGeo/gdal/commit/1164ba2259457011d6a7be7c7368cf194f529810#diff-360c35bb4393b380a99c8d258b1a8f20dbcc9634e4ccb6e22fe7ad82593892e9L405. 

So the manual flag checks implemented in #290 to fix #303 are now not needed anymore.

visr:
Ha nice find. For all these bool integer returns I guess we need to test `> 0`.
At least GDAL.jl has a new testset now: https://github.com/JuliaGeo/GDAL.jl/commit/4f96fac893ab4a0269322f35fe452186d9309be4

PR #315, opened 2022-08-05
introduce a vector dataset writer

maxfreu:
Hi! Here comes the vector dataset writer. It still is far from perfect and needs tests, but you can already take a look.


src/ArchGDAL.jl
```diff
@@ -12,6 +12,9 @@ using CEnum
...
...S", "DERIVED", "DIMAP", "DIPEx", "DOQ1", "DOQ2", "DTED", "ECRGTOC", "ECW", "EEDAI", "EHdr", "EIR", "ELAS", "ENVI", "ERS", "ESAT", "ESRIC", "EXR", "FAST", "FIT", "FITS", "GenBin", "GeoRaster", "GFF", "GIF", "GPKG", "GRASS", "GRASSASCIIGrid", "GRIB", "GS7BG", "GSAG", "GSBG", "GSC", "GTA", "GTiff", "GXF", "HDF4", "HDF5", "HEIF", "HF2", "HFA", "RST", "ILWIS", "IRIS", "ISCE", "ISG", "ISIS2", "ISIS3", "JDEM", "JP2ECW", "JP2KAK", "JP2LURA", "JP2MrSID", "JP2OpenJPEG", "JPEG", "JPEGXL", "JPIPKAK", "KEA", "KMLSuperoverlay", "KRO", "KTX2", "L1B", "LAN", "LCP", "Leveller", "LOSLAS", "MAP", "MRF", "MBTiles", "MEM", "MFF", "MFF2", "MrSID", "MSG", "MSGN", "NDF", "netCDF", "NGSGEOID", "NGW", "NITF", "NTv2", "NWT_GRD", "NWT_GRC", "OGCAPI", "OZI", "JAXAPALSAR", "PAux", "PCIDSK", "PCRaster", "PDF", "PDS", "PDS4", "PLMosaic", "PNG", "PNM", "PostGISRaster", "PRF", "R", "Rasdaman", "Rasterlite", "SQLite", "RDB", "RIK", "RMF", "ROI_PAC", "RPFTOC", "RRASTER", "RS2", "SAFE", "SAR_CEOS", "SAGA", "SDTS", "SENTINEL2", "SGI", "SIGDEM", "SNODAS", "SRP", "SRTMHGT", "STACIT", "STACTA", "Terragen", "TGA", "TIL", "TileDB", "TSX", "USGSDEM", "VICAR", "VRT", "WCS", "WEBP", "WMS", "WMTS", "XPM", "XYZ", "Zarr", "ZMAP"]
```
yeesian, line 16: 1. Unfortunately this feels like a step back from the vision in https://trac.osgeo.org/gdal/wiki/rfc46_gdal_ogr_unification and will be hard to track with future versions of GDAL.

2. Might there be drivers that appear in both vectors?

src/dataset.jl
```diff
@@ -178,7 +178,71 @@ function write(
...
+                                 dataset=target,
+                                 geom=wkbUnknown,  # GPKG only works when using unknown here, deleting the field and recreating it later
+                                 spatialref=getspatialref(sourcelayer)) do targetlayer
+                targetlayerdefn = layerdefn(targetlayer)
+                
+                # the following try catch block is a bit hacky
+                # geopackage files dont work with the default geom field definition,
+                # because the driver expects a field "GEOMETRY" but the default is "Geometry"!?
+                # on the other hand, the ESRI Shapefile driver does not support GDAL.ogr_l_creategeomfield
+                # hence the try catch block as a workaround; it seems to work even after having delted the geomdefn
```
yeesian, line 214: Hahaha this is so annoying, thank you for the investigation and context!

src/dataset.jl
```diff
@@ -178,7 +178,71 @@ function write(
...
 )::Nothing
-    destroy(unsafe_copy(dataset, filename = filename; kwargs...))
+    drivername = GDAL.gdalgetdrivershortname(getdriver(dataset).ptr)
+    if drivername in gdal_raster_drivers
+        destroy(unsafe_copy(dataset, filename = filename; kwargs...))
+    elseif drivername in gdal_vector_drivers
+        writevectords(dataset, filename; kwargs...)
+    else
+        error("Dataset format not recognized.")
+    end
```
yeesian, line 188: 1. I think I'd almost prefer to have `writerasters(...)` and `writelayers(...)` and deprecate the `write(...)` function if there isn't a consistent GDAL API for it to be writing to both formats.
2. Would an approach of counting number of rasters and layers not work for identifying whether it should write to raster or vector?

src/dataset.jl
```diff
@@ -178,7 +178,71 @@ function write(
...
+                # hence the try catch block as a workaround; it seems to work even after having delted the geomdefn
+                try
+                    deletegeomdefn!(targetlayerdefn, 0)
+                    # add geometry field definitions
+                    for geomfieldidx in 0:ngeom(sourcelayer)-1
+                        addgeomdefn!(targetlayer, getgeomdefn(sourcelayerdef, geomfieldidx))
+                    end
+                catch err
+                    if (err isa GDAL.GDALError) && err.msg == "CreateGeomField() not supported by this layer.\n"
+                        nothing
```
yeesian, line 223: Can we have a comment explaining why is `nothing` the right behavior here?

yeesian (commented):
Just leaving some thoughts while it's a draft, thank you for digging into it!

src/ArchGDAL.jl
maxfreu: 1. Yes. It might be better to implement the approach you suggested below.
2. Yes, there are. E.g. PDF. So this is ambiguous and not a good solution.

src/dataset.jl
maxfreu: 1. I am not a master of GDAL. But with regard to python even having separate libraries for vector and raster data, it might be justified to split it into separate functions. But one could keep the write function around for backward compatibility or so.
2. That's probably a better idea than looking up the name in an ambiguous list. I'll look into that.

src/dataset.jl
maxfreu: Do you know where the difference in field names for GPKG comes from? Does ArchGDAL set any field names to some default values? Probably not, I gues...

src/dataset.jl
maxfreu: Yes, I can write that. It is just to mute this one specific error, based on the error message, which is quite brittle code. But the error codes are not that detailed as to filter this particular case out. So I couldn't help myself other than that.

src/dataset.jl
yeesian: > Do you know where the difference in field names for GPKG comes from? Does ArchGDAL set any field names to some default values?

I'm afraid I don't know.. ArchGDAL does not set field names to default values

src/ogr/featuredefn.jl
```diff
@@ -330,7 +330,7 @@ function directly, but use OGRLayer::DeleteGeomField() instead.
 This method should only be called while there are no OGRFeature objects in
 existence based on this OGRFeatureDefn.
 """
-function deletegeomdefn!(featuredefn::FeatureDefn, i::Integer)::FeatureDefn
```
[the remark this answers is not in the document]
yeesian: I don't think it's very likely anyone is depending on the return type of a "delete!" function. I began annotating functions with their return types when I was vetting the entire package for consistency of function behavior with the corresponding names and it was difficult for me to analyze otherwise (even if just visually based on the function signature).

src/dataset.jl
```diff
@@ -172,16 +172,132 @@ end
...
+For the keyword arguments for vector datasets, consult the documentation of `writelayers` 
+and for raster datasets the docs for `writerasters`.
 """
 function write(
     dataset::AbstractDataset,
     filename::AbstractString;
     kwargs...,
 )::Nothing
+    
+    if nraster(dataset) > 0
```
maxfreu, line 185: I've added checks based on the raster and layer count to sort things out.

src/dataset.jl
```diff
@@ -172,16 +172,132 @@ end
...
+* `strict`        ``true`` if the copy must be strictly equivalent, or more
+normally ``false`` if the copy may adapt as needed for the output format.
+* `options`       additional format dependent options controlling creation
+of the output file. `The APPEND_SUBDATASET=YES` option can be specified to
+avoid prior destruction of existing dataset.
+
+### Returns
+nothing
+"""
+function writerasters(dataset::AbstractDataset,
```
maxfreu, line 217: If you want more time to think about whether you really want this, I can take it out again. Now it corresponds 1:1 to the old `write`.

src/dataset.jl
```diff
@@ -172,16 +172,132 @@ end
     write(dataset::AbstractDataset, filename::AbstractString; kwargs...)
 
 Writes the dataset to the designated filename.
+
+For the keyword arguments for vector datasets, consult the documentation of `writelayers` 
+and for raster datasets the docs for `writerasters`.
 """
 function write(
```
maxfreu, line 179: I've added `writerasters` (1:1 copy of this write) and `writelayers`. Therefore `write` could be deprecated, but I didn't add the macro yet.

src/dataset.jl
```diff
@@ -172,16 +172,132 @@ end
...
+### Returns
+nothing
+"""
+function writelayers(dataset::AbstractDataset,
+                     filename;
+                     driver=getdriver(dataset),
+                     driver_options::Vector{String}=[""], 
+                     layer_options::Vector{String}=[""],
+                     chunksize=20000,
+                     use_gdal_copy=false)
```
maxfreu, line 253: The user can now choose between the gdal copy mechanism (slow but quite failsafe) or the julia version (fast, but there are some quirks e.g. with GPKG and empty column names).

src/dataset.jl
```diff
@@ -172,16 +172,132 @@ end
...
+    create(filename; driver=driver, options=driver_options, width=0, height=0, nbands=0) do target
+        for layeridx in 0:nlayer(dataset)-1
+            sourcelayer = getlayer(dataset, layeridx)
+            sourcelayerdef = layerdefn(sourcelayer)
+
+            if use_gdal_copy
+                unsafe_copy(sourcelayer; dataset=target, name=getname(sourcelayer))
+            else
+                createlayer(name=getname(sourcelayer),
```
maxfreu, line 262: does it matter whether `unsafe_createlayer` or `createlayer` is used? I think not really, right?

test/test_dataset.jl
```diff
@@ -1,6 +1,16 @@
 using Test
 import ArchGDAL as AG
 
+const supported_vector_drivers = ["FlatGeobuf", "GeoJSON", "GeoJSONSeq", "GML", "GPKG", "JML", "KML", "MapML", "ESRI Shapefile","SQLite"]
```
maxfreu, line 4: I don't know if this and below `assertsimilar` really belong here.

test/test_dataset.jl
```diff
@@ -63,6 +73,15 @@ import ArchGDAL as AG
             end
         end
 
+        for driver in supported_vector_drivers
```
maxfreu, line 76: Here we basically read data, write it to disk, read it back in and compare layer and feature count with the original.

maxfreu:
Seems like some tests are still failing on mac and windows, maybe I have to remove some drivers from the test and rewrite the test a bit.

src/dataset.jl
yeesian: I think my personal preference is to not use `unsafe_` behavior unless it is absolutely required.

src/dataset.jl
```diff
@@ -172,16 +172,133 @@ end
...
+                     layer_options::Vector{String}=[""],
+                     chunksize=20000,
+                     use_gdal_copy=false)
+    create(filename; driver=driver, options=driver_options, width=0, height=0, nbands=0) do target
+        for layeridx in 0:nlayer(dataset)-1
+            sourcelayer = getlayer(dataset, layeridx)
+            sourcelayerdef = layerdefn(sourcelayer)
+
+            if use_gdal_copy
+                unsafe_copy(sourcelayer; dataset=target, name=getname(sourcelayer))
```
yeesian, line 261: Why is `unsafe_copy` being used here? Do we need to `destroy` the object that's returned by it?

yeesian (commented):
I really appreciate the effort here.

In case it's lost among the comments: just wanted to make it clear that I think a single `write` function that "does the right thing" (i.e. writes to file the way a GDAL user would like to) is my preferred approach here. I haven't seen a compelling reason to have separate `writerasters` and `writevectors` function.

src/dataset.jl
maxfreu: Ok, I'll remove it then and just improve the docs of `write` a bit.

src/dataset.jl
```diff
@@ -172,16 +172,133 @@ end
...
     filename::AbstractString;
     kwargs...,
 )::Nothing
+    
+    if nraster(dataset) > 0
+        destroy(unsafe_copy(dataset, filename = filename; kwargs...))
+    elseif nlayer(dataset) > 0
+        writelayers(dataset, filename; kwargs...)
+    elseif nraster(dataset) > 0 && nlayer(dataset) > 0
+        error("Writing datasets with raster and vector data is not supported.")
```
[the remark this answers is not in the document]
maxfreu: Right. That was dumb, I'll pull it up.

src/dataset.jl
```diff
@@ -172,16 +172,133 @@ end
...
+    else
+        error("Could not determine dataset type (raster or vector).")
```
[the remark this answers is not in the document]
maxfreu: I think it should error, because without rasters or layers it's empty, which then probably was by accident.

src/ogr/featuredefn.jl
```diff
@@ -330,7 +330,7 @@ function directly, but use OGRLayer::DeleteGeomField() instead.
...
+function deletegeomdefn!(featuredefn::Union{FeatureDefn, IFeatureDefnView}, i::Integer)
```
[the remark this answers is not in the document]
maxfreu: I didn't think it through, but without this the GPKG workaround will fail.

src/dataset.jl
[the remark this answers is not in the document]
visr: GDAL does support writing empty datasets or not? If I have a vector of unknown size `n` that I want to store in a file, I'd rather not get an error for the empty case, because that may be interesting information in itself. E.g. I can create JSON files with empty arrays just fine.

src/dataset.jl
```diff
@@ -306,7 +423,7 @@ function create(
     width::Integer = 0,
     height::Integer = 0,
     nbands::Integer = 0,
-    dtype::DataType = Any,
+    dtype = GDT_Unknown,
```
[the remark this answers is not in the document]
maxfreu: `create` calls `GDAL.gdalcreate` which expects a `GDALDataType` which is an enum, therefore an int, and not a datatype if I get it right. There seems to be `convert` defined for conversion of `Any` and number types to `GDALDataType. So the way it was is fine, but after this change you can either specify the julia number type or the gdal data type directly. I now added the change to `unsafe_create` as well.

src/dataset.jl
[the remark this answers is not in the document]
maxfreu: Ok, I don't mind removing it.

src/dataset.jl
```diff
@@ -172,16 +172,133 @@ end
...
+and for raster datasets the docs for `writerasters`.
 """
 function write(
     dataset::AbstractDataset,
     filename::AbstractString;
     kwargs...,
 )::Nothing
+    
+    if nraster(dataset) > 0
+        destroy(unsafe_copy(dataset, filename = filename; kwargs...))
```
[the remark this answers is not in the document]
maxfreu: I don't know of any drivers that support both, raster data + `createlayer`, so this should never be the case, correct?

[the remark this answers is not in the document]
visr: I'm not sure if this is what you're asking, but there are several driver names that occur in both https://gdal.org/drivers/raster/index.html and https://gdal.org/drivers/vector/index.html, like `netCDF`. Though that driver is not yet included in the build.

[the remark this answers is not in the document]
maxfreu: Yes, right, but I've never seen a gdal dataset containing raster and vector data. But that's maybe my limited experience. Is there a driver included in the build which supports both? Wouldn't it be good to have netCDF support? And if there are drivers which support both, how could a write function look like? Or should we just skip that case for the time being?

maxfreu:
I just removed the `writerasters` again and implemented the comments you have made. I hope it looks better now :)

src/dataset.jl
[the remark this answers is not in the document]
visr: Yeah I hope we can soon add netCDF support to the build. I haven't seen datasets containing both either so far, so if it is difficult we could leave it for later.

maxfreu:
I just got the tests on windows to pass. The failing tests on mac look unrelated.

src/dataset.jl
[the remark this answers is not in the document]
yeesian: Ohh I see. Can we add a annotation for `dtype`, like `dtype::Union{DataType, GDALDataType}` in that case?

Personally I'd prefer to not expose `GDALDataType` (since it doesn't provide any benefit, it's an extra symbol for the package to maintain, and it confuses the user to have to make that choice), but I don't feel strongly about it.

src/dataset.jl
```diff
@@ -172,13 +172,114 @@ end
...
+`nothing`
 """
 function write(
     dataset::AbstractDataset,
     filename::AbstractString;
     kwargs...,
 )::Nothing
-    destroy(unsafe_copy(dataset, filename = filename; kwargs...))
+    if nraster(dataset) > 0 && nlayer(dataset) > 0
+        error("Writing datasets with raster and vector data is not supported.")
```
yeesian, line 199: If we hit this case, can we print out the driver etc and encourage them to file an issue in this package with the printed information?

src/dataset.jl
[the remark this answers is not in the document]
yeesian: Okay, since we don't yet have a dataset with both rasters and vectors (or know if we can create it), I'm okay with resolving this thread in lieu of https://github.com/yeesian/ArchGDAL.jl/pull/315/files#r943079678. Thank you so much for the discussion!

src/dataset.jl
```diff
@@ -172,13 +172,114 @@ end
...
+                            spatialref=getspatialref(sourcelayer),
+                            options=layer_options) do targetlayer
+
+                    targetlayerdef = layerdefn(targetlayer)
+                    
+                    # work around GPKG driver quirks
+                    # GPKG still has issues with empty geometry field names,
+                    # especially when the field name is empty
+                    if shortname(driver) == "GPKG"
+                        deletegeomdefn!(targetlayerdef, 0)
```
yeesian, line 259: So I think this is technically [not allowed by GDAL](https://github.com/OSGeo/gdal/blob/023d25cf4508e6a4fdae8deaf399bc8fc00825cd/ogr/ogrfeaturedefn.cpp#L839): https://github.com/yeesian/ArchGDAL.jl/blob/04d067e15b2d9fea2987cb2dd2e130f866dda73c/src/ogr/featurelayer.jl#L550-L558.

I'm curious it's working for GPKG but I think it might be worth filing an issue with https://github.com/OSGeo/gdal to ask about the situation with GPKG and see what should be done.

src/ogr/featuredefn.jl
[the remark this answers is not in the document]
yeesian: I might be convinced otherwise, but I doublechecked and don't think this is the way to go; https://github.com/yeesian/ArchGDAL.jl/pull/315/files#r943083917.

src/dataset.jl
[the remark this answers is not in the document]
maxfreu: I think as this change is not related to the vector writer I might as well revert it. The other option would be to just remove the type annotation and set the default to `Any`, but you like to have the type annotations for clarity, right?

src/dataset.jl
maxfreu: Yup, I'll change the info.

src/dataset.jl
maxfreu: I think it's ok to use it in this case, for following practical reasons: 1) The layer is empty by the time this is called, as requested by the docstring. 2) The alternative mentioned in the docstring is not implemented yet. 3) It is used in other places of the gdal codebase, e.g. in the HANA driver. 4) It makes things work for one driver more, where it otherwise wouldn't.

maxfreu: I just found a better workaround by setting the geometry column name to the already source geom col name.

maxfreu:
Ok next round. I improved the GPKG workaround, restored the old definition of `deletefeaturedefn` and changed `GDTUnknown` to `Any`. So I think the last point is whether to add the type annotation for dtype, which I think is maybe not necessary as the name already suggests what it is for.

src/dataset.jl
```diff
@@ -172,13 +172,111 @@ end
...
+            sourcelayerdef = layerdefn(sourcelayer)
+            
+            if shortname(driver) == "GPKG"
+                if !any(Base.contains.(layer_options, "GEOMETRY_NAME"))
+                    if ngeom(sourcelayer) > 0  # for GPKG there can be only one geometry column / layer
+                        geometry_column_name = getname(getgeomdefn(sourcelayerdef, 0))
+                        push!(layer_options, "GEOMETRY_NAME=$geometry_column_name")
+                    end
+                end
+            end
```
yeesian, line 253: This looks much better to me compared to the previous approach -- I'm trying to vet with what's being done in https://github.com/Toblerity/Fiona and why they didn't have similar issues for GPKG for my own general understanding. Can you comment on what's happening with layer options here?

yeesian (commented):
Thank you so much, this is looking close to ready for merge!

I still have some reservations about the approach for GPKG but I might be convinced that this is a strict improvement over the current state of things and just need some time to develop confidence in the correctness of the GPKG workaround (via comparison with other compatibly-licensed GDAL packages).

src/dataset.jl
yeesian: There seems to be a requirement in geopackage:

> The `geometry_type_name` value in a `gpkg_geometry_columns` row SHALL be one of the uppercase geometry type names specified in [Geometry Types (Normative)](https://www.geopackage.org/spec/#geometry_types).

(source: https://www.geopackage.org/spec/#_gpkg_geometry_columns)

src/dataset.jl
yeesian: > Can you comment on what's happening with layer options here?

Okay I think I understand now, sorry it took me this long. My remaining question for this thread (for my understanding) is whether we would have the right behavior for GPKG if `use_gdal_copy=true`? (My guess is "yes", but I'd like to be sure here.)

src/dataset.jl
```diff
@@ -172,13 +172,111 @@ end
...
+                     chunksize=20000,
+                     use_gdal_copy=false)
+    create(filename; driver=driver, options=options) do target
+        for layeridx in 0:nlayer(dataset)-1
+            sourcelayer = getlayer(dataset, layeridx)
+            sourcelayerdef = layerdefn(sourcelayer)
+            
+            if shortname(driver) == "GPKG"
+                if !any(Base.contains.(layer_options, "GEOMETRY_NAME"))
+                    if ngeom(sourcelayer) > 0  # for GPKG there can be only one geometry column / layer
```
yeesian, line 248: If this is true, can we validate that there can be "only one geometry column" and "only one layer" (they are not the same thing) for GPKG?

I.e. we need two checks,

```julia
if shortname(driver) == "GPKG"
    if layeridx > 0
        error("expected nlayer=1 for GPKG driver, got nlayer=$(nlayer(dataset)).")
    end
    ...
```
and
```julia
if ngeom(sourcelayer) > 0
    if ngeom(sourcelayer) != 1
        error("expected ngeom=1 for GPKG driver, got ngeom=$(ngeom(sourcelayer)).")
    end
    ...
```

yeesian: It seems to me that [geopackage can have multiple layers](https://gis.stackexchange.com/questions/287997/counting-the-number-of-layers-in-a-geopackage), in which case I think `push!(layer_options, "GEOMETRY_NAME=$geometry_column_name")` further beyond might introduce multiple entries of `GEOMETRY_NAME=`, and I haven't seen corresponding test cases for how it will behave in that setting.

src/dataset.jl
```diff
@@ -172,13 +172,111 @@ end
...
+Currently working drivers: FlatGeobuf, GeoJSON, GeoJSONSeq, GML, GPKG, JML, KML, MapML, ESRI Shapefile, SQLite
+
+### Parameters
+*`dataset`  The source dataset
+*`filename` The file name to write to
+
+### Keyword arguments
+`driver`           The driver to use, you have to manually select the right driver for the file extension you wish
+`options`   A vector of strings containing KEY=VALUE pairs for driver-specific creation options
+`layer_options`    Driver specific options for layer creation
```
yeesian, line 227: Might be worth mentioning that the same set of `layer_options` will be used for all the layers of the dataset (i.e. we do not allow for them to specify different options for different layers; unless we allow them to pass in a `Dict{Integer, Vector{String}}`).

src/dataset.jl
```diff
@@ -172,13 +172,111 @@ end
...
+
+Currently working drivers: FlatGeobuf, GeoJSON, GeoJSONSeq, GML, GPKG, JML, KML, MapML, ESRI Shapefile, SQLite
+
+### Parameters
+*`dataset`  The source dataset
+*`filename` The file name to write to
+
+### Keyword arguments
+`driver`           The driver to use, you have to manually select the right driver for the file extension you wish
+`options`   A vector of strings containing KEY=VALUE pairs for driver-specific creation options
```
yeesian, line 226: Should this be named `dataset_options` (to distinguish it from `layer_options`)?

src/dataset.jl
yeesian: I looked up [the corresponding implementation in GDAL](https://github.com/OSGeo/gdal/blame/7a4c4903428d289c5a63f51d5175f9564fb5d5d7/gcore/gdaldataset.cpp#L5106-L5447): we should 
1. explain that we haven't [handled the coordination transformations](https://github.com/OSGeo/gdal/blame/7a4c4903428d289c5a63f51d5175f9564fb5d5d7/gcore/gdaldataset.cpp#L5209-L5222)
2. default to `use_gdal_copy=true` (which is the correct; albeit slower implementation), unless we decide to port the GDAL implementation over and clean up the edge cases and make it correct here. We should still maintain the warning that users might encounter errors if they set `use_gdal_copy=false` (but it will be faster).

src/dataset.jl
maxfreu: Separating the layer options for each layer is a good idea I think.

src/dataset.jl
maxfreu: Then you would still need the `options` kwarg for the raster writer or fiddle with the kwargs to rename `dataset_options` to `options`.

src/dataset.jl
maxfreu: Ok I have to sort out the multiple layer case. But I think the "GEOMETRY_NAME" option won't be multiplied as of now, as I check for its presence and only push if it's not there.

src/dataset.jl
yeesian: > I think the basic question is, why it is slower than my implementation. I guess this is due to the low count of features per database transaction (128 vs 20000). So it might be worthwhile to file a feature request with gdal to introduce this as a parameter to `CopyLayer` or so.

Yeah that was what I found too: https://github.com/Toblerity/Fiona/pull/491#discussion_r148936648 and I agree with your assessment about filing the feature request for it to be a parameter rather than hardcoded constant.

> In the meantime, we can cook our own soup and I will:
> 
> 1. explain the missing possibility of coordinate transformations when use_gdal_copy is false in the docstring(before merge)
> 2. set the default of use_gdal_copy to true (before merge)
> 3. work on the julia implementation to get rid of the need for the workaround altogether (after merge?)
> 4. implement the coord transforms in julia (after merge?)

Yeah, that works for me, thank you!

maxfreu:
I just switched the default of `use_gdal_copy` to true, made per-layer options possible (given as vector instead of dict), changed the docs and added some tests.

src/dataset.jl
```diff
@@ -172,13 +172,131 @@ end
...
+* `use_gdal_copy`:    Set this to false (default is true) to achieve higher write speeds at the cost of possible errors. 
+Note that when set to true, no coordinate transformations are possible while writing the features.
+
+### Returns
+nothing
+"""
+function writelayers(dataset::AbstractDataset,
+                     filename;
+                     driver=getdriver(dataset),
+                     options::Vector{String}=[""],
```
yeesian, line 247: nit:
```julia
options = StringList(C_NULL),
```

(It should be untyped and defaulting to NULL for consistency with the rest of the file)

src/dataset.jl
```diff
@@ -172,13 +172,131 @@ end
...
+        error("Layer options not recognized. Please provide a Vector{String} to set the same options for all layers or a
+        Vector{Vector{String}} to set individual options per layer.")
+    end
+    create(filename; driver=driver, options=options) do target
+        for layeridx in 0:nlayer(dataset)-1  # GDAL indexing starts with 0
+            if layer_options isa Vector{Vector{String}}
+                current_layer_options = layeridx+1 <= length(layer_options) ? layer_options[layeridx+1] : [""]
+            else
+                current_layer_options = layer_options
+            end
```
yeesian, line 261: (outside)
```julia
getoptions(options::Dict{T, Vector{String}, i::T) where {T <: Integer} = get(options, i, StringList(C_NULL))
getoptions(options::Vector{String}, i::Integer) = options
```

here
```julia
current_layer_options  = getoptions(layer_options, layeridx)
```

yeesian (changes requested):
LGTM once the remaining comments are addressed.

src/dataset.jl
```diff
@@ -172,13 +172,131 @@ end
...
+    `nothing`
 """
 function write(
     dataset::AbstractDataset,
     filename::AbstractString;
     kwargs...,
 )::Nothing
-    destroy(unsafe_copy(dataset, filename = filename; kwargs...))
+    if nraster(dataset) > 0 && nlayer(dataset) > 0
+        drivername = :driver ∈ keys(kwargs) ? shortname(kwargs[:driver]) : shortname(getdriver(dataset))
```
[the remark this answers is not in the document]
maxfreu: I did not know the get function, thanks for the hint!

src/dataset.jl
```diff
@@ -212,7 +330,7 @@ function unsafe_create(
     width::Integer = 0,
     height::Integer = 0,
     nbands::Integer = 0,
-    dtype::DataType = Any,
+    dtype = Any,
```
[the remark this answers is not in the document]
maxfreu: yup

src/dataset.jl
```diff
@@ -172,13 +172,131 @@ end
...
+* `dataset`  The source dataset
+* `filename` The file name to write to
+
+### Keyword arguments
+* `driver`:           The driver to use, you have to manually select the right driver for the file extension you wish
+* `options`:          A vector of strings containing KEY=VALUE pairs for driver-specific creation options
+* `layer_options`:    Driver specific options for layer creation. The options can either be a Vector{String} to provide the
+same options for each layer, or a Vector{Vector{String}} to provide individual options per layer, in the order of their 
+appearance in the dataset. The strings have to be KEY=VALUE pairs. An example for two layers: 
+`[["FORMAT=WKT", "LAUNDER=NO"], ["STRICT=NO"]]`
```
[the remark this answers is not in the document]
maxfreu: Ah yes, I overlooked the possibility that one might define sth for e.g. the last layer only.

src/dataset.jl
```diff
@@ -172,13 +172,131 @@ end
...
+                     filename;
+                     driver=getdriver(dataset),
+                     options::Vector{String}=[""], 
+                     layer_options=[""],
+                     chunksize=20000,
+                     use_gdal_copy=true)
+    if !(layer_options isa Vector{String}) && !(layer_options isa Vector{Vector{String}})
+        error("Layer options not recognized. Please provide a Vector{String} to set the same options for all layers or a
+        Vector{Vector{String}} to set individual options per layer.")
+    end
```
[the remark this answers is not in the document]
maxfreu: Having it enforced by the signature will generate just one more incomprehensible `MethodError: no method matching...` whereas this prints something the user can directly work with. It looks dirty, but I think it fosters usability.

src/dataset.jl
maxfreu: I made `options` and `layer_options` untyped, but kept the `[""]` default. Using a `StringList(C_NULL)` breaks the GPKG workaround, because there is no `contains` function for `Ptr{Cstring}` and `String`. Exposing a `Ptr{Cstring}` is maybe also a bit unwieldy for users.

maxfreu:
`layer_options` now expects a Dict and I changed some type annotations. Thanks for the comments! Hope it is fine now, but don't hesitate to insist if you still dislike sth.

src/dataset.jl
yeesian: Got it, that makes sense, thank you! My (mild and possibly unfounded) concern was that `[""]` might result in different behavior than `StringList(C_NULL)`, but I agree with you that the GPKG workaround is more important here.

src/dataset.jl
```diff
@@ -172,13 +172,131 @@ end
...
+Note that when set to true, no coordinate transformations are possible while writing the features.
+
+### Returns
+nothing
+"""
+function writelayers(dataset::AbstractDataset,
+                     filename;
+                     driver=getdriver(dataset),
+                     options::Vector{String}=[""], 
+                     layer_options=[""],
```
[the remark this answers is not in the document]
yeesian: Resolved in lieu of https://github.com/yeesian/ArchGDAL.jl/pull/315#discussion_r948035841

src/dataset.jl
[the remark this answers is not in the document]
yeesian: I'm okay with that even if it isn't my preference, thank you for the rationale!

maxfreu:
Has been a pleasure! Thanks for the comments and let's see where this takes us :)

PR #316, opened 2022-08-05
performance and type stability fixes for geometries

rafaqz:
Improves type stability where possible by keeping enums as types in `Val{wkbLineString}` and by forcing child object types from `getgeom`.

~~Also adds a low allocation `getpoint` method.~~

Edit: review when I get tests passing, can't always test on a local machine at present

### DIFF
--- a/src/geointerface.jl
+++ b/src/geointerface.jl
@@ -71,10 +71,10 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
         return getcoorddim(geom)
     end
 
-    GeoInterface.x(::GeometryTraits, geom::AbstractGeometry) = getx(geom, 0)
-    GeoInterface.y(::GeometryTraits, geom::AbstractGeometry) = gety(geom, 0)
-    GeoInterface.z(::GeometryTraits, geom::_AbstractGeometryZ) = getz(geom, 0)
-    GeoInterface.m(::GeometryTraits, geom::_AbstractGeometryM) = getm(geom, 0)
+    GeoInterface.x(::GeoInterface.AbstractPointTrait, geom::AbstractGeometry) = getx(geom, 0)
+    GeoInterface.y(::GeoInterface.AbstractPointTrait, geom::AbstractGeometry) = gety(geom, 0)
+    GeoInterface.z(::GeoInterface.AbstractPointTrait, geom::_AbstractGeometryZ) = getz(geom, 0)
+    GeoInterface.m(::GeoInterface.AbstractPointTrait, geom::_AbstractGeometryM) = getm(geom, 0)
 
     function GeoInterface.getcoord(::GeometryTraits, geom::AbstractGeometry, i)
         if i == 1
--- a/src/geointerface.jl
+++ b/src/geointerface.jl
@@ -62,45 +62,6 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
 
     GeoInterface.isgeometry(::Type{<:AbstractGeometry}) = true
     @enable_geo_plots AbstractGeometry
-
-    function GeoInterface.geomtrait(geom::AbstractGeometry)
-        # TODO Dispatch directly once #266 is merged
-        gtype = getgeomtype(geom)
-        return if gtype in pointtypes
-            GeoInterface.PointTrait()
-        elseif gtype in multipointtypes
-            GeoInterface.MultiPointTrait()
-        elseif gtype in linetypes
-            GeoInterface.LineStringTrait()
-        elseif gtype == wkbLinearRing
-            GeoInterface.LinearRingTrait()
-        elseif gtype in multilinetypes
-            GeoInterface.MultiLineStringTrait()
-        elseif gtype in polygontypes
-            GeoInterface.PolygonTrait()
-        elseif gtype in multipolygontypes
-            GeoInterface.MultiPolygonTrait()
-        elseif gtype in collectiontypes
-            GeoInterface.GeometryCollectionTrait()
-        elseif gtype == wkbCircularString
-            GeoInterface.CircularStringTrait()
-        elseif gtype == wkbCompoundCurve
-            GeoInterface.CompoundCurveTrait()
-        elseif gtype == wkbCurvePolygon
-            GeoInterface.CurvePolygonTrait()
-        elseif gtype == wkbMultiSurface
-            GeoInterface.MultiSurfaceTrait()
-        elseif gtype == wkbPolyhedralSurface
-            GeoInterface.PolyhedralSurfaceTrait()
-        elseif gtype == wkbTIN
-            GeoInterface.TINTrait()
-        elseif gtype == wkbTriangle
-            GeoInterface.TriangleTrait()
-        else
-            nothing
-        end
-    end
-
     GeoInterface.is3d(::GeometryTraits, geom::AbstractGeometry) = is3d(geom)
     function GeoInterface.ismeasured(::GeometryTraits, geom::AbstractGeometry)
         return ismeasured(geom)
@@ -312,4 +273,21 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
         )
         return f(GeoInterface.coordinates(geom))
     end
+
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, pointtypes)...}) = GeoInterface.PointTrait()
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, multipointtypes)...}) = GeoInterface.MultiPointTrait()
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, linetypes)...}) = GeoInterface.LineStringTrait()
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, multilinetypes)...}) = GeoInterface.MultiLineStringTrait()
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, polygontypes)...}) = GeoInterface.PolygonTrait()
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, multipolygontypes)...}) = GeoInterface.MultiPolygonTrait()
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, collectiontypes)...}) = GeoInterface.GeometryCollectionTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbLinearRing}) = GeoInterface.LinearRingTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCircularString}) = GeoInterface.CircularStringTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCompoundCurve}) = GeoInterface.CompoundCurveTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCurvePolygon}) = GeoInterface.CurvePolygonTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbMultiSurface}) = GeoInterface.MultiSurfaceTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbPolyhedralSurface}) = GeoInterface.PolyhedralSurfaceTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbTIN}) = GeoInterface.TINTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbTriangle}) = GeoInterface.TriangleTrait()
 end
+
--- a/src/geointerface.jl
+++ b/src/geointerface.jl
@@ -76,7 +76,7 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
     GeoInterface.z(::GeoInterface.AbstractPointTrait, geom::_AbstractGeometryZ) = getz(geom, 0)
     GeoInterface.m(::GeoInterface.AbstractPointTrait, geom::_AbstractGeometryM) = getm(geom, 0)
 
-    function GeoInterface.getcoord(::GeometryTraits, geom::AbstractGeometry, i)
+    function GeoInterface.getcoord(::GeoInterface.AbstractPointTrait, geom::AbstractGeometry, i)
         if i == 1
             getx(geom, 0)
         elseif i == 2
--- a/src/geointerface.jl
+++ b/src/geointerface.jl
@@ -71,6 +71,11 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
         return getcoorddim(geom)
     end
 
+    GeoInterface.x(::GeometryTraits, geom::AbstractGeometry) = getx(geom, 0)
+    GeoInterface.y(::GeometryTraits, geom::AbstractGeometry) = gety(geom, 0)
+    GeoInterface.z(::GeometryTraits, geom::_AbstractGeometryZ) = getz(geom, 0)
+    GeoInterface.m(::GeometryTraits, geom::_AbstractGeometryM) = getm(geom, 0)
+
     function GeoInterface.getcoord(::GeometryTraits, geom::AbstractGeometry, i)
         if i == 1
             getx(geom, 0)
--- a/src/ogr/geometry.jl
+++ b/src/ogr/geometry.jl
@@ -1068,6 +1068,14 @@ const _AbstractGeometry3d = Union{map(x -> AbstractGeometry{x}, wkbEnums3d)...}
 const _IGeometry3d = Union{map(x -> IGeometry{x}, wkbEnums3d)...}
 const _Geometry3d = Union{map(x -> Geometry{x}, wkbEnums3d)...}
 
+const _AbstractGeometryHasM = Union{_AbstractGeometryM,_AbstractGeometryZM}
+const _IGeometryHasM = Union{_IGeometryM,_IGeometryZM}
+const _GeometryHasM = Union{_GeometryM,_GeometryZM}
+
+const _AbstractGeometryNoM = Union{_AbstractGeometry,_AbstractGeometryZ,_AbstractGeometry25d}
+const _IGeometryNoM = Union{_IGeometry,_IGeometryZ,_IGeometry25D}
+const _GeometryNoM = Union{_Geometry,_GeometryZ,_Geometry25D}
+
 """
     is3d(geom::AbstractGeometry)
 
@@ -1083,6 +1091,8 @@ is3d(geom::_AbstractGeometry3d) = true
 Returns `true` if the geometry has a m coordinate, otherwise `false`.
 """
 ismeasured(geom::AbstractGeometry)::Bool = GDAL.ogr_g_ismeasured(geom.ptr) != 0
+ismeasured(geom::_AbstractGeometryHasM) = true
+ismeasured(geom::_AbstractGeometryNoM) = false
 
 """
     isempty(geom::AbstractGeometry)
--- a/src/ogr/geometry.jl
+++ b/src/ogr/geometry.jl
@@ -103,19 +103,19 @@ end
 
 Returns a copy of the geometry with the original spatial reference system.
 """
-function clone(geom::AbstractGeometry)::IGeometry
+function clone(geom::AbstractGeometry{T}) where T
     if geom.ptr == C_NULL
-        return IGeometry()
+        return IGeometry{wkbUnknown}()
     else
-        return IGeometry(GDAL.ogr_g_clone(geom.ptr))
+        return IGeometry{T}(GDAL.ogr_g_clone(geom.ptr))
     end
 end
 
-function unsafe_clone(geom::AbstractGeometry)::Geometry
+function unsafe_clone(geom::AbstractGeometry{T}) where T
     if geom.ptr == C_NULL
-        return Geometry()
+        return Geometry{wkbUnknown}()
     else
-        return Geometry(GDAL.ogr_g_clone(geom.ptr))
+        return Geometry{T}(GDAL.ogr_g_clone(geom.ptr))
     end
 end
 
@@ -133,6 +133,17 @@ creategeom(geomtype::OGRwkbGeometryType)::IGeometry =
 unsafe_creategeom(geomtype::OGRwkbGeometryType)::Geometry =
     Geometry(GDAL.ogr_g_creategeometry(geomtype))
 
+# When the geometry type `T` is known, pass it wrapped in `Val` for type 
+# stability. `T` is usually equal to `geomtype`, except in the case 
+# of `geomtype == wkbLinearRing`, in which case `T` is `wkbLineString`
+creategeom(::Val{T}) where T = IGeometry{T}(GDAL.ogr_g_creategeometry(T))
+unsafe_creategeom(::Val{T}) where T = Geometry{T}(GDAL.ogr_g_creategeometry(T))
+
+# Special-case createlinearring, because we need to pass
+# wkbLinearRing create but gdal returns a wkbLineString
+creategeom(::Val{wkbLinearRing}) = IGeometry{wkbLineString}(GDAL.ogr_g_creategeometry(wkbLinearRing))
+unsafe_creategeom(::Val{wkbLinearRing}) = Geometry{wkbLineString}(GDAL.ogr_g_creategeometry(wkbLinearRing))
+
 """
     haspreparedgeomsupport()
 
@@ -146,11 +157,11 @@ has_preparedgeom_support() = Bool(GDAL.ogrhaspreparedgeometrysupport())
 Create an prepared geometry of a geometry. This can speed up operations which interact
 with the geometry multiple times, by storing caches of calculated geometry information.
 """
-preparegeom(geom::AbstractGeometry)::IPreparedGeometry =
-    IPreparedGeometry(GDAL.ogrcreatepreparedgeometry(geom.ptr))
+preparegeom(geom::AbstractGeometry{T}) where T =
+    IPreparedGeometry{T}(GDAL.ogrcreatepreparedgeometry(geom.ptr))
 
-unsafe_preparegeom(geom::AbstractGeometry)::PreparedGeometry =
-    PreparedGeometry(GDAL.ogrcreatepreparedgeometry(geom.ptr))
+unsafe_preparegeom(geom::AbstractGeometry{T}) where T =
+    PreparedGeometry{T}(GDAL.ogrcreatepreparedgeometry(geom.ptr))
 
 """
     forceto(geom::AbstractGeometry, targettype::OGRwkbGeometryType, [options])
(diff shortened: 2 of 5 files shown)


yeesian (approved):
It's come such a long way from https://github.com/yeesian/ArchGDAL.jl/pull/160 congratulations y'all! ❤️ 

I haven't spotted anything, but I feel more confident in @evetion than my own judgment at this point!

src/geointerface.jl
```diff
@@ -142,6 +103,14 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
         return getgeom(geom, i - 1)
     end
 
+    function GeoInterface.getpoint(
+        ::GeoInterface.AbstractLineStringTrait,
+        geom::AbstractGeometry{wkbLineString},
+    )
+        refs = Ref{Float64}(), Ref{Float64}(), Ref{Float64}()
+        (getpoint!(geom, i, refs...) for i in (0:GeoInterface.npoint(geom)-1))
+    end
```
yeesian, line 112: Should we add a unit test for this?

src/ogr/geometry.jl
```diff
@@ -1215,42 +1223,70 @@ function getgeom(geom::AbstractGeometry, i::Integer)::IGeometry
...
+        result = GDAL.ogr_g_getgeometryref(geom.ptr, i)
+        result == C_NULL && return IGeometry{wkbUnknown}()
+        return IGeometry{$child}(GDAL.ogr_g_clone(result))
+    end
+    @eval function unsafe_getgeom(geom::AbstractGeometry{$parent}, i::Integer)::Union{Geometry{wkbUnknown},Geometry{$child}}
+        geom.ptr == C_NULL && return Geometry{wkbUnknown}()
+        result = GDAL.ogr_g_getgeometryref(geom.ptr, i)
+        result == C_NULL && return Geometry{wkbUnknown}()
+        return Geometry{$child}(GDAL.ogr_g_clone(result))
     end
```
yeesian, line 1290: The package doesn't have test coverage for this yet.

src/geointerface.jl
```diff
@@ -307,4 +276,21 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
...
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, multipolygontypes)...}) = GeoInterface.MultiPolygonTrait()
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, collectiontypes)...}) = GeoInterface.GeometryCollectionTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbLinearRing}) = GeoInterface.LinearRingTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCircularString}) = GeoInterface.CircularStringTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCompoundCurve}) = GeoInterface.CompoundCurveTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCurvePolygon}) = GeoInterface.CurvePolygonTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbMultiSurface}) = GeoInterface.MultiSurfaceTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbPolyhedralSurface}) = GeoInterface.PolyhedralSurfaceTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbTIN}) = GeoInterface.TINTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbTriangle}) = GeoInterface.TriangleTrait()
```
yeesian, line 294: (optional) Consider adding testing for these to https://github.com/yeesian/ArchGDAL.jl/blob/master/test/test_geometry.jl.

src/geointerface.jl
rafaqz: Yep. I'll actually put it in a separate PR. It doesn't really belong here, I just profiled it all at once.

src/ogr/geometry.jl
rafaqz: I've added tests for these.

src/geointerface.jl
rafaqz: TIN, Triangle, MultiSurface and PolyhedralSurface are really outside the range of my knowledge, I don't actually know how to create them! But everything else will be tested. 

I think we should make an issue to add proper tests for these.

rafaqz:
The failures are an unrelated issue with GDAL_jll on mac. Otherwise this is good to go.

yeesian (dismissed):
Thank you for https://github.com/yeesian/ArchGDAL.jl/pull/316/commits/bee5cb47aa4f7ad36f80263bf36f0f0577528faf !

rafaqz:
@evetion do you have time to review this?

evetion:
> @evetion do you have time to review this?

Just got back from holiday, will take a look soon

test/test_geometry.jl
```diff
@@ -243,10 +243,32 @@ import GeoFormatTypes as GFT
...
+                    atol = 1e-6,
+                )
+                @test AG.toWKT(geom) == "LINESTRING (1 4,2 5,3 6)"
+                AG.closerings!(geom)
+                @test AG.toWKT(geom) == "LINESTRING (1 4,2 5,3 6)"
+                AG.setpoint!(geom, 1, 10, 10)
+                @test AG.toWKT(geom) == "LINESTRING (1 4,10 10,3 6)"
+                @test GFT.val(convert(GFT.WellKnownText, geom)) == AG.toWKT(geom)
+                @test typeof(geom) == AG.Geometry{AG.wkbLineString}
+                # child = AG.unsafe_getgeom(geom, 0)
```
evetion, line 266: Why are these lines commented out?

test/test_geometry.jl
```diff
@@ -261,236 +283,266 @@ import GeoFormatTypes as GFT
...
+                    GI.coordinates(geom),
+                    [[1, 4], [2, 5], [3, 6]],
+                    atol = 1e-6,
+                )
+                @test AG.toWKT(geom) == "LINEARRING (1 4,2 5,3 6)"
+                AG.setpointcount!(geom, 5)
+                @test AG.toWKT(geom) == "LINEARRING (1 4,2 5,3 6,0 0,0 0)"
+                AG.empty!(geom)
+                @test AG.toWKT(geom) == "LINEARRING EMPTY"
+                @test typeof(geom) == AG.Geometry{AG.wkbLineString} # GDAL only uses the LinearRing enum during construction
```
evetion, line 316: Nice improvement!

src/geointerface.jl
```diff
@@ -307,4 +268,21 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
         )
         return f(GeoInterface.coordinates(geom))
     end
+
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, pointtypes)...}) = GeoInterface.PointTrait()
```
evetion, line 272: Nice improvement

src/ogr/geometry.jl
```diff
@@ -133,6 +133,15 @@ creategeom(geomtype::OGRwkbGeometryType)::IGeometry =
 unsafe_creategeom(geomtype::OGRwkbGeometryType)::Geometry =
     Geometry(GDAL.ogr_g_creategeometry(geomtype))
 
+# When the geometry type `T` is known, pass it wrapped in `Val` for type 
+# stability. `T` is usually equal to `geomtype`, except in the case 
+# of `geomtype == wkbLinearRing`, in which case `T` is `wkbLineString`
```
evetion, line 138: So this single exception precludes using geomtype directly, also at `creategeom` at line [130](https://github.com/yeesian/ArchGDAL.jl/pull/316/files#diff-99ff903910c45c3a29f4c99d07cf05347c19c4f9f1119ec5485de8c0b0d39cacR130)?

src/ogr/geometry.jl
```diff
@@ -1215,42 +1224,70 @@ function getgeom(geom::AbstractGeometry, i::Integer)::IGeometry
...
     result = GDAL.ogr_g_getgeometryref(geom.ptr, i)
-    if result == C_NULL
-        return IGeometry()
-    else
-        return IGeometry(GDAL.ogr_g_clone(result))
-    end
+    result == C_NULL && return IGeometry{wkbUnknown}()
+    return IGeometry(GDAL.ogr_g_clone(result))
 end
 function getgeom(
```
evetion, line 1232: This function can be removed now right?

test/test_geometry.jl
```diff
@@ -261,236 +283,266 @@ import GeoFormatTypes as GFT
...
-            [1.0, 2.0, 3.0],
-            [4.0, 5.0, 6.0],
-            [7.0, 8.0, 9.0],
-        ) do geom
-            @test AG.toWKT(geom) == "LINESTRING (1 4 7,2 5 8,3 6 9)"
-            AG.setpoint!(geom, 1, 10, 10, 10)
-            @test AG.toWKT(geom) == "LINESTRING (1 4 7,10 10 10,3 6 9)"
-            AG.addpoint!(geom, 11, 11, 11)
-            @test AG.toWKT(geom) == "LINESTRING (1 4 7,10 10 10,3 6 9,11 11 11)"
+            AG.destroy(geom)
```
evetion, line 286: Good catch

evetion (changes requested):
Lovely PR! I would only ask to improve the test coverage, to include a few tests for the new preparedgeometry types and the empty (unknown) geometries. 

It could also be interesting to sprinkle a few `@inferred` statements, especially around the `getgeom` calls?

Last but not least, could you showcase/tell us about performance improvements you noticed to document this PR? I also see you tried your hand at a simple `getpoint` method, but that didn't work out?

test/test_geometry.jl
rafaqz: Because `unsafe_geom` doesn't work, and I don't know why...

src/ogr/geometry.jl
rafaqz: Yeah. We need to pass both `T` and `geomtype` as in this case they are not the same. GDAL takes a `wkbLinearRing` and returns a `kwbLineString`. Pretty annoying really.

rafaqz:
@evetion `getpoint` works great, it's just a bit unrelated and less finished, so I will PR and test separately. Either we can use GDAL to get all the points eagerly (allocating a vector) or get them lazily one at a time only allocating one point all up. I need to test which is best.

As for performance improvements this is just from getting rid of all the red and yellow bits in ProfileView.jl rasterising GADM.jl output with Rasters.jl.

The before and after included all the changes to Rasters.jl and DimensionalData.jl as well as I was optimising the whole stack... so I'm not totally clear on total times attributable to ArchGDAL. But the fraction in ArchGDAL.jl shrunk a lot from this PR combined with low-allocating `getgeom`.

src/ogr/geometry.jl
rafaqz: Actually not it turns out. But I don't know why `getgeom` works differently for points? is a `wkbPoint` never actually used anywhere?

test/test_geometry.jl
yeesian: > Because `unsafe_geom` doesn't work, and I don't know why...

I have a feeling (i) `getgeom` only works for polygons, geometrycollections, and multi* etc, (ii) getpoint() will work here. But I haven't tested to be sure.

If (ii) is right, maybe we should make getgeom dispatch to getpoint for linestrings?

src/ogr/geometry.jl
```diff
@@ -1501,7 +1528,7 @@ other libraries or applications.
 
 ### Parameters
 * `flag`: `true` if non-linear geometries might be returned (default value).
-          `false` to ask for non-linear geometries to be approximated as linear
+          `false` to ask for non-linear geonetries to be approximated as linear
```
evetion, line 1531: Typo

test/test_geometry.jl
evetion: This is correct, `ogr_g_getgeometryref` doesn't work on linestrings.

yeesian: Yeah I think that why we need the code that was deleted in https://github.com/yeesian/ArchGDAL.jl/pull/316/commits/792fd22d38259c9e3be98b44a54a1bbcc7ad75f7.

rafaqz:
I put together some benchmarks of these changes. There is still another 2x improvment (or more) in fixing `getgeom`, but just type stability gives order of magnitude performance gains for rasterization:

Setup code:
```julia
using Rasters, GADM, BenchmarkTools
kw = (; order=ForwardOrdered(), span=Regular(0.1), sampling=Intervals(Start()), crs=EPSG(4326))
ds = Y(Projected(15.0:0.1:55.0; kw...)), X(Projected(70.0:0.1:140; kw...))
x = GADM.get("CHN")
```

This branch:
```julia
julia> @btime rasterize(x.geom; to=ds, missingval=0, fill=1, boundary=:touches);
  512.476 ms (5890904 allocations: 116.34 MiB)

julia> @benchmark rasterize(x.geom; to=ds, missingval=0, fill=1, boundary=:touches);

julia> @benchmark rasterize(x.geom; to=ds, missingval=0, fill=1, boundary=:touches)
BenchmarkTools.Trial: 8 samples with 1 evaluation.
 Range (min … max):  508.071 ms … 681.143 ms  ┊ GC (min … max): 0.00% … 9.94%
 Time  (median):     675.751 ms               ┊ GC (median):    9.88%
 Time  (mean ± σ):   653.399 ms ±  59.138 ms  ┊ GC (mean ± σ):  8.62% ± 3.44%

  ▁                                                   ▁    ▁▁██
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁████ ▁
  508 ms           Histogram: frequency by time          681 ms <


 Memory estimate: 116.34 MiB, allocs estimate: 5890904.
```


On master:

```julia
julia> @btime rasterize(x.geom; to=ds, missingval=0, fill=1, boundary=:touches);
  7.986 s (29741237 allocations: 939.08 MiB)

julia> @benchmark rasterize(x.geom; to=ds, missingval=0, fill=1, boundary=:touches)
BenchmarkTools.Trial: 2 samples with 1 evaluation.
 Range (min … max):  4.445 s …    5.336 s  ┊ GC (min … max): 4.43% … 4.82%
 Time  (median):     4.890 s               ┊ GC (median):    4.65%
 Time  (mean ± σ):   4.890 s ± 629.958 ms  ┊ GC (mean ± σ):  4.65% ± 0.27%

  █                                                        █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  4.44 s         Histogram: frequency by time         5.34 s <

 Memory estimate: 939.08 MiB, allocs estimate: 29741237.
```

rafaqz:
The abstract types in this PR are getting out of control, because we dont get dimensionality or object type from the enum, we have to use huge `Union` types to define everything in groups we can dispatch on.

It might be nicer to use generic type parameters to indicate the kinds of objects and if they are 2D, 3D or have M fields, and delete all the `Union` groups for e.g. all 3d object types, all M object types etc.

So, something like:
```juila
abstract type AbstractGeometry{E,T,D,M} end
struct Geometry{E,T,D,M} <: AbstractGeometry{E,T,D,M}
    ....
end
struct IGeometry{E,T,D,M} <: AbstractGeometry{E,T,D,M} 
    ...
end
# With default constructors like this:
Geometry{E}(x) where E = Geometry{W,geomtype(W),dimensionality(W),hasm(W)}(x)
```

Then, for example, the types would be:
```julia
Geometry{wkbLineStringZM,LineString,3,HasM}
Geometry{wkbPoint,Point,2,NoM}
```

And we can just use dispatch to match any of the type paramters instead of using so many huge `Union` types.

Any comments on this @yeesian @evetion ?

yeesian:
I think that sounds reasonable; it feels nice to have reached a point where this doesn't feel like over-engineering. I didn't fully follow the example (E.g. would `Geometry{wkbLineStringZM,Polygon,3,NoM}` correspond to a valid `type`?) but it makes sense to have dimensionality, geomtype, etc as type parameters to me.

We might want to think a little more about whether we want to restrain it such the type parameters are independent of each other and all constructed types are valid (versus e.g. allowing for potentially invalid types to be constructed, but having it be easier for implementing the dispatch logic), and the pros and cons of the options, but I think yeah there'll definitely be something which is better than the status quo haha.

rafaqz:
Probably that type shouldn't exist and we could make sure of that in the inner constructor. 

We could also not include the enum in the type at all, and generate it when needed.

yeesian:
> We could also not include the enum in the type at all, and generate it when needed.

I think I like this more. How about something like

```julia
Geometry{wkbLineStringZM,3}
Geometry{wkbPoint,2}
```

And we can generate (i) `LineString` and `HasM` from `(wkbLineStringZM, 3)` and (ii) `Point` and `NoM` from `(wkbPoint,2)`.

rafaqz:
I actually meant the other way round:

```julia
Geometry{LineString,3,HasM}
```

As `wkbLineString` has no use in dispatch, so we would still need all the big unions I have defined.

We can get `wkbLineString` Frome the other three as we don't need it for dispatch.

It might be better to merge this as-is when I fix the tests, and think about this separately. I'm not sure.

yeesian:
> I actually meant the other way round

Ah yes sorry you're right!

> It might be better to merge this as-is when I fix the tests, and think about this separately. I'm not sure.

Yeah I think it'll be a big change regardless of the outcome of the proposal, and agree about thinking about it separately. Happy with this PR once the tests are fixed

rafaqz:
I think this is good to go. 

I found a few existing type-related problems while doing this: `flattento2d!` and `setcoorddim!` etc don't actually work because we don't update the type. Even though its and in place function we will have to use the return value in future. I'll make some issues and fix elsewhere.

rafaqz:
It's still saying changes requested. @evetion do you want to check this and approve if its ok?

src/ogr/geometry.jl
```diff
@@ -1215,42 +1354,74 @@ function getgeom(geom::AbstractGeometry, i::Integer)::IGeometry
...
+    p = getpoint(geom, i)
+    return createpoint(p[1], p[2]; m=getm(geom, i))
+end
+function getgeom(geom::AbstractGeometry{wkbLineString25D}, i::Integer)
+    p = getpoint(geom, i)
+    return createpoint(p[1], p[2], p[3])
+end
+function getgeom(geom::AbstractGeometry{wkbLineStringZM}, i::Integer)
+    p = getpoint(geom, i)
+    return createpoint(p[1], p[2], p[3]; m=getm(geom, i))
```
yeesian, line 1377: Can we add unit tests for these lines?

src/ogr/geometry.jl
```diff
@@ -1584,23 +1746,44 @@ for (f, rt) in ((:create, :IGeometry), (:unsafe_create, :Geometry))
...
-            geom = $f1()
-            addpoint!(geom, coords...)
-            return geom
-        end
+    f1 = Symbol("$(f)point")
+    @eval function $f1(cs::Real...; m=nothing)
+        isnothing(m) ? $f1(cs) : $f1(cs, m)
+    end
+    @eval $f1(coords::Vector) = $f1(Tuple(coords))
+    @eval $f1(coords::Vector, m) = $f1(Tuple(coords), m)
```
yeesian, line 1754: Can we add a unit test for this line?

src/ogr/geometry.jl
```diff
@@ -940,19 +956,143 @@ function empty!(geom::G)::G where {G<:AbstractGeometry}
...
+is3d(geom::AbstractGeometry)::Bool = GDAL.ogr_g_is3d(geom.ptr) != 0 # Is a fallback still needed?
+is3d(geom::_AbstractGeometry2d) = false
+is3d(geom::_AbstractGeometry3d) = true
 
 """
     ismeasured(geom::AbstractGeometry)
 
 Returns `true` if the geometry has a m coordinate, otherwise `false`.
 """
 ismeasured(geom::AbstractGeometry)::Bool = GDAL.ogr_g_ismeasured(geom.ptr) != 0
```
yeesian, line 1093: Can we add a unit test for this line? It seems the current unit tests only covers the new lines.

src/geointerface.jl
```diff
@@ -110,6 +71,11 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
         return getcoorddim(geom)
     end
 
+    GeoInterface.x(::GeometryTraits, geom::AbstractGeometry) = getx(geom, 0)
+    GeoInterface.y(::GeometryTraits, geom::AbstractGeometry) = gety(geom, 0)
+    GeoInterface.z(::GeometryTraits, geom::_AbstractGeometryZ) = getz(geom, 0)
+    GeoInterface.m(::GeometryTraits, geom::_AbstractGeometryM) = getm(geom, 0)
```
yeesian, line 77: Can we add unit tests for these lines?

src/geointerface.jl
```diff
@@ -312,4 +278,21 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
...
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, multipolygontypes)...}) = GeoInterface.MultiPolygonTrait()
+    GeoInterface.geomtrait(geom::Union{map(T -> AbstractGeometry{T}, collectiontypes)...}) = GeoInterface.GeometryCollectionTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbLinearRing}) = GeoInterface.LinearRingTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCircularString}) = GeoInterface.CircularStringTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCompoundCurve}) = GeoInterface.CompoundCurveTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbCurvePolygon}) = GeoInterface.CurvePolygonTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbMultiSurface}) = GeoInterface.MultiSurfaceTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbPolyhedralSurface}) = GeoInterface.PolyhedralSurfaceTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbTIN}) = GeoInterface.TINTrait()
+    GeoInterface.geomtrait(geom::AbstractGeometry{wkbTriangle}) = GeoInterface.TriangleTrait()
```
yeesian, line 296: Can we add unit tests for these lines?

src/types.jl
```diff
@@ -221,41 +221,51 @@ mutable struct ISpatialRef <: AbstractSpatialRef
     end
 end
 
-function _infergeomtype(ptr::GDAL.OGRGeometryH = C_NULL)::OGRwkbGeometryType
-    return if ptr != C_NULL
-        convert(OGRwkbGeometryType, GDAL.ogr_g_getgeometrytype(ptr))
-    else
+_infergeomtype(ptr::GDAL.OGRGeometryH) = wkbUnknown
```
yeesian, line 224: Should this line be removed? It doesn't seem to be covered by unit tests.

yeesian (changes requested):
It'll be good to get the unit test coverage back up -- I've noticed that uncovered lines are historically a rich source of bugs. LGTM otherwise

src/geointerface.jl
rafaqz: The problem here is they were never tested half of these before either, and I don't really know what a `PolyhedralSurface` or `TIN` is, so this is really asking to write proper tests for these types. We do need that, but its not a small amount of work.

yeesian: Oh that's fair, thanks for explaining!

src/ogr/geometry.jl
rafaqz: I've deleted it, it's never hit now.

src/ogr/geometry.jl
rafaqz: Tested the LineString25D, and deleted the others as they were not really supported.

src/types.jl
rafaqz: removed

src/geointerface.jl
rafaqz: All tested except `m`, as I'm not sure we can even construct any `_AbstractGeometryM` objects, and I've pushed that to another PR so this can merge.

rafaqz:
Ok, made a few changes. Measures handling was really not ready so I wound that back a little. This means `GI.m` and `ismeasured` are tested for the `true` case as there isn't an obvious way to make measured objects. 

I've tested most of the types more extensively for `is3d`, for type parameters and child type parameters. The missing lines now (besides `GI.m` and `ismeasured`) seem to be errors in coverage, like constants counted as lines of code that don't run.

test/test_geometry.jl
```diff
@@ -243,17 +265,35 @@ import GeoFormatTypes as GFT
...
+                )
+                @test AG.toWKT(geom) == "LINESTRING (1 4,2 5,3 6)"
+                AG.closerings!(geom)
+                @test AG.toWKT(geom) == "LINESTRING (1 4,2 5,3 6)"
+                AG.setpoint!(geom, 1, 10, 10)
+                @test AG.toWKT(geom) == "LINESTRING (1 4,10 10,3 6)"
+                @test GFT.val(convert(GFT.WellKnownText, geom)) == AG.toWKT(geom)
+            end
+            # unsafe_create
+            geom = AG.unsafe_createlinestring([1.0, 2.0, 3.0], [4.0, 5.0, 6.0])
```
yeesian, line 292: To my understanding, `AG.createlinestring(...)` is going to call `AG.unsafe_createlinestring(...)` [under-the-hood](https://github.com/yeesian/ArchGDAL.jl/blob/fe68c14e10a08a68be70c551a42c260e24c4da48/src/context.jl#L265), so I'm curious why we're duplicating the tests here.

If there is reason for doing so, can we indent all the code between `geom = AG.unsafe_createlinestring(...)` and `AG.destroy(geom)` so that it's clear where the scope of `geom` should be?

(Same comment(s) for `unsafe_createlinearring` below.)

rafaqz: For linestrring hey're actually the same code that swaps in `create` and `unsafe_create` in an `@eval` loop:
https://github.com/rafaqz/ArchGDAL.jl/blob/stability/src/ogr/geometry.jl#L1714-L1720

IMO fine to test both of those.

Linearring actually has completely separate functions, because there is no GDAL enum for `wkbLinearRing25D` So we cant pass the type through, and have to rewrap with the correct `IGeometry{wkbLinearRing25D}`/`Geometry{wkbLinearRing}` 
afterwards. 

Probably they we could get rid of all of that complexity and have `create_x` always just call `unsafe_create_x` and add a finaliser to the result. I would like to get rid of a lot of that code generation, its pretty confusing and means you cant search or grep for method definitions in the code. But maybe thats  another PR.

yeesian: > Probably we could get rid of all of that complexity and have create_x always just call unsafe_create_x and add a finaliser to the result. I would like to get rid of a lot of that code generation, its pretty confusing and means you cant search or grep for method definitions in the code. But maybe thats another PR.

Yeah okay, sounds good; I understand it's all really knotty right now, thank you so much!

yeesian (approved):

yeesian:
I'm going to consider the review by @evetion in https://github.com/yeesian/ArchGDAL.jl/pull/316#pullrequestreview-1072118118 as resolved by https://github.com/yeesian/ArchGDAL.jl/pull/316#issuecomment-1245931594 and the subsequent commits to increase test coverage.

PR #319, opened 2022-08-15
Return extent based on dimensionality.

evetion:
Before we always returned the 3d extent, even on 2d geometries.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c9ce4bd3-c3d5-55b8-8973-c0e20141b8c3"
 keywords = ["GDAL", "IO"]
 license = "MIT"
 desc = "A high level API for GDAL - Geospatial Data Abstraction Library"
-version = "0.9.1"
+version = "0.9.2"
 
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
--- a/src/geointerface.jl
+++ b/src/geointerface.jl
@@ -267,12 +267,18 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
     end
 
     function GeoInterface.extent(::GeometryTraits, a::AbstractGeometry)
-        env = envelope3d(a)
-        return Extent(
-            X = (env.MinX, env.MaxX),
-            Y = (env.MinY, env.MaxY),
-            Z = (env.MinZ, env.MaxZ),
-        )
+        nc = getcoorddim(a)
+        if nc == 2
+            env = envelope(a)
+            return Extent(X = (env.MinX, env.MaxX), Y = (env.MinY, env.MaxY))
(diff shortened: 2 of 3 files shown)


rafaqz:
Lol I have an unfinished PR for this too.

rafaqz:
It should be possible to make this type stable? It could use `is3d`, which could also be type stable.

evetion:
As in having two functions and passing `Val(is3d())` right?

evetion:
Ok, will do, but then we need to wait on #316.

rafaqz:
We don't need to wait tho it doesn't change the types

evetion:
Your PR introduces the AbstractGeometry{T} right? So 

```julia
function GeoInterface.is3d(
        ::GeometryTraits,
        geom::Union{map(T -> AbstractGeometry{T}, ztypes)...},
    )
        return true
    end
```

would require #316

rafaqz:
Yeah you're right. My other PR for this just used a Union in the geometry types before I just changed the abstract types.

I might add `is3d` to the other PR as it falls under type stability improvements. Then this PR can just call `is3d`

evetion:
To be fair, I would opt for using `GI.is3d` here. I like to make all GeoInterface methods compile time based and leave the ArchGDAL functions calling GDAL as is.

rafaqz:
Yeah that sounds like a good idea. 

But also it turns out to be not so easy to do this at all because there is no `wkbPointZ`. So out points are type unstable between 2/3 tuples. We probably need to ecode the dimensionality as an additional parameter in the point type.

evetion:
There's `wkbPoint25D` right? It's confusing, I agree.

rafaqz:
Yeah but no Z when there is ZM? I cant tell how its all mean to work.

evetion:
```
wkbPoint =>    X Y 
wkbPoint25D => X Y Z
wkbPointM =>   X Y M
wkbPointZM =>  X Y Z M
```

rafaqz:
Yeah thats what threw me... there are Z geometries and 25D geometries.

evetion:
If @rafaqz agrees, this can now be merged as is. `is3d` can be sped up in the other PR(s).

rafaqz:
Yep let's do that.

evetion:
Note that I can't merge because we need at least 1 reviewer approving ;)

rafaqz (commented):
LGTM

yeesian (approved):
> Note that I can't merge because we need at least 1 reviewer approving ;)

Sorry I set it up per https://github.com/SciML/ColPrac/blob/b6cacb2942753a0bb6b721c81543b649f9765063/README.md?plain=1#L29

@rafaqz FYI I think you should have the requisite permissions to approve PRs through the github UI too :)

yeesian:
Thank you for this!

PR #324, opened 2022-08-27
Add conversion for Enum types.

evetion:
As a table with an Vector{<:Enum} now can't be written. Enum stores the flags as Integer, so we already have the means to write its values. The only drawback is that we cannot restore the Enum once the value has been written.

### DIFF
--- a/src/ogr/feature.jl
+++ b/src/ogr/feature.jl
@@ -619,6 +619,14 @@ end
 function setfield!(
     feature::AbstractFeature,
     i::Integer,
+    value::Enum,
+)::AbstractFeature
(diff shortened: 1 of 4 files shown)


yeesian (approved):

yeesian:
> The only drawback is that we cannot restore the Enum once the value has been written.

I think that's reasonable? At most they'll have to restore it themselves, like

```julia
julia> ArchGDAL.OGRFieldType(Int64(ArchGDAL.OFTInteger64))==ArchGDAL.OFTInteger64
true
```

evetion:
True, but I wanted to mention it explicitly.

PR #352, opened 2022-12-20
Fix macro callback.

evetion:
We never created the cfunction correctly in the macro, but because it became C_NULL, this was never noticed. (Also it seems no-one uses progress callbacks.). 

Oddly enough, on 1.8.2 the undefined behaviour changed, which started crashing M1 macs.
Probably fixes https://github.com/JuliaLang/julia/issues/47193.


evetion:
CI issues seem to stem from #345

yeesian (approved):
Thank you for this, and the investigation in https://github.com/JuliaLang/julia/issues/47193!

PR #353, opened 2022-12-21
Add Julian progress functions.

evetion:
Should be ever so slightly slower, as some unsafe_conversions are done, but enables a much more Julian interface, using the existing API.


yeesian (approved):
Ah I just this! I think it supersedes #352? Feel free to close that PR if so after this PR is submitted

evetion:
I think the other PR, which this is based on, should be merged, as ArchGDAL (GeoArrays) is not fully functional on Julia >=1.8.2 (can't use Cloud Optimized Geotiff for example).

This PR could warrant some more discussion, because I've tried to be backwards compatible, but that yields some inconsistent naming schemes. We might want to break this part of the API?

src/utils.jl
```diff
@@ -228,20 +228,45 @@ macro cplwarn(code, message)
     end
 end
 
+function showprogress(progress, message = "")
+    print("$(round(Int, progress)*100)..$message")
+    return true
+end
+
+function dummyprogress(progress, message = "")
```
yeesian, line 236: I think we should make this an internal function (I.e. `_dummyprogress`) and document it to always return `true`

Also, can we annotate `progress` with `::Cdouble`?

src/utils.jl
```diff
@@ -228,20 +228,45 @@ macro cplwarn(code, message)
     end
 end
 
+function showprogress(progress, message = "")
+    print("$(round(Int, progress)*100)..$message")
+    return true
+end
```
yeesian, line 234: Where is this function used? Consider removing it if it isn't

src/utils.jl
```diff
@@ -228,20 +228,45 @@ macro cplwarn(code, message)
...
+
+function dummyprogress(progress, message = "")
+    return true
+end
+
+function progressfunc_wrapper(
+    dfComplete::Cdouble,
+    pszMessage::Cstring,
+    pProgressArg::Ptr{Cvoid},
+)::Cint
```
yeesian, line 244: Can we make this an internal function and provide documentation for it? E.g. (based on the documentation for [GDALTermProgress](https://gdal.org/api/cpl.html#_CPPv416GDALTermProgressdPKcPv)),
```julia
"""
    _progresscallback(dfComplete::Cdouble, pszMessage::Cstring, pProgressArg::Ptr{Cvoid})::Cint

The progress callback to be passed to GDAL.

### Parameters
* `dfComplete` – completion ratio from 0.0 to 1.0.
* `pszMessage` – optional message (made available as an argument to the user-defined progress function).
* `pProgressArg` – callback data argument (i.e. user-defined progress function).
"""
```

src/utils.jl
```diff
@@ -228,20 +228,45 @@ macro cplwarn(code, message)
...
+    pProgressArg == C_NULL && return true
+    f = unsafe_pointer_to_objref(pProgressArg)
+    isa(f, Ref) || return true
```
yeesian, line 247: Can you provide a comment above this line explaining it?

src/raster/rasterband.jl
```diff
@@ -283,15 +283,15 @@ function copywholeraster!(
     source::T,
     dest::AbstractRasterBand;
     options = StringList(C_NULL),
-    progressdata = C_NULL,
-    progressfunc::Function = GDAL.gdaldummyprogress,
+    progressfunc::Function = progressfunc_wrapper,
+    progressdata = dummyprogress,
```
yeesian, line 287: Can we consider replacing 
```julia
    progressfunc::Function = progressfunc_wrapper,
    progressdata = dummyprogress,

[...]

        @cplprogress(_progresscallback),
        Ref(progressfunc),
```

with

```julia
    progressfunc::Function = dummyprogress,

[...]

        @cplprogress(_progresscallback),
        Ref(progressfunc),
```

everywhere?

yeesian (changes requested):
Thank you for this PR -- it helped me to understand the whole callback mechanism (which I never did in the past). My comments are just to see if we can simplify the end-user API)

src/raster/rasterband.jl
evetion: Yes, and removing progressdata, fully agree. But it is a breaking API change. Unless we say this was never part of the public API or this never worked anyway.

yeesian: I'm fine with it being a breaking API change

yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/353/commits/24c1304555c48b8e4931b9755e3113f35e312a52

src/utils.jl
yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/353/commits/24c1304555c48b8e4931b9755e3113f35e312a52

src/utils.jl
yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/353/commits/24c1304555c48b8e4931b9755e3113f35e312a52

src/utils.jl
yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/353/commits/24c1304555c48b8e4931b9755e3113f35e312a52

src/utils.jl
yeesian: Resolved by https://github.com/yeesian/ArchGDAL.jl/pull/353/commits/24c1304555c48b8e4931b9755e3113f35e312a52

yeesian (changes requested):
LGTM once tests are passing, thanks! (Trying to look into why they're failing with `_progresscallback` undefined for now; https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#More-About-Callbacks)

evetion:
I'm clearly missing some understanding of the intricacies of cfunction. If we define the cfunction ArchGDAL._cprogresscallback before.

```julia
julia> using ArchGDAL
julia> f = @cfunction(ArchGDAL._progresscallback, Cint, (Cdouble, Cstring, Ptr{Cvoid}))
Ptr{Nothing} @0x00007f7cd4705da0
julia> ccall(Base.unsafe_convert(Ptr{Cvoid}, f), Cint, (Cdouble, Cstring, Ptr{Cvoid}), 1.0, "test", C_NULL)
[ Info: (1.0, Cstring(0x00007f7ce110a238), Ptr{Nothing} @0x0000000000000000)
1

julia> ccall(Base.unsafe_convert(Ptr{Cvoid}, ArchGDAL._cprogresscallback), Cint, (Cdouble, Cstring, Ptr{Cvoid}), 1.0, "test", C_NULL)

signal (11): Segmentation fault
in expression starting at REPL[5]:1
unknown function (ip: 0x7f8bcb10e1e0)
top-level scope at ./REPL[5]:1
Allocations: 3605364 (Pool: 3601633; Big: 3731); GC: 3
```

visr:
I don't really understand this PR or how it should fix the problem that not all architectures are supported. Could you explain?

In your example, `Base.unsafe_convert(Ptr{Cvoid}, f)` does not seem neccessary, `ccall` will do that for us, in a safer manner I believe.

The cfunction docstring mentions

> And that these arguments will be evaluated in global scope during compile-time
  (not deferred until runtime). Adding a '$' in front of the function argument changes this to instead create a
  runtime closure over the local variable callable (this is not supported on all architectures).

So it looks like the cfunctions inside functions like `copywholeraster!` need to have a `$` added in front of the first argument to use the local variable, e.g. `@cfunction($_progresscallback`. But then the not about this not being supported on all architectures applies.

evetion:
This PR is about making user defined progress functions work in a Julian way. Any function that takes a float and string, while returning a bool should work.

src/utils.jl
```diff
@@ -267,15 +268,18 @@ function _progresscallback(
...
-    quote
-        @cfunction(
-            $(Expr(:$, esc(_progresscallback))),
-            Cint,
-            (Cdouble, Cstring, Ptr{Cvoid})
-        )
-    end
-end
+const _cprogresscallback =
+    @cfunction(_progresscallback, Cint, (Cdouble, Cstring, Ptr{Cvoid}))
```
yeesian, line 272: Can we add a comment why this constant is being introduced? Is it due to some nuance of https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#Closure-cfunctions (to prevent something from being finalized)?

evetion:
Ok, after some tryouts, the best thing here is to change GDAL.jl; to change all calls with pData in them from `Ptr{Cvoid}` into `Any`. This will enable ccall to handle the conversion in a safe(r) manner than I have been trying to do with `Ref` and `GC.@preserve`.

Like this, but it requires a change in the generator script. The clue is that it's the `Ptr{Cvoid}` following `pProgressFunc`.

```diff
diff --git a/src/libgdal.jl b/src/libgdal.jl
index 74afada..4f1c71f 100644
--- a/src/libgdal.jl
+++ b/src/libgdal.jl
@@ -2551,7 +2551,7 @@ function gdalcreatescaledprogress(arg1, arg2, arg3, arg4)
         ccall(
             (:GDALCreateScaledProgress, libgdal),
             Ptr{Cvoid},
-            (Cdouble, Cdouble, GDALProgressFunc, Ptr{Cvoid}),
+            (Cdouble, Cdouble, GDALProgressFunc, Any),
             arg1,
             arg2,
             arg3,
@@ -4567,7 +4567,7 @@ function vsisync(
                 Cstring,
                 Ptr{Cstring},
                 GDALProgressFunc,
-                Ptr{Cvoid},
+                Any,
                 Ptr{Ptr{Cstring}},
             ),
             pszSource,
```

Works locally on both Linux and M1, so no platform differences.

yeesian:
> Ok, after some tryouts, the best thing here is to change GDAL.jl; to change all calls with pData in them from `Ptr{Cvoid}` into `Any`.

I was curious about the claim for this before I found it in https://julialang.org/blog/2013/05/callback/#passing_closures_via_pass-through_pointers.

evetion:
Yeah that has been my reference indeed. Crazy to see it's now 10 years old, but the Any conversion actually works and is the cleanest solution.

My own tryouts with Ref always led to segmentation faults for some cases, even though it worked 99% of the time (100% of the time on my M1, which threw me off for a while). So my guess would be that the GC cleaned up my Ref or even the function pointer. Note that you can't take a pointer to a Function directly in Julia (immutable things are moved around).

I'll make a PR for GDAL. Happy new year 🎉

yeesian (dismissed):
LGTM, thanks for seeing this all the way through!

My remaining nit is to include the function signature of `progressfunc` inside the respective docstrings, otherwise end-users don't have a way to know what types of functions would be valid.

I guess one downside of this approach is users can't plug in GDALTermProgress, but I guess we can provide our own port (or better, maybe e.g. via progress meter.jl) in the future

evetion:
> My remaining nit is to include the function signature of progressfunc inside the respective docstrings, otherwise end-users don't have a way to know what types of functions would be valid.

I've updated the docstrings. Feel free to merge if you're happy with it.

> I guess one downside of this approach is users can't plug in GDALTermProgress, but I guess we can provide our own port (or better, maybe e.g. via progress meter.jl) in the future

You actually can! Although it starts to feel like pingpong a bit:
```julia
progressfunc = (progress, message) -> GDAL.gdaltermprogress(progress, message, C_NULL)
0...10...20...30...40...50...60...70...80...90...100 - done.

yeesian (approved):

PR #356, opened 2022-12-30
Fix tests for GDAL 3.6 release.

evetion:
Also adds a first changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,29 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+
+- Added CHANGELOG.md
+
+### Fixed
+
+
+### Changed
+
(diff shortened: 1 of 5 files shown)


evetion:
Might be nice to merge #353 (breaking) before we release this 0.10?

visr:
I'd say so, unless it would take much longer.

Can you remove GDAL_jll from the deps again, now that the compat is removed? Perhaps it would be good to add https://github.com/JuliaTesting/Aqua.jl/ to the tests, since it will complain about deps with missing compat.

Otherwise, looks good to me, thanks for working on this.

visr:
Looks like the benchmark cannot run since the two versions it wants to compare need different GDAL_jll versions

```
ERROR: LoadError: Unsatisfiable requirements detected for package GDAL_jll [a7073[274](https://github.com/yeesian/ArchGDAL.jl/actions/runs/3807712715/jobs/6477652583#step:6:275)]:
 GDAL_jll [a7073274] log:
 ├─possible versions are: 3.0.3-301.600.100 or uninstalled
 ├─restricted to versions 300.400-300.500 by ArchGDAL [c9ce4bd3], leaving only versions 300.400.100-300.500.200
 │ └─ArchGDAL [c9ce4bd3] log:
 │   ├─possible versions are: 0.9.4 or uninstalled
 │   └─ArchGDAL [c9ce4bd3] is fixed to version 0.9.4
 └─restricted to versions 301.600.100 by an explicit requirement — no versions left
```

visr (approved):

PR #359, opened 2023-01-05
[breaking] switch from Base.convert to GeoInterface.convert

visr:
This is a breaking change.

For background see https://github.com/JuliaGeo/GeoInterface.jl/pull/66. Brought up in https://github.com/yeesian/ArchGDAL.jl/issues/189#issuecomment-1368537576.

Soon https://github.com/JuliaGeo/GeoInterface.jl/pull/85 will provide extra options for conversion, though this doesn't help ArchGDAL as much, since you always convert to `IGeometry` anyway. And it is not breaking so no need to wait for that.

After this is in, shall we tag v0.10?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - breaking: Updated compat for GDAL.jl 1.5 and fixed tests for GDAL 3.6. These include changes to `gdalnearblack`, `fillunsetwithdefault!` and `gdalgetgeotransform`.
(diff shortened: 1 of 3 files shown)


yeesian (approved):
Thank you!

yeesian:
Added "[breaking]" to the title so we might get reminded of it when tagging for the next release: https://github.com/yeesian/ArchGDAL.jl/compare/v0.9.4...master

visr:
Thanks, yes I also added a "breaking" label that helps with querying PRs.

What do you think, should we tag 0.10 now? Would be nice to have GDAL 3.6 available.

yeesian:
https://github.com/yeesian/ArchGDAL.jl/commit/dd34eedd7605d350811f614adcebe20cd15c3f11#commitcomment-95375145

PR #362, opened 2023-01-14
implement module convert for geometries

rafaqz:
This allows us to do `GeoInterface.convert(ArchGDAL, geom)` with any GeoInterface.jl compatible geometry.


yeesian (approved):
Adding @visr as a second pair of eyes just in case I'm missing anything.

visr (approved):

PR #364, opened 2023-01-18
relax type constraint in unsafe_executesql

maxfreu:
Hi! I just switched the spatialfilter type constraint to `AbstractGeometry`, so that one can use `IGeometry{wkbPolygon}` as filter. I hope this is valid; at least it seems to work.


yeesian (approved):
Thanks! Yeah that looks reasonable to me :)

PR #365, opened 2023-01-19
Allow `GeoInterface.convert(ArchGDAL, obj)` to "just work"

asinghvi17:
This doesn't work for FeatureCollections yet, and no tests have been added.  It's essentially a quick and dirty implementation where I extended the type constraints on the main `GeoInterface.convert` method and implemented `ArchGDAL.geointerface_geomtype(arg) = IGeometry`.  Since it operates on traits, it seemed a fair bet, and any feature-like object can of course be overloaded.


src/geointerface.jl
```diff
@@ -353,4 +357,7 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
     function GeoInterface.geomtrait(geom::AbstractGeometry{wkbTriangle})
         return GeoInterface.TriangleTrait()
     end
+
+    geointerface_geomtype(trait::GeometryTraits) = IGeometry
+
```
asinghvi17, line 362:
```suggestion
```
Whoops, didn't notice that this was already implemented in the same way at the top of the file.  Can be deleted.

rafaqz: Yeah, it already works ;)

rafaqz:
This is already tested as working. But yes we also need to handle features somehow. I guess I should have handle that in GeoInterface as well.

rafaqz:
A test ?

asinghvi17:
Closing since we don't want to be doing this anymore

PR #366, opened 2023-01-26
convert any geometries to gdal types in geometry methods

rafaqz:
Add methods to operations like `disjoint`, `crosses` etc to accept arbitrary GeoInterface.jl compatible geometries and convert them with the shared `to_gdal` method and `geointerface_geomtype`.

We don't use `convert(ArchGDAL, geom)` here as it has a small overhead to call a pacage method, which is unnecessary to pay locally - especially for points of anything there are a lot of.

Needs tests.


yeesian (approved):
Thank you for the PR description -- it answered my questions. There doesn't seem to be a way to look at code coverage delta, but LGTM apart from maintaining coverage there. I'll also ask for approval from @visr @evetion since they have context from https://github.com/JuliaGeo/GeoInterface.jl/pull/78

src/ogr/geometry.jl
```diff
@@ -1,3 +1,15 @@
...
+# `geom` can be points, so this is a hot path that needs to be fast.
+function to_gdal(geom)
+    GeoInterface.isgeometry(geom) || _not_a_geom_error()
+    trait = GeoInterface.geomtrait(geom)
+    typ = geointerface_geomtype(trait)
+    return GeoInterface.convert(typ, trait, geom)
+end
+
+@noinline _not_a_geom_error() =
+    throw(ArgumentError("Ojbect of type $(typeof(geom)) is not a GeoInterface compatible geometry"))
```
yeesian, line 12: Ojbect -> Object

src/ogr/geometry.jl
```diff
@@ -848,43 +949,53 @@ if the difference is empty.
...
 """
-    geomlength(geom::AbstractGeometry)
+    geomlength(geom)
 
 Returns the length of the geometry, or 0.0 for unsupported geometry types.
 """
 geomlength(geom::AbstractGeometry)::Float64 = GDAL.ogr_g_length(geom)
+geomlength(geom) = geomlength(to_gdal(geom))
+# TODO Add `length` method to match GeoInterface naming conventions?
+# this will mean using `Base.length` everywhere
```
yeesian, line 987: Should we turn this into a github issue?

test/test_geos_operations.jl
```diff
@@ -176,91 +206,145 @@ using Extents
...
+            AG.convexhull(AG.fromWKT("POLYGON((0 0, 10 0, 10 10, 0 10, 0 0))")),
+        ) == "POLYGON ((0 0,0 10,10 10,10 0,0 0))"
+
+        # @test AG.asbinary(
+        #     AG.fromWKT("POLYGON((0 0, 10 0, 10 10, 0 10, 0 0))"),
+        # )[1:10] ==
+        #       UInt8[0x01, 0x03, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x05]
+
+        # @test AG.astext(AG.fromWKT("POLYGON((0 0, 10 0, 10 10, 0 10, 0 0))")) ==
+        #       "POLYGON ((0 0,10 0,10 10,0 10,0 0))"
```
yeesian, line 344: Out of curiosity, why are these commented out?

rafaqz: Oh because they dont exist, but maybe should to match geointerface

yeesian (changes requested):
I just found the report of the coverage -- https://app.codecov.io/gh/yeesian/ArchGDAL.jl/pull/366 and it seems a good number of methods still needs coverage

test/test_geos_operations.jl
yeesian: Oh thanks for explaining!

rafaqz:
Yeah im slowly working throufh them

test/test_geos_operations.jl
rafaqz: I'll add a comment and a github issue, it's basically the same as for length

PR #372, opened 2023-03-24
Introduce `importUSER` for setting SRS from custom strings.

evetion:
Now we can do:

```julia
julia> p = ArchGDAL.importUSER("EPSG:4326+3855")
Spatial Reference System: +proj=longlat +datum=WGS84 +geoidgr ... _defs
```

Note how there's now `+geoidgrid` present, which represents the vertical transform of EPSG:3855.

### DIFF
--- a/src/spatialref.jl
+++ b/src/spatialref.jl
@@ -312,7 +312,7 @@ unsafe_importEPSG(code::Integer; kwargs...)::SpatialRef =
     importEPSG!(unsafe_newspatialref(; kwargs...), code)
 
 """
-    importUSER(code::AbstractString; [order=:compliant])
+    importUserInput(code::AbstractString; [order=:compliant])
 
 Construct a Spatial Reference System from a user provided code that is parsed by GDAL.
 This is useful when the input code is in an unknown format or a shortcut not covered by more constrained methods.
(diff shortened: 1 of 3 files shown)


src/spatialref.jl
```diff
@@ -311,6 +311,33 @@ importEPSG(code::Integer; kwargs...)::ISpatialRef =
...
+Construct a Spatial Reference System from a user provided code that is parsed by GDAL.
+This is useful when the input code is in an unknown format or a shortcut not covered by more constrained methods.
+An example is the code "EPSG:4326+3855", the shortest way to describe a combination of a horizontal and vertical crs.
+
+# Keyword Arguments
+- `order`: Sets the axis mapping strategy. `:trad` will use traditional lon/lat
+    axis ordering in any actions done with the crs. `:compliant`, will use axis
+    ordering compliant with the relevant CRS authority.
+"""
+importUSER(code::AbstractString; kwargs...)::ISpatialRef =
```
yeesian, line 326: Can we name it from `importUSER` to `importUserInput` instead? That makes it unambiguous that `USER` isn't an acronym

yeesian (changes requested):
Overall LGTM apart from bike-shedding on function name(s)

yeesian (approved):
Thank you!

PR #374, opened 2023-03-24
Parse GeoFormatTypes EPSG with 2 codes

evetion:
To be merged after #372 and https://github.com/JuliaGeo/GeoFormatTypes.jl/pull/22


yeesian (approved):
Looks great, thank you! Out of curiosity: is there any way to look up documentation for EPSG with 2 codes?

evetion:
Not currently no. It would be a compound CRS, and you can lookup the individual codes. Like the test examples, it would be a combination of a 2D CRS and a vertical one, making it 3d. Some of those combinations are common, and have their own code: 4326+3855 == 9518.

Maybe the Proj.identify function could help with this, but it doesn't yet know about the EPSG{2} convention.

PR #379, opened 2023-04-28
fix inference on GI.getcoord

rafaqz:
Throw an error on wrong index to `getcoord` instead of returning `nothing`. This was pretty bad for inference and ran into bugs in base and other poorly inferenced functions in Meshes.jl to cause hangs.

@ErickChacon this should fix the other half of your inference bug. Thanks for persevering to find problems in both packages and Base!

### DIFF
--- a/src/geointerface.jl
+++ b/src/geointerface.jl
@@ -87,21 +87,23 @@ let pointtypes = (wkbPoint, wkbPoint25D, wkbPointM, wkbPointZM),
     end
     function GeoInterface.z(
         ::GeoInterface.AbstractPointTrait,
-        geom::_AbstractGeometryZ,
+        geom::AbstractGeometry,
     )
+        geom isa _AbstractGeometry3d || throw(ArgumentError("Geometry does not have `Z` values")) 
         return getz(geom, 0)
     end
     function GeoInterface.m(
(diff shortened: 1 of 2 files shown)


visr:
Looks fine to me. Could possibly be seen as a breaking change. But GeoInterface docs doesn't say much about what to return when a coordinate doesn't exists and this seems sensible. Only that for shorter tuples and vectors it throws an ArgumentError. I see `GI.z` and `GI.m` also return nothing for XY geometries, should that also be changed directly?

rafaqz:
Yeah it should always error instead of `nothing` then the conpiler has clean Float64

yeesian:
Thanks for this! LGTM modulo tests passing

rafaqz:
Turns out half the problem was probably our horrific fallback for `getz` in GeoInterface.jl could be hit because of a slightly wrong type used in the dispatch types in `GI.getz` here. It was yet another source of `nothing`.

The fix means `getcoord` only accepts `Integer` now where before it would pass through `Nothing`. `GI.z(geom)` will also error if there is no `z` now, rather than returning nothing.

So maybe its a breaking change, IDK.

visr (approved):
Looks good! I think this is fine to merge.

yeesian (approved):

PR #381, opened 2023-06-02
GeoPackage test no longer errors

visr:
This is all that is required to fix tests after https://github.com/JuliaGeo/GDAL.jl/pull/153

Don't really understand, but it now only errors on `use_gdal_copy = false`, whereas the default, true, does not fail.


yeesian (approved):

maxfreu:
Maybe support for writing fields of type real has been added to the GPKG driver and gdal knows which settings to set correctly, and we're not setting it in the custom write or so?

PR #382, opened 2023-06-09
factor out layer copying into its own function

maxfreu:
Hi! I have factored out the layer copying from writelayers into its own function. So basically I changed (pseudocode):

```julia
function writelayers()
  create() do
    copystuff
  end
end
```

to 
```julia
function writelayers()
  create() do
    copylayers(src, dst)
  end
end

function copylayers(src, dst, ...)
  copystuff
end
```
That should now allow to copy data from and to arbitrary datasets, whether on disk or not. You can also give the layer indices you want to be copied. I added a tiny test for that. Furthermore I adjusted the docstrings for nicer rendering in the REPL. I'm not sure with the naming `copylayers`, is it ok? What do you think in general? Lastly, this supersedes https://github.com/yeesian/ArchGDAL.jl/pull/381, but I can take that part out if you want to merge this before the change to GDAL.

### DIFF
--- a/src/dataset.jl
+++ b/src/dataset.jl
@@ -307,7 +307,7 @@ Dict{<:Integer, Vector{String}} to set individual options per layer.",
         sourcelayer = getlayer(source_dataset, layeridx)
         sourcelayerdef = layerdefn(sourcelayer)
 
-        if shortname(getdriver(source_dataset)) == "GPKG"
+        if shortname(getdriver(target_dataset)) == "GPKG"
             if !any(Base.contains.(current_layer_options, "GEOMETRY_NAME"))
                 if ngeom(sourcelayer) > 0  # for GPKG there can be only one geometry column per layer
                     geometry_column_name =
--- a/src/dataset.jl
+++ b/src/dataset.jl
@@ -180,20 +180,21 @@ Writes the dataset to the designated filename.
 * `filename`: The filename, UTF-8 encoded.
 
 ### Keyword Arguments
-* `driver` (ArchGDAL.Driver): The driver to use, you have to manually select the right driver via `getdriver(drivername)` matching the file extension you wish.
-Otherwise the driver of the source dataset will be used.
+* `driver` (ArchGDAL.Driver): The driver to use, you have to manually select the right driver via `getdriver(drivername)`
+                              matching the file extension you wish. Otherwise the driver of the source dataset will be used.
 * `options` (Vector{String}): A vector of strings containing KEY=VALUE pairs for driver-specific creation options.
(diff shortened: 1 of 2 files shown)


src/dataset.jl
```diff
@@ -255,8 +257,36 @@ function writelayers(
...
+"""
+    copylayers(source_dataset, target_dataset; kwargs...)
+
+Copies a (sub)set of layers from source to target dataset. The keyword arguments are the same as for `writelayers`.
+Note that eventual layer options have to conform with the target dataset driver.
+"""
+function copylayers!(
+    source_dataset::AbstractDataset,
+    target_dataset::AbstractDataset;
+    layers = 0:(nlayer(source_dataset)-1),
```
yeesian, line 288: Can we rename it from `layers` to `source_layer_indices` (or `source_layer_index` if it's `copylayer!`)?

src/dataset.jl
```diff
@@ -255,8 +257,36 @@ function writelayers(
...
+    return nothing
+end
+
+"""
+    copylayers(source_dataset, target_dataset; kwargs...)
+
+Copies a (sub)set of layers from source to target dataset. The keyword arguments are the same as for `writelayers`.
+Note that eventual layer options have to conform with the target dataset driver.
+"""
+function copylayers!(
```
yeesian, line 285: In the spirit of factoring things out -- should we make this `copylayer!(source_dataset, target_dataset, source_layer_index)` and letting the user write a loop over the layers they want to copy?

src/dataset.jl
```diff
@@ -255,8 +257,36 @@ function writelayers(
...
-    options = [""],
-    layer_options = [""],
+    options = String[],
+    layers = 0:(nlayer(dataset)-1),
+    layer_options = String[],
+    chunksize::Integer = 20000,
+    use_gdal_copy::Bool = true,
+)
+    create(filename; driver = driver, options = options) do target
+        return copylayers!(
```
yeesian, line 267: Can we add unit test(s) for `copylayer(s)!` without depending on it being called through `writelayers` in the unit tests?

yeesian (changes requested):
Overall LGTM (modulo the changes requested), thank you!

src/dataset.jl
maxfreu: I think that would not give the user any benefit, so I'd keep it here.

maxfreu:
I've just made the requested changes except for one :)

yeesian (approved):
@visr can I get your review here since it supersedes https://github.com/yeesian/ArchGDAL.jl/pull/381?

visr (approved):
Looks good to me, thanks!

maxfreu:
Ok, done - sorry for the forth and back!

visr (approved):

PR #391, opened 2023-07-20
add support for converting to WKT to EPSG

alex-s-gardner:


.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
yeesian, line 1: Can we revert the inclusion of this file?

src/spatialref.jl
```diff
@@ -600,6 +600,8 @@ Convert this SRS into a nicely formatted WKT string for display to a person.
 function toWKT(spref::AbstractSpatialRef, simplify::Bool)::String
     wktptr = Ref{Cstring}()
     result = GDAL.osrexporttoprettywkt(spref, wktptr, simplify)
+
+    println(restul)
```
yeesian, line 604: Please remove print statements

src/spatialref.jl
```diff
@@ -616,6 +618,31 @@ function toPROJ4(spref::AbstractSpatialRef)::String
...
+    result = GDAL.osrgetauthorityname(spref.ptr, "PROJCS")
+    projcs = "PROJCS"
+
+    if isnothing(result)
+        result = GDAL.osrgetauthorityname(spref.ptr, "GEOGCS")
+        projcs = "GEOGCS"
+    end
+
+    if isnothing(result)
+        error("No PROJCS or GEOGCS Authority found")
```
yeesian, line 636: Would people prefer an empty string, or an error in this case?

yeesian (changes requested):
Thank you!

src/spatialref.jl
alex-s-gardner: I was thinking the same thing but passing an empty string or "EPSG:" to GFT.EPSG() will throw an error

alex-s-gardner:
a more full featured retrieval of EPSG will require a pyproj like to_epsg:

https://github.com/JuliaGeo/Proj.jl/issues/88

src/spatialref.jl
rafaqz: `GFT.EPSG` actually holds an integer internally, so the constructor checks that a string is `epsg:NNNN` and parses then Ns.

src/spatialref.jl
```diff
@@ -616,6 +618,31 @@ function toPROJ4(spref::AbstractSpatialRef)::String
...
 end
 
+"""
+    toEPSG(spref::AbstractSpatialRef)
+
+If possible, export EPSG code for this coordinate system.
+"""
+function toEPSG(spref::AbstractSpatialRef)::String
+    result = GDAL.osrgetauthorityname(spref.ptr, "PROJCS")
+    projcs = "PROJCS"
```
rafaqz, line 628: Its confusing defining `projcs` here - why not as an `else` path of the conditional below?

alex-s-gardner: now moved

src/spatialref.jl
alex-s-gardner: so I think our only options are [1] throw an error or [2] set epsg = 0

.vscode/settings.json
alex-s-gardner: done

src/spatialref.jl
yeesian: I'm okay with keeping the current/proposed behavior of raising an error, so I'll resolve this -- but feel free to re-open or create a ticket for discussion if it remains controversial.

src/convert.jl
```diff
@@ -112,10 +112,16 @@ end
 function unsafe_convertcrs(::Type{<:GFT.ProjString}, crsref)
     return GFT.ProjString(toPROJ4(crsref))
 end
+function unsafe_convertcrs(::Type{<:GFT.EPSG}, crsref)
```
yeesian, line 115: Can we add a test for this method? It seems to not be covered per https://app.codecov.io/gh/yeesian/ArchGDAL.jl/pull/391.

yeesian (dismissed):

src/convert.jl
evetion: This method is overwritten at line 124, so we need to remove the duplicate.

src/spatialref.jl
```diff
@@ -616,6 +617,31 @@ function toPROJ4(spref::AbstractSpatialRef)::String
     return unsafe_string(projptr[])
 end
 
+"""
+    toEPSG(spref::AbstractSpatialRef)
+
+If possible, export EPSG code for this coordinate system.
+"""
+function toEPSG(spref::AbstractSpatialRef)::String
```
evetion, line 625: EPSG is a code, so this should return an Int? GFT.EPSG(1) does also work, so it should pose no further changes.

src/spatialref.jl
```diff
@@ -616,6 +617,31 @@ function toPROJ4(spref::AbstractSpatialRef)::String
     return unsafe_string(projptr[])
 end
 
+"""
+    toEPSG(spref::AbstractSpatialRef)
+
+If possible, export EPSG code for this coordinate system.
```
evetion, line 623:
```suggestion
Export EPSG code for this coordinate system if available.
```

src/spatialref.jl
```diff
@@ -616,6 +617,31 @@ function toPROJ4(spref::AbstractSpatialRef)::String
...
+        projcs = "GEOGCS"
+    end
+
+    if isnothing(result)
+        error("No PROJCS or GEOGCS Authority found")
+    elseif result == "EPSG"
+            epsg = GDAL.osrgetauthoritycode(spref.ptr, projcs)
+        return "EPSG:$(epsg)"
+    else
+        @ogrerr result "is not an EPSG authority"
```
evetion, line 641: The use of the @ogrerr macro only triggers if GDAL itself failed. Ideally you use this after each call to `osrgetauthoritycode` and `osrgetauthorityname`. Here you probably just want to use error().

evetion (changes requested):

src/spatialref.jl
alex-s-gardner: I'm not sure what we want to do here... changing this to an Int won't be a prblem for GFT but are there other applicaitons that use crs text?

evetion: Well you introduced this function here, so no one uses it yet.

alex-s-gardner: Now updated to return an Int32

src/convert.jl
alex-s-gardner: now fixed

evetion (approved):

PR #395, opened 2023-09-22
fix tests with GDAL 3.7.2

visr:
This fixes issues with [GDAL 3.7.2](https://github.com/OSGeo/gdal/blob/v3.7.2/NEWS.md). Nothing major, see individual commit messages for details. This will only pass after https://github.com/JuliaGeo/GDAL.jl/pull/161 is merged.

### DIFF
--- a/test/test_feature.jl
+++ b/test/test_feature.jl
@@ -2,8 +2,8 @@ using Test
 import ArchGDAL as AG
 import GeoInterface as GI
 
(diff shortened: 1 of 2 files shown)


visr:
This actually works both before and after https://github.com/JuliaGeo/GDAL.jl/pull/161. Should be good to go.

yeesian (approved):
Thank you!

PR #405, opened 2023-12-21
Implement GeoInterface.crs

evetion:
As found here https://discourse.julialang.org/t/retrieving-crs-from-vector-and-raster-data/107903/4
One can now do `GeoInterface.crs` on a geometry, layer and dataset.

Also implements a convert method to GFT from an AbstractSpatialRef.


rafaqz (approved):
LGTM

PR #411, opened 2024-02-01
Make buffer for `read!` an AbstractArray

### LINKED ISSUE #410: Type of buffer in `read!` is too restricted
When I try to plot a Raster which is based on an ArchGDAL overview array this does not work because I can't `read!` into a PermutedDims array. See https://github.com/rafaqz/Rasters.jl/issues/596.

Can we expand the type that `read!` accepts so that we could read also into AbstractMatrix and not only Matrix?

felixcremer:
This widens the dispatch of the `read!` and `rasterio!` functions so that the buffer can be an AbstractArray instead of a normal Array. This will allow to also read into for example views or other array wrapper types like PermuteDimsArrays. 

This would close #410. 
It will also close https://github.com/rafaqz/Rasters.jl/issues/596

Should we also widen the buffer type for the write! functions?

### DIFF
--- a/src/raster/rasterio.jl
+++ b/src/raster/rasterio.jl
@@ -1,12 +1,12 @@
 
 """
-    rasterio!(dataset::AbstractDataset, buffer::Array{<:Any, 3},
+    rasterio!(dataset::AbstractDataset, buffer::AbstractArray{<:Any, 3},
         bands; <keyword arguments>)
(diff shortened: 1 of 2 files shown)


felixcremer:
The failure on Julia nightly is due to a change in Julia Base removing length from Iterators.Stateful which is used in DiskArrays. I am going to open an issue there. https://github.com/JuliaLang/julia/pull/51747

yeesian (approved):
Thank you for this, adding the tests and investigating the failures!

PR #415, opened 2024-02-15
Forward gdalinfo for RasterDataset

felixcremer:
This allows to use AG.gdalinfo on  a RasterDataset.


yeesian (approved):
Sorry I missed this for so long

yeesian (approved):

PR #418, opened 2024-03-04
Create new minor release

yeesian:


yeesian:
Should we make one last minor release for ArchGDAL (i.e. this PR) to `v0.10.3` before I make another release of ArchGDAL to `v0.11.0` that will only supports 1.9+ (that can benefit from the effects system)?

rafaqz:
> Thay makes sense to do. Will need DiskArrays dropped back to 0.3

PR #421, opened 2024-04-05
Check if a geometry is a linear ring before assigning trait

asinghvi17:


asinghvi17:
There is a wkbLinearRing type, but for some reason
```julia
julia> typeof(GI.getgeom(water1, 1))
ArchGDAL.IGeometry{ArchGDAL.wkbLineString}
```
returns a linestring, even though it's a polygon read through ArchGDAL.  Maybe that's a bug in the retrieval code?

rafaqz:
I left some comments about this here:

https://github.com/yeesian/ArchGDAL.jl/blob/master/src/ogr/geometry.jl#L136-L146

evetion:
Yeah I don't think this is fixable. GDAL doesn't treat LinearRing as a stand-alone thing (see https://gdal.org/api/ogrgeometry_cpp.html#_CPPv413OGRLinearRing), and hence there are no LinearRing traits for 2.5D etc (even if you would intercept it in `getgeom(polygon)`). A way out would be adding isring as a type information, but that's a rather large change. Another way out is adding new LinearRing25D types ourselves, diverging/patching from GDAL calls itself, which is also rather ugly. I fear there's no easy way out.

Issue #425, opened 2024-05-10
Precompile error in v0.10.3


evandjpl:
My CI pipeline hit an error today while trying to precompile Julia libraries. Let me know if you need more info than this:

```
#17 6.069 ERROR: LoadError: UndefVarError: GDT_Int8 not defined
#17 6.743 Stacktrace:
#17 6.743   [1] eval
#17 6.743     @ ./boot.jl:368 [inlined]
#17 6.743   [2] eval
#17 6.743     @ ~/.julia/packages/ArchGDAL/SiT9o/src/ArchGDAL.jl:1 [inlined]
#17 6.743   [3] _broadcast_getindex_evalf
#17 6.743     @ ./broadcast.jl:670 [inlined]
#17 6.743   [4] _broadcast_getindex
#17 6.743     @ ./broadcast.jl:643 [inlined]
#17 6.743   [5] getindex
#17 6.743     @ ./broadcast.jl:597 [inlined]
#17 6.743   [6] copyto_nonleaf!(dest::Vector{ArchGDAL.GDALDataType}, bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Tuple{Base.OneTo{Int64}}, typeof(ArchGDAL.eval), Tuple{Base.Broadcast.Extruded{Vector{Any}, Tuple{Bool}, Tuple{Int64}}}}, iter::Base.OneTo{Int64}, state::Int64, count::Int64)
#17 6.765     @ Base.Broadcast ./broadcast.jl:1055
#17 6.765   [7] copy
#17 6.765     @ ./broadcast.jl:907 [inlined]
#17 6.765   [8] materialize
#17 6.765     @ ./broadcast.jl:860 [inlined]
#17 6.765   [9] (::ArchGDAL.var"#1#5")(a::Expr)
#17 6.771     @ ArchGDAL ./none:0
#17 6.771  [10] iterate(g::Base.Generator{NTuple{16, Expr}, ArchGDAL.var"#1#5"}, s::Int64)
#17 6.772     @ Base ./generator.jl:47
#17 6.772  [11] var"@convert"(__source__::LineNumberNode, __module__::Module, args::Vararg{Any})
#17 6.807     @ ArchGDAL ~/.julia/packages/ArchGDAL/SiT9o/src/utils.jl:102
#17 6.807  [12] include(mod::Module, _path::String)
#17 6.807     @ Base ./Base.jl:419
[...]

PR #423, opened 2024-05-06
add GDT_Int8 data type

tiemvanderdeure:
A Signed Int8 data type was added in GDAL 3.7: https://gdal.org/development/rfc/rfc87_signed_int8.html

This PR adds support for this datatype.

The GDALDataType enum defined in this package does not completely match the enum in the latest version of GDAL (I found it here: https://gdal.org/doxygen/gdal_8h.html).

For now I just added GDT_Int8 as field 14 and moved GDT_TypeCount to field 15. I'm not sure what consequences it would have to change the other fields?

PR #426, opened 2024-05-23
bump GDAL version to 1.7

rafaqz:


rafaqz:
@visr / @yeesian can we get a quick merge on this. I will also fix the registry now for the previous version

visr (approved):
Thanks! For context, see #425 which was a result of this missing compat in #423.

rafaqz:
The registry fix is here

https://github.com/JuliaRegistries/General/pull/107486

visr:
Sorry I directly triggered a new release, when that is merged in a few minutes, can you update your branch on top of that?

https://github.com/JuliaRegistries/General/pull/107490

PR #441, opened 2024-11-12
Fix Tables schema for empty layers

### LINKED ISSUE #440: Empty layer has incorrect schema
Related to https://github.com/evetion/GeoDataFrames.jl/issues/84

When reading an empty layer as a DataFrame, you get:
```julia
0×1 DataFrame
 Row │ ptr
     │ Ptr…
─────┴──────
```

Normally the schema is built automatically based on each feature (iteration), as Tables.schema is explicitely `nothing`. In this case, there are no features, so no schema.

evetion:
Fixes #440

This lets Tables.schema return an actual schema, but only when no features are present. 

If we always use the schema, we would need to also determine whether features attributes (columns) can be missing, and set the types accordingly. Without scanning all data (which is bad), this would result in all `isnullable` fields becoming `Union{Missing,T}` (also bad).


yeesian (changes requested):
Thanks for this! Can we also add a test for it in https://github.com/yeesian/ArchGDAL.jl/blob/master/test/test_tables.jl ?

evetion:
> Thanks for this! Can we also add a test for it in [`master`/test/test_tables.jl](https://github.com/yeesian/ArchGDAL.jl/blob/master/test/test_tables.jl) ?

Done! And I fixed some macros, as interpolating did not work correctly (most `@ogrerr` statements don't interpolate their string message, but `deletefeature!` does.

yeesian (approved):
It looks great, thank you!

PR #443, opened 2024-11-13
Update benchmark.yml to use `lts` instead of `1.6`.

asinghvi17:


evetion (approved):

rafaqz:
Huh
```  
  GDALError (CE_Failure, code 1):
  	OGRFieldDefn::SetDefault() not allowed on a sealed object
```

evetion:
I think some tests are just broken on master, might be a recent gdal release?

rafaqz:
Yes it must be. But looks like some fixing is needed

yeesian (approved):

yeesian:
Is the failure in https://github.com/yeesian/ArchGDAL.jl/actions/runs/11822349442/job/32939196358?pr=443 expected with this change?

evetion:
> Is the failure in [yeesian/ArchGDAL.jl/actions/runs/11822349442/job/32939196358?pr=443](https://github.com/yeesian/ArchGDAL.jl/actions/runs/11822349442/job/32939196358?pr=443) expected with this change?

I think that's because the latest macos runner has changed infrastructure? Probably best to pin the architecture to x64 (and maybe ubuntu instead of macos?)

PR #447, opened 2024-11-13
Fix tests for GDAL 3.9.

### LINKED ISSUE #444: Broken tests on recent GDAL
Some are easily fixed, but others require https://github.com/JuliaGeo/GDAL.jl/issues/187.

evetion:
Fixes #444

Changes
- FastGetExtent is now supported
- GDAL now [seals layer definitions](https://gdal.org/en/latest/development/rfc/rfc97_feature_and_fielddefn_sealing.html), which we cannot undo via C, making this test mostly moot.
- unsafe_gdaldem crashes the Julia process because of invalid arguments

The latter two require further changes, but it would be good to have the CI green for other PRs first.

### DIFF
--- a/test/test_drivers.jl
+++ b/test/test_drivers.jl
@@ -83,7 +83,7 @@ import ArchGDAL as AG
                 "RandomWrite" => false,
                 "RandomRead" => true,
                 "CurveGeometries" => false,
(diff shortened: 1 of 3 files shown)


yeesian (approved):
Thank you for this!

PR #448, opened 2024-11-13
Add a JLD2 extension to allow serializing geometries

### LINKED ISSUE #445: GDAL geometries can not be saved to JLD2 files
When trying to save GDAL geometries to .jld2 files they are saved as `NULL Geometry`. @asinghvi17 thinks this has to do with the c-pointers to memory that is not handled correctly by JLD2.

Being able to save  GDAL geometries to .jld2 files would be super useful for users like me who work with DimensionalData arrays and GDAL geometries in the same table.

asinghvi17:
Currently, this just overrides the JLD2 API to serialize and deserialize to/from WKB.

In future the framework is there to ship a CRS along, but we need the supporting API in ArchGDAL to set the spatial reference of a geometry, which has to be wrapped from GDAL still.  Would also have to serialize a spatial reference, probably as WKT.

MWE is in the tests.  It could potentially use more tests though...

Fixes #445


yeesian (approved):

yeesian (approved):
The failures are unrelated and I'm okay to add more tests as needed after merging this

yeesian:
It looks like we're getting the following failure in the tests:

```
JLD2 serialization: Error During Test at /home/runner/work/ArchGDAL.jl/ArchGDAL.jl/test/test_geometry.jl:1043
  Test threw exception
  Expression: AG.toWKT(geom2) == AG.toWKT(geom)
  GDALError (CE_Failure, code 10):
  	Pointer 'hGeom' is NULL in 'OGR_G_ExportToWkt'.
```

evetion:
> It looks like we're getting the following failure in the tests (for julia 1.6):
> 
> ```
> JLD2 serialization: Error During Test at /home/runner/work/ArchGDAL.jl/ArchGDAL.jl/test/test_geometry.jl:1043
>   Test threw exception
>   Expression: AG.toWKT(geom2) == AG.toWKT(geom)
>   GDALError (CE_Failure, code 10):
>   	Pointer 'hGeom' is NULL in 'OGR_G_ExportToWkt'.
> ```

FWIW, Do we still want to support 1.6? LTS is 1.10 now, and the merged package extension will require 1.9?

asinghvi17:
The extension probably doesn't trigger on 1.6, so I'm not surprised.

PR #451, opened 2024-11-15
Drop support for Julia 1.6.

evetion:
As 1.10 is the new LTS, and 1.6 gives failures.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - master
     tags: '*'
+# needed to allow julia-actions/cache to delete old caches that it has created
+permissions:
+  actions: write
+  contents: read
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - '1.10'  # lts
```
felixcremer, line 16: Couldn't we just use lts here directly? We might have to change the julia-action version but that should be possible

felixcremer (commented):
Looks good to me but for the future we could directly use lts in the github workflows so that it would always test on the current lts version.

rafaqz (approved):

yeesian (approved):

PR #454, opened 2024-12-16
Add generic setfield fallbacks.

evetion:
Encountered in GeoDataFrames when trying to write data with FixedPointNumbers (via LasIO). This PR allows writing any Real (converted to Float64) and Integer (converted to Int64). I've chosen the widest types possible.

I feel this is a better approach than making many package extensions for each new Number out there (which still can be done with these changes).


rafaqz:
Seems sensible, but can't say I really understand stand these internals, I've never used `setfield!` on features.

Probably @yeesian knows better

visr (approved):
Looks good to me!

PR #457, opened 2025-01-08
Define FeatureTrait for AbstractFeature

asinghvi17:
From @felixcremer's slack thread on GeometryOps reproject + https://github.com/rafaqz/Rasters.jl/pull/853


rafaqz:
Tests??

rafaqz (approved):

PR #462, opened 2025-02-09
Precompute chunks for RasterDataset

rafaqz:
I think this makes sense to compute up front as its kind of expensive

### DIFF
--- a/src/raster/array.jl
+++ b/src/raster/array.jl
@@ -21,9 +21,10 @@ us to be able to index into it like we would an array.
 Constructing a RasterDataset will error if the raster bands do not have all the
 same size and a common element data type.
 """
-struct RasterDataset{T,DS<:AbstractDataset} <: AbstractDiskArray{T,3}
+struct RasterDataset{T,DS<:AbstractDataset,C} <: AbstractDiskArray{T,3}
     ds::DS
     size::Tuple{Int,Int,Int}
+    chunks::C
 end
 
 function RasterDataset(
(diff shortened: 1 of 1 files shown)


src/raster/array.jl
```diff
@@ -21,9 +21,10 @@ us to be able to index into it like we would an array.
 Constructing a RasterDataset will error if the raster bands do not have all the
 same size and a common element data type.
 """
-struct RasterDataset{T,DS<:AbstractDataset} <: AbstractDiskArray{T,3}
+struct RasterDataset{T,DS<:AbstractDataset,C} <: AbstractDiskArray{T,3}
     ds::DS
     size::Tuple{Int,Int,Int}
+    chunks::C
```
yeesian, line 27: Might it be worth giving some kind of indication (maybe as a comment) what the chunk type `C` is like?

yeesian (dismissed):
Thanks! Just have a question about the chunk type since I'm not very familiar with the DiskArrays interface

src/raster/array.jl
```diff
@@ -21,9 +21,10 @@ us to be able to index into it like we would an array.
 Constructing a RasterDataset will error if the raster bands do not have all the
 same size and a common element data type.
 """
-struct RasterDataset{T,DS<:AbstractDataset} <: AbstractDiskArray{T,3}
+struct RasterDataset{T,DS<:AbstractDataset,C} <: AbstractDiskArray{T,3}
```
rafaqz, line 24:
```suggestion
struct RasterDataset{T,DS<:AbstractDataset,C<:DiskArrays.GridChunks} <: AbstractDiskArray{T,3}
```

We can put the type on the parameter

It's a wrapper for a tuple of RegularChunks AbstracVectors, that lazily generate the ranges for each axis of the chunking

src/raster/array.jl
yeesian: https://github.com/yeesian/ArchGDAL.jl/pull/462/commits/e1b8d101059574b2db900ccbde9ce497d1c4dcaf

yeesian (approved):

meggart (approved):

asinghvi17:
CI looks good barring the known MacOS failure, so I'll merge

PR #494, opened 2026-06-22
CI: run macOS tests natively on aarch64 instead of x64

asinghvi17:
Fix macos by manually adding the aarch64 testing matrix entries.

There's still an sqlite3 linking issue or something, probably needs a fix in the jll

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,6 +26,22 @@ jobs:
           - windows-latest
         arch:
           - x64
+        exclude:
+          # macOS-latest runners are Apple Silicon (arm64); requesting x64 runs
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -22,10 +22,19 @@ jobs:
           - 'nightly'
         os:
           - ubuntu-latest
-          - macOS-latest
           - windows-latest
         arch:
           - x64
```
evetion: Better to dump the arch and just use OS with their (good) defaults.

asinghvi17: Ah yeah good point, I forgot you could just rely on the default

asinghvi17:
Closing as resolved in another PR

evetion (approved):

PR #496, opened 2026-08-04
Expose the FID as a column in the tables interface

### LINKED ISSUE #142: Loading GeoPackage - not exposing id field

asinghvi17:
Fixes evetion/GeoDataFrames.jl#142 — reading a GeoPackage dropped the id column that QGIS and R's `sf` both show.

Database-like drivers (GPKG, PostGIS, SQLite, …) give each feature a persistent primary key, which GDAL keeps out of the field definitions, so it never reached the tables interface. Where a driver reports one via `OGR_L_GetFIDColumn` it now leads the table, under that driver's name for it. Drivers reporting `""` (Shapefile, FlatGeobuf, …) are unchanged.

```julia
julia> DataFrame(ArchGDAL.getlayer(ArchGDAL.read("rivers.gpkg"), 0))
357×5 DataFrame
 Row │ id     geom                     property_0  property_1      property_2
   1 │     1  Geometry: wkbLineString  6           Sutlej          null
```

Worth a look:

- **Features carry the column name.** It's a layer property, but schemas are built from rows, so features capture it at construction — also the only point it's knowable. A feature built from a featuredefn has no layer and no FID, and reports no FID column.
- **The GeoJSON driver promotes an `id` field to the FID *and* keeps `id` as a field.** The field wins; otherwise `columntable` throws `duplicate field name`.
- **Iteration state** is now `Tuple{Int64,Symbol}` rather than `Int64`, so the name is resolved per layer, not per feature (which measured ~+96% on iteration). An `Integer`-state method is kept for hand-driven callers. Iteration is unchanged vs master (161 vs 166 ns/feature, min of 7).
- **The FID takes index 1**, shifting the rest. Note `getcolumn(row, i)` and `columnnames` already disagree on master — index 1 is the first *field*, name 1 the first *geometry*. Left alone; happy to fix separately.

Suite: 1955 passed, 3 errored — the same 3 as master here (SQLite driver, network).

### DIFF
--- a/docs/src/tables.md
+++ b/docs/src/tables.md
@@ -26,3 +26,37 @@ dataset1 = ArchGDAL.read("data/multi_geom.csv", options = ["GEOM_POSSIBLE_NAMES=
 
 DataFrames.DataFrame(ArchGDAL.getlayer(dataset1, 0))
 ```
+
+## Feature IDs
+
+Database-like formats (GeoPackage, PostGIS, SQLite, ...) give every feature a
+persistent primary key, the *FID*. GDAL models it separately from the ordinary
+fields, so it does not appear among the field definitions. Where a driver
+reports such a column, via
+[`ArchGDAL.fidcolumnname`](@ref), it is included as the leading column of the
+table, under the name the driver gives it:
+
+```julia
+julia> dataset2 = ArchGDAL.read("data/rivers.gpkg")
+
+julia> ArchGDAL.fidcolumnname(ArchGDAL.getlayer(dataset2, 0))
+"fid"
+
+julia> DataFrames.DataFrame(ArchGDAL.getlayer(dataset2, 0))
+357×5 DataFrame
+ Row │ fid    geom                     property_0  property_1      property_2
+     │ Int64  IGeometry                String      String          String
+─────┼──────────────────────────────────────────────────────────────────────────
+   1 │     1  Geometry: wkbLineString  6           Sutlej          null
+   2 │     2  Geometry: wkbLineString  4           Svernaya Dvina  null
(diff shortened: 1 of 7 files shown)


evetion:
Nice! I have a draft somewhere that did it on the feature level and never finished it because of the high cost. Do you have a link to the docs/source code of GDAL clarifying FID is 1:n?

evetion (approved):

PR #506, opened 2026-09-05
Implement copy and deepcopy for geometries

### LINKED ISSUE #452: Implement `Base.(deep)copy` for `clone`?
Currently, a copy of a DataFrame containing ArchGDAL geometries is still aliased to the old dataframe's geometries. One has to explicitly clone them to separate them.

asinghvi17:
## Summary

- implement Base.copy for ArchGDAL geometries using clone
- implement the deepcopy hook using the same clone operation
- preserve repeated references when geometries are nested in deeply copied containers
- cover independent handles, mutation isolation, and NULL geometries

Fixes #452.

## Testing

- test_geometry.jl: 273/273 passed
- full suite: 1,978 passed; 3 environment-related errors from SQLite driver initialization and SSL certificate validation

### DIFF
--- a/src/ogr/geometry.jl
+++ b/src/ogr/geometry.jl
@@ -111,11 +111,52 @@ function clone(geom::AbstractGeometry{T}) where {T}
     end
 end
 
-Base.copy(geom::AbstractGeometry) = clone(geom)
+# Clone `geom` into a `W{T}` wrapper, preserving the geometry type parameter.
+# `clone` and `unsafe_clone` collapse NULL geometries to `wkbUnknown`, which
+# `Base.deepcopy` rejects: it asserts that `deepcopy_internal` returns a value
+# of the same type as its argument.
+function _clonegeom(::Type{W}, geom::AbstractGeometry{T}) where {W,T}
+    return W{T}(
+        geom.ptr == C_NULL ? GDAL.OGRGeometryH(C_NULL) : GDAL.ogr_g_clone(geom),
+    )
+end
 
+"""
+    copy(geom::AbstractGeometry)
+
+Returns a copy of the geometry with the original spatial reference system.
+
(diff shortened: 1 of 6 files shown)


asinghvi17:
Yes, Base.deepcopy_internal is actually public API

evetion:
> Yes, Base.deepcopy_internal is actually public API

TIL. But why do we need to override it?

asinghvi17:
We need to get the pointer cloned, otherwise deepcopy will happily just copy the pointer directly to the new struct

src/ogr/geometry.jl
```diff
@@ -111,6 +111,15 @@ function clone(geom::AbstractGeometry{T}) where {T}
     end
 end
 
+Base.copy(geom::AbstractGeometry) = clone(geom)
+
+function Base.deepcopy_internal(geom::AbstractGeometry, stackdict::IdDict)
```
evetion, line 155: If copy already clones, overriding deepcopy might not be needed at all?

evetion (dismissed):

src/ogr/geometry.jl
asinghvi17: I'm pretty sure deepcopy will skip that, but let me try it and find out

asinghvi17: Yeah just confirmed, deepcopy will indeed skip that

evetion (approved):
You might also consider not implementing this for preparedgeometries (erroring instead), given that you need more code and suddenly need to carry the originally geometry. I'm not sure about side effects for the latter, as in what happens if you destroy the original geometry? Or is GC further delayed now there's references around.

asinghvi17:
We also copy the geometry when copying.  But I am sort of inclined to say that prepared geometry can't be copied.

